PR #6, opened 2021-10-21
PYI-267: Add manifest.yml

AmritSidhu:
### What changed

Create a minimal manifest file to deploy the initial app to PaaS

### Why did it change

We want to view this application in PaaS

### Issue tracking

- [PYI-267](https://govukverify.atlassian.net/browse/PYI-267)

### Other considerations

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,0 +1,6 @@
+---
+applications:
+  - name: di-ipv-core-front
+    memory: 256M
+    buildpack: nodejs_buildpack
+    command: yarn start


manifest.yml
```diff
@@ -0,0 +1,6 @@
+---
+applications:
+  - name: di-ipv-core-front
+    memory: 512M
+    buildpack: nodejs_buildpack
+    command: yarn start
```
merlinc: When we deploy this, we might want to temporarily switch this to `yarn dev` as that will keep the process alive - currently `yarn start` `console.log`s and then exits.

manifest.yml
```diff
@@ -0,0 +1,6 @@
+---
+applications:
+  - name: di-ipv-core-front
+    memory: 512M
```
merlinc: Do we know how much memory we currently need?

AmritSidhu: I just took this size from the previous core-front app. In all the docs, the average size for deploying apps seem to be 256M?

merlinc: I just had a nice walk down memory lane to remember how to memory profile node apps - I'll show you what I did tomorrow after standup :)

It was a little bit of an unfair question, because the app can't be memory profiled at the moment because it exits!

When I add a `setInterval` to fire every 100ms and `console.log`, it's using about 6MB of memory. So for now, let's drop it down to 256, and re-review it when we have an webserver running. We'll want to have memory in the sweet spot of usage - there's no point spending £££ on memory / processor if we don't need it.

merlinc (dismissed):

manifest.yml
merlinc: @merlinc `yarn dev` also exits back into `nodemon` so we'll probably want a proper app here to do this properly

merlinc (dismissed):

merlinc (approved):

PR #15, opened 2021-11-12
Pyi 266 add button

merlinc:
### What changed

Add a GDS styled page with a button on it.

### Why did it change

While we're in the initial parts of development it will be useful to have an obvious page where we connect to the `core-back` system. This should make it easier to debug the initial flows as we're building them.

In the future this might be changed to some sort of debug page for development usage, much like the alpha homepage.

Note: There are 2 styles of page in this draft PR:
1. GDS raw template - to prove that customisation can be done
2. HMPO simple template - to prove that most of the page is boiler plate

#### To Do
- [ ] This button needs to connect to a backend system
- [ ] The HMPO style page needs to be able to change the label of the button, or needs to be more manually created
 

### Issue tracking

- [PYI-266](https://govukverify.atlassian.net/browse/PYI-266)

### Other considerations

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,10 +35,19 @@
     "mocha": "9.1.3",
     "nodemon": "2.0.13",
     "prettier": "2.4.1",
+    "sass": "1.43.4",
     "sinon": "11.1.2",
     "sinon-chai": "3.7.0"
   },
   "dependencies": {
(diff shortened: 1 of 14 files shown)


AmritSidhu:
This can be closed

PR #27, opened 2021-11-26
PYI-370: Add route to handle credential issuer callback

AmritSidhu:
### What changed

Added a route to handle the callback route and the params from the credential issuer

Added unit tests to test that 
 - the params received are being saved into the session
 - that the debug page is rendered on the callback

### Why did it change

So that core-front can retrieve authorization code from the credential issuer

### Issue tracking

- [PYI-370](https://govukverify.atlassian.net/browse/PYI-370)

### Other considerations

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -3,12 +3,30 @@ const {
   CREDENTIAL_ISSUER_AUTH_PATH,
 } = require("../../lib/config");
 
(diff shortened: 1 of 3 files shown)


AmritSidhu:
This will need to be wired up with the core-back/credential-issuer-stub work.

merlinc (approved):
This looks looks ok on its own, but will need changes to integrate with the credential issuer

PR #34, opened 2021-12-01
SonarCloud scan with coverage

idrop:
## Proposed changes

Run SonarCloud checks before PR merge. According to SonarCloud, "Next scan will generate a Quality Gate." which will give us stats like here: https://sonarcloud.io/summary/overall?id=alphagov_di-ipv-core-front

There's no coverage information yet, I'm hoping it'll be provided once there's an initial branch merge run through SonarCloud.

### DIFF
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -23,3 +23,11 @@ jobs:
         run: yarn lint
       - name: Run unit tests
         run: yarn test
+      - name: Run test coverage
+        run: yarn test:coverage
+      - name: Run sonarcloud scan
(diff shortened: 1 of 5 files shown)


merlinc:
This PR looks an awful lot like the un-pushed branch I had :grin: 

I've added a few files to tweak this PR, it should be pretty clear what they're doing.

I'm not sure what the issue with Sonar coverage reports. I think I've configured it correctly, but it still doesn't show up in SonarCloud :shrug: However, when I run Sonar locally via Docker, I see code coverage & all the other stats, so something might be working?

Anyway, I'm happy to merge this PR for now, and then come back to it if the coverage doesn't appear.

merlinc (approved):

PR #42, opened 2021-12-08
BAU: Extract BASE_URL from the environment

Wynndow:
### What changed

We're using the BASE_URL env variable to construct a redirect URL to
send to the credential issuer, but never actually extracting it from the
environment. This resulted in redirect URL's with `undefined` as the
host.


Wynndow:
Closing in favour of https://github.com/alphagov/di-ipv-core-front/pull/40

PR #57, opened 2022-01-26
PYI-518: Provide shared attributes to the credential issuers

AmritSidhu:
### What changed

Make a call to core-back to retrieve the JWT
Store JWT in session and send it as part of the auth request to the CRI's

Added testing
Using Regex to check that the jwt is base64 encoded
Raising an error if JWT is missing, is over 6000 characters and if is not valid base64 encoded

### Why did it change

When a user is redirected to a frontend CRI, the shared attribute set should be included, encoded in a JWT in the authorization request.

### Issue tracking

- [PYI-518](https://govukverify.atlassian.net/browse/PYI-518)

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -2,13 +2,40 @@ const axios = require("axios");
 const {
   API_BASE_URL,
   API_REQUEST_EVIDENCE_PATH,
+  API_ISSUED_JWT_PATH,
   EXTERNAL_WEBSITE_HOST,
 } = require("../../lib/config");
 
 module.exports = {
+  getJwt: async (req, res, next) =>  {
+    try {
+      const apiResponse = await axios.get(
+        `${API_BASE_URL}${API_ISSUED_JWT_PATH}`,
+        {
+          headers: {
+            "ipv-session-id": req.session.ipvSessionId
+          }
+        }
+      );
+
+      const jwt = apiResponse?.jwt;    //will need to be confirmed
+
+      if (!jwt) {
+        res.status(500);
+        return res.send("Missing jwt");
+      }
+
+      req.session.jwt = jwt;
+
+      next();
+    } catch (e) {
+      next(e);
+    }
+  }, 
+
   buildCredentialIssuerRedirectURL: async (req, res, next) => {
     const cri = req.session?.criConfig?.find(criConfig => criConfig.id === req.query.id);
-
+   
     if (!cri) {
       res.status(500);
       return res.send("Could not find configured CRI");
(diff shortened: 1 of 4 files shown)


src/app/credential-issuer/middleware.js
```diff
@@ -2,13 +2,47 @@ const axios = require("axios");
...
+        }
+      );
+
+      const jwt = apiResponse?.data?.jwt;    //will need to be confirmed
+      const base64regex = /^([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))?$/;
+
+      if (!jwt) {
+        res.status(500);
+        return res.send("Missing JWT");
+      } else if (jwt.length > 6000) { // this figure can be changed
```
dannyhayes-gds: Is it worth moving this number into an env variable? And then in the config.js file we can default it to 6000 if the env var is missing?

src/app/credential-issuer/middleware.js
```diff
@@ -2,13 +2,47 @@ const axios = require("axios");
...
+        res.status(500);
+        return res.send("JWT exceeds maximum limit");
+      } else if (!base64regex.test(jwt)) { 
+        res.status(500);
+        return res.send("Invalid base64 encoded JWT");
+      }
+
+      req.session.jwt = jwt;
```
dannyhayes-gds: Small change here, but could we change the req.session.jwt to be called something like req.session.sharedAttributesJwt - Just might make it slightly clearer when we have other JWT's going though the system.

src/app/credential-issuer/middleware.js
```diff
@@ -19,6 +53,7 @@ module.exports = {
     req.redirectURL.searchParams.append("client_id", cri.ipvClientId);
     req.redirectURL.searchParams.append("state", "test-state");
     req.redirectURL.searchParams.append("redirect_uri", `${EXTERNAL_WEBSITE_HOST}/credential-issuer/callback?id=${cri.id}`);
+    req.redirectURL.searchParams.append("request", req.session.jwt);
```
dannyhayes-gds: And same here

src/app/credential-issuer/middleware.js
```diff
@@ -2,13 +2,48 @@ const axios = require("axios");
...
+      const apiResponse = await axios.get(
+        `${API_BASE_URL}${API_SHARED_ATTRIBUTES_JWT_PATH}`,
+        {
+          headers: {
+            "ipv-session-id": req.session.ipvSessionId
+          }
+        }
+      );
+
+      const shared_attributes_jwt = apiResponse?.data?.shared_attributes_jwt;
```
dannyhayes-gds: Can we make this var camel case?

dannyhayes-gds (approved):

PR #59, opened 2022-02-02
Pyi 632/fix shared attributes request

dannyhayes-gds:
### What changed

- Fixed the shared attributes request
- Updated the base64 regex pattern to match against a JWT string structure.
- Fixed unit tests

### Why did it change
Incorrect property name was being used with the axios response object, and the base 64 regex pattern failed to match on the returned JWT's.

### Issue tracking

- [PYI-632](https://govukverify.atlassian.net/browse/PYI-632)


Danomanic (approved):
LGTM

PR #72, opened 2022-03-14
PYI-775 - Update JourneyHandler to respond to CRI event

LaiTang-GDS:
# What changed

Updated JourneyHandler to redirect the user if the Journey response is type of CRI response.

- [PYI-775](https://govukverify.atlassian.net/browse/PYI-775)


src/app/journey/middleware.js
```diff
@@ -21,8 +21,18 @@ async function journeyApi(action, ipvSessionId) {
 
 async function handleJourneyResponse(action, res, ipvSessionId) {
   const response = await journeyApi(action, ipvSessionId);
-  if (response.data?.redirect?.event) {
-    await handleJourneyResponse(response?.data?.redirect?.event, res);
+  if (response.data?.redirect) {
+    if(response.data?.redirect?.event) {
```
dannyhayes-gds, line 26: The returned data structure won't be in this format. I think it will just things like:

```suggestion
    if(response.data?.journey) {
```

```suggestion
    if(response.data?.page) {
```

```suggestion
    if(response.data?.cri) {
```

src/app/journey/middleware.js
```diff
@@ -21,8 +21,18 @@ async function journeyApi(action, ipvSessionId) {
...
+      await handleJourneyResponse(response?.data?.redirect?.event, res);
+    }
+    if(response.data?.redirect?.cri) {
+      if(!response?.data?.redirect?.cri?.authorizeUrl) {
+        res.error = 'AuthorizeUrl is missing'
+        res.status(500);
+        return;
+      }
+      return res.redirect(`${response?.data?.redirect?.cri?.authorizeUrl}`);
```
dannyhayes-gds: There is going to need to be a bit more here in order to redirect onto the cri. If you look back at the old way we redirect, there are also a bunch of query params needed.

All of the values for these will be coming back within thin cri response. (specifically: id, ipvClientId, authorizeUrl)

dannyhayes-gds: You also still need to retrieve the shared attributes and include that in the "request" query param

src/app/journey/middleware.js
LaiTang-GDS: Will be resolved in a separate ticket as we were working with the main branch version of the openApi

src/app/journey/middleware.js
LaiTang-GDS: Refactored code

AmritSidhu (approved):

PR #97, opened 2022-03-29
BAU: Remove redirect_uri param from CRI return request

Wynndow:
### Why did it change

The redirect URI was being included as a param to the CRI return lambda.
That lambda exchanges the auth code for a token and then uses it to
fetch the VC from the CRI - the redirect URI is not used for these
operations.

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -2,7 +2,6 @@ const axios = require("axios");
 const {
   API_BASE_URL,
   API_CRI_RETURN_PATH,
-  EXTERNAL_WEBSITE_HOST,
 } = require("../../lib/config");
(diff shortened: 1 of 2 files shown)


src/app/credential-issuer/middleware.js
```diff
@@ -21,7 +20,6 @@ module.exports = {
     const evidenceParam = new URLSearchParams([
       ["authorization_code", req.credentialIssuer.code],
       ["credential_issuer_id", req.query.id],
-      ["redirect_uri", `${EXTERNAL_WEBSITE_HOST}/credential-issuer/callback?id=${req.query.id}`],
```
dannyhayes-gds, line 24: I think the back-end still requires this value. It passes it into the /token endpoint of the CRI for validation reasons.

The Pojo class that core-back converts this request body to is expecting this var

PR #137, opened 2022-04-28
Pyic 1018 docker ecr build

devdutoitGDS:
## Proposed changes
- On every merge, Github Actions should build the docker image and push it to ECR. 
- Upload the cloudformation template.yaml along with the commit sha (image_tag.txt created during GA run) to S3: di-ipv-core-build-artifacts/core-front/template.zip

### Issue tracking
- [PYIC-1018](https://govukverify.atlassian.net/browse/PYIC-1018)

### DIFF
--- a/.github/workflows/ecr-docker-build.yml
+++ b/.github/workflows/ecr-docker-build.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
-      contents: read
+      contents: write
     environment: test
     steps:
       - uses: actions/checkout@dcd71f646680f2efd8db4afa5ad64fdcba30e748
--- a/.github/workflows/ecr-docker-build.yml
+++ b/.github/workflows/ecr-docker-build.yml
@@ -8,7 +8,8 @@ jobs:
     name: Docker build and push
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      id-token: write
+      contents: read
     environment: test
     steps:
       - uses: actions/checkout@dcd71f646680f2efd8db4afa5ad64fdcba30e748
--- a/.github/workflows/ecr-docker-build.yml
+++ b/.github/workflows/ecr-docker-build.yml
@@ -2,7 +2,7 @@ name: Docker build and push to ECR
 on:
   push:
     branches:
-      - PYIC-1018-docker_ecr_build
+      - main
 jobs:
   dockerBuildAndPush:
     name: Docker build and push
@@ -35,7 +35,6 @@ jobs:
           cd ${GITHUB_WORKSPACE}
           docker build -t "core-front-build:${IMAGE_TAG}" .
           docker tag "core-front-build:${IMAGE_TAG}" "${{ steps.login-ecr.outputs.registry }}/core-front-build:${IMAGE_TAG}"
-          docker image ls
       - name: Push docker image
         env:
           IMAGE_TAG: ${{ steps.create-tag.outputs.image_tag }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ecr-docker-build.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Create tag
+        id: create-tag
+        run: |
+          IMAGE_TAG="${{ github.sha }}-$(date +'%Y-%m-%d-%H%M%S')"
+          echo "::set-output name=image_tag::$IMAGE_TAG"
+      - name: Build docker image
+        env:
+          IMAGE_TAG: ${{ steps.create-tag.outputs.image_tag }}
+        run: |
+          cd ${GITHUB_WORKSPACE}
```
danworth: Just ran the bits of shell through `shellcheck` and we should use the following whenever we attempt to `cd`, but make it `|| exit 1` so the job fails.

```
cd ${GITHUB_WORKSPACE}
^--------------------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
```

It also points out some places where we reference variables without quoting to prevent splitting, so this line should become...

```
cd "${GITHUB_WORKSPACE}" || exit 1
```

.github/workflows/ecr-docker-build.yml
```diff
@@ -0,0 +1,50 @@
+name: Docker build and push to ECR
```
danworth: We should probably chance the file name and the workflow name now its pushing the template as well.... maybe `post-merge.yml`?

.github/workflows/ecr-docker-build.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          docker push "${{ steps.login-ecr.outputs.registry }}/core-front-build:${IMAGE_TAG}"
+      - name: Create template.yaml and sha zip file and upload to artifacts S3
+        env:
+          IMAGE_TAG: ${{ steps.create-tag.outputs.image_tag }}
+        run: |
+          cd ${GITHUB_WORKSPACE}/deploy
+          echo "${IMAGE_TAG}" > image_tag.txt
+          zip template.zip template.yaml image_tag.txt
+          aws s3 cp template.zip s3://di-ipv-core-build-artifacts/core-front/
```
Wynndow: Might be worth sticking a new line on the end?

.github/workflows/ecr-docker-build.yml
```diff
@@ -0,0 +1,50 @@
...
+    branches:
+      - main
+jobs:
+  dockerBuildAndPush:
+    name: Docker build and push
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    environment: test
```
Wynndow: Do we need this `environment`. I've had a look at the docs and its use is a little unclear. We don't use it in other workflows though.

.github/workflows/ecr-docker-build.yml
devdutoitGDS: good shout, done

.github/workflows/ecr-docker-build.yml
devdutoitGDS: done

.github/workflows/ecr-docker-build.yml
devdutoitGDS: good spot, killed it

danworth (dismissed):
Looks good mobbed with Dev, Chris and myself 🏅

danworth (approved):
Looks good 👍

PR #138, opened 2022-04-29
BAU: Remove Oauth params from redirect request onto the CRI's

dannyhayes-gds:
### What changed
Removed the Oauth params from the core-front redirect request onto the CRI.

### Why did it change
We no longer need these query params as all the Oauth values are inside the JAR. Both the real passport cri and the stubs are setup to read these params from the contents of the "request" JWT.

### DIFF
--- a/src/app/ipv/middleware.test.js
+++ b/src/app/ipv/middleware.test.js
@@ -164,7 +164,7 @@ describe("journey middleware", () => {
 
     it("should be redirected to a valid redirectURL", async function() {
       await middleware.updateJourneyState(req, res, next);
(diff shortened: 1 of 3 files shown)


kerrr (approved):

Danomanic (approved):
I can see the unneeded params have now been removed.
Looks good!

AmritSidhu:
```
redirectToAuthorize: async (req, res) => {
    res.redirect(req.redirectURL);
}
```
Will this still redirect fine with the things being moved into the request JWT? Otherwise all good

dannyhayes-gds:
@AmritSidhu Yeah it should still redirect fine. It just now redirects with slightly different request params. The actual url we are redirecting to is unchanged only the params at the end.

PR #153, opened 2022-05-04
PYIC-1067: Ship logs to Splunk

Wynndow:
### What changed

Add a subscription filter for the API gateway access logs, and the ECS
logs, to send them to Splunk via Cyber's Centralised Security Logging
Service (CSLS).

### Why did it change

See here for more details:
https://github.com/alphagov/centralised-security-logging-service

Also defines a retention period for the log groups of 14 days. This is
in line with what we're doing elsewhere and should be enough in case the
shipping breaks for some reason.

### Issue tracking

- [PYIC-1067](https://govukverify.atlassian.net/browse/PYIC-1067)


vam-gds (approved):
lgtm

PR #157, opened 2022-05-06
PYIC-1129 Conditional for dev envs to use development ecr

devdutoitGDS:
## Proposed changes
Currently, developer environments use the ENVIRONMENT variable when creating a stack in concourse.

In the current form this template uses ENVIRONMENT to access the correct ECR repository, which will fail as we don't have a repository for each developer.

This change will use the development ecr repository for those builds.

### What changed
To accommodate developer environment stacks:
- Change the allowed environments to allow for dev*
- Create a condition that is set if the environment is not build, staging, integration or production. Use that condition to set the ecr repository to `development`.

### Issue tracking
- [PYIC-1129](https://govukverify.atlassian.net/browse/PYIC-1129)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -16,12 +16,7 @@ Parameters:
   Environment:
     Description: The name of the environment to deploy to.
     Type: String
-    AllowedValues:
-      - development
-      - build
-      - staging
-      - integration
-      - production
+    AllowedPattern: ((production)|(integration)|(staging)|(build)|(dev.*))
   ImageTag:
     Description: The tag of core-front image to deploy in the task definition.
     Type: String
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -45,6 +40,13 @@ Parameters:
       The number of core front ecs tasks to run.
     Type: Number
 
+Conditions:
+  IsNotDevelopment: !Or
+    - !Equals [ !Ref Environment, build ]
+    - !Equals [ !Ref Environment, staging ]
+    - !Equals [ !Ref Environment, integration ]
+    - !Equals [ !Ref Environment, production ]
+
```
devdutoitGDS, line 49: could not get a wildcard to work for `dev*`, so had to use this inverse approach.

danworth (approved):
Looks good (assuming it lints correctly but it looks ok to me; adding CFN linter/ generate change sets on PRs is on the to-do list) 🏅

PR #182, opened 2022-05-18
BAU Utility Script to deploy developer version of Core Front

danworth:
This script will build the docker image from the local source code, push
the image into the development account's core-front-development registry
and update the developer's core-front stack to use it.

### More Info
This is the first iteration of a script to help developers test out their local core-front code in their developer environments. Likely a few rough edges so please feel free to update and correct as its used.


devdutoitGDS (approved):
LGTM

PR #197, opened 2022-05-25
PYIC-1311: Add pyi-kbv-fail error page

AmritSidhu:
### What changed

Create an error page for the pyi-kbv-fail

Designs are copied from the [figma](https://www.figma.com/file/nglBHtbzJYEeST43Iu1jW3/PYI-DBS-Return-Journey-Mockups-(Unhappy-Paths)) diagrams

Figma:
Screenshot 2022-05-25 at 11 27 25

core-front:
Screenshot 2022-05-25 at 11 34 44

### Why did it change

We need to create a new error page for when a user fails a KBV check and can not continue their journey.

### Issue tracking

- [PYIC-1311](https://govukverify.atlassian.net/browse/PYIC-1311)

### DIFF
--- a/src/locales/en/pages.yml
+++ b/src/locales/en/pages.yml
@@ -21,8 +21,15 @@ pyiNoMatch:
     - <a href="/ipv/journey/next" class="govuk-button" data-module="govuk-button"> {{ translate("buttons.next") }} </a>
     - <a href="https://signin.account.gov.uk/contact-us" class="govuk-link"> Contact the GOV.UK account team (opens in a new tab) </a>
 
-
-# Pages created from the prototype
+pyiKbvFail:
+  title: Sorry, we cannot prove your identity right now
+  content:
+    - This might be because you gave the wrong answer to some of the security questions.
+    - To keep your information safe, we cannot show which questions you answered incorrectly. 
+    - <h2 class="govuk-heading-m"> What you can do </h2>
+    - Continue to the service you were trying to use and look for other ways to prove your identity.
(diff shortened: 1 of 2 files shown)


LaiTang-GDS (approved):

src/locales/en/pages.yml
```diff
@@ -21,8 +21,15 @@ pyiNoMatch:
     - <a href="/ipv/journey/next" class="govuk-button" data-module="govuk-button"> {{ translate("buttons.next") }} </a>
     - <a href="https://signin.account.gov.uk/contact-us" class="govuk-link"> Contact the GOV.UK account team (opens in a new tab) </a>
 
-
-# Pages created from the prototype
+pyiKbvFail:
+  title: Sorry, we cannot prove your identity right now
+  content:
+    - This might be because you gave the wrong answer to some of the security questions.
+    - To keep your information safe, we cannot show which questions you answered incorrectly.
```
Wynndow, line 28: This is correct as it follows what's in figma, but I'm just wondering about the use of `cannot`. I seem to remember learning from Sabrina that we shouldn't use negative contractions like this, and it should be `can not`. I'll check in with her. We can always update it if needed.

sdurnberger: 'Cannot' is fine, that's correct! It would be something like 'can't' that's a contraction and against our style guide 😁

Wynndow: Thank you!

Wynndow (approved):

PR #198, opened 2022-05-25
PYIC-1254 Prepare to remove ApiBaseUrl param

danworth:
Add a default value so that we can remove setting it via Concourse. The
entire ApiBaseUrl will then be removed from the template as it is no
longer used.

## Proposed changes
As above. The next step is to stop Concourse setting the value before removing it entirely.


vam-gds (approved):
lgtm

PR #201, opened 2022-05-25
BAU: Remove ApiBaseUrl from frontend deploy script

Wynndow:
### What changed

Remove ApiBaseUrl from frontend deploy script

### Why did it change

It has been removed from the template. Keeping it here was causing a
validation error.


danworth (approved):
🏅 🏅

PR #231, opened 2022-06-10
BAU: Remove authorize route from CRI router

Wynndow:
### What changed

Remove authorize route from CRI router

### Why did it change

Building the URL to send a user to a CRI is handled in the
handleJourneyResponse method of the ipv middleware. We never use this
authorize route so should remove it.


AmritSidhu (approved):
Good spot

PR #234, opened 2022-06-13
PYIC-1398 test GTM event tracking code and custom javascript build

wilsond-gds:
## Proposed changes

This ticket adds basic JavaScript code to fire a google tag manager tracking event on core-front when the user clicks the 'feedback' link in the phase banner

### What changed

`application.js` is added to `src/assets/javascript/` and the code in that file should be integrated into the design system javascript bundle when `yarn build` runs

### Why did it change

To begin the process of understanding how GTM can work on the front end and to give an early indication to the analytics team of the form of the information they will receive.

### Issue tracking

- [PYIC-1398](https://govukverify.atlassian.net/browse/PYIC-1398)

### DIFF
--- a/src/assets/javascript/application.js
+++ b/src/assets/javascript/application.js
@@ -1,5 +1,5 @@
 const feedback = document.querySelector('.govuk-phase-banner a');
-feedback.addEventListener("click", event => { // eslint-disable-line no-unused-vars
+feedback.addEventListener("click", function(event){ // eslint-disable-line no-unused-vars
   window.dataLayer = window.dataLayer || [];
   window.dataLayer.push({
     event: "gaEvent",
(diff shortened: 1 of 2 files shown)


src/assets/javascript/application.js
```diff
@@ -0,0 +1,14 @@
+const feedback = document.querySelector('.govuk-phase-banner a');
```
kr8n3r, line 1: use of ES6 will exclude ie11 and below

AmritSidhu (dismissed):

src/assets/javascript/application.js
wilsond-gds: ah, thanks Jani - I've refactored the code and tested it in IE11.

LaiTang-GDS (approved):

PR #251, opened 2022-06-22
PYIC-1460 Increase Desired Task Count in Production

danworth:
This is necessary for increased resilience by running across two
availability zones. No other changes should be necessary since Fargate's
default deployment strategy is to spread across AZs.

Update the template so that the task desired count is defined within the
template rather than passing from CI configuration. This should make it
simpler to amend and find in future and simplify the roll-out of such
changes.

## Proposed changes
As above
### What changed
As above

### Why did it change
In preparation for going live we ensuring our system is running across at least 2 AZ. We may likely have to make further adjustments to desired task count (and memory) once we've done some performance testing but this meets the requirements whilst balancing cost.

### Notes
This has been applied into my dev environment and works as expected.

CFN `Mappings` section does not support defaults. All of the possible values passed to the `Environments` param are unknown since the developer environments are numerous and change with onboarding/offboarding. I tried different ways of making the `!FindInMap` work with the `Environment` parameters which I think is more readable than the account ID but unfortunately we are restricted by CFN functionality. Even making the `!FindInMap` conditional does work because the validation of the template still fails.

### Issue tracking

- [PYIC-1460](https://govukverify.atlassian.net/browse/PYIC-1460)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -34,8 +34,11 @@ Parameters:
     Type: String
   DesiredTaskCount:
     Description: >-
-      The number of core front ecs tasks to run.
+      This is being replaced by the desiredTaskCount with the Mapping section. Its removal
+      must be coordinated with changes to our CI configuration to avoid breaking
+      deployments in the interim.
     Type: Number
+    Default: 1
 
 Conditions:
   IsNotDevelopment: !Or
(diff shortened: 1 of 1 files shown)


Wynndow (approved):
Looks good. Using the account IDs with the well placed comment seems good to me!

PR #255, opened 2022-06-22
PYIC-1456: create core-front dynamoDB table

dannyhayes-gds:
### What changed
Create a new DynamoDB table to be used by core front to hold our front-end session data. 

The table has an "id" partition key with ttl enabled based off the "expires" field.

Also added a new policy to the core-front execution IAM role to allow access to the table from the ECS instance.

### Why did it change
We are using DynamoDB as our persistent data storage of the front-end session data. We are using an npm package to handle the connection to and reading from the dynamoDB table which expects to use an "id" param to read from the table.

### Issue tracking

- [PYIC-1456](https://govukverify.atlassian.net/browse/PYIC-1456)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -315,19 +315,17 @@ Resources:
             Statement:
               - Effect: Allow
                 Action:
-                  - "dynamodb:BatchGet*"
-                  - "dynamodb:DescribeStream"
+                  - "dynamodb:BatchGetItem"
                   - "dynamodb:DescribeTable"
-                  - "dynamodb:Get*"
+                  - "dynamodb:GetItem"
                   - "dynamodb:Query"
                   - "dynamodb:Scan"
-                  - "dynamodb:BatchWrite*"
-                  - "dynamodb:CreateTable"
-                  - "dynamodb:Delete*"
-                  - "dynamodb:Update*"
+                  - "dynamodb:BatchWriteItem"
+                  - "dynamodb:DeleteItem"
+                  - "dynamodb:UpdateItem"
                   - "dynamodb:PutItem"
                 Resource:
-                  - "arn:aws:dynamodb:*:*:table/core-front-sessions-*"
+                  - !GetAtt CoreFrontSessionsTable.Arn
 
   ECSTaskRole:
     Type: 'AWS::IAM::Role'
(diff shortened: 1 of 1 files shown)


AmritSidhu (dismissed):

deploy/template.yaml
```diff
@@ -419,6 +439,26 @@ Resources:
...
+      TableName: !Sub "core-front-sessions-${Environment}"
+      BillingMode: "PAY_PER_REQUEST"
+      AttributeDefinitions:
+        - AttributeName: "id"
+          AttributeType: "S"
+      KeySchema:
+        - AttributeName: "id"
+          KeyType: "HASH"
+      TimeToLiveSpecification:
+        AttributeName: "expires"
```
Wynndow, line 453: Could this be changed to `ttl` to be consistent with how we configure our other Dynamo tables?

deploy/template.yaml
```diff
@@ -309,6 +309,26 @@ Resources:
...
+                  - "dynamodb:Get*"
+                  - "dynamodb:Query"
+                  - "dynamodb:Scan"
+                  - "dynamodb:BatchWrite*"
+                  - "dynamodb:CreateTable"
+                  - "dynamodb:Delete*"
+                  - "dynamodb:Update*"
+                  - "dynamodb:PutItem"
+                Resource:
+                  - "arn:aws:dynamodb:*:*:table/core-front-sessions-*"
```
Wynndow: I think you could make this:
```suggestion
                  - !GetAtt CoreFrontSessionsTable.Arn
```

deploy/template.yaml
dannyhayes-gds: I'm not sure. I've used expires here as this is by default what the dynamodb-sessions package we are using on the front-end calls the param. We'd need to look and see if this can be configured on the front end to match

deploy/template.yaml
dannyhayes-gds: Ah nice, thanks! I'll update it to use this

deploy/template.yaml
Wynndow: Ah ok, if it's not an arbitrary choice don't worry, no need to make things more complicated.

Wynndow (dismissed):

kerrr (approved):

PR #258, opened 2022-06-23
PYIC-1493 Add a healthcheck endpoint

danworth:
Adds a basic health check endpoint to confirm the server is running. This
is then used by the Application Load Balancer Target Group where it
expects a 200 http status code.

This change removes the noisy 404 http status codes in the app logs. The
request to the health check and its 200 response are not logged in the
application logs as per feedback from the team. The Target Group metrics
show the healthy host count and can be used to investigate.

## Proposed changes
As above
### What changed
Added a `/healthcheck`

### Why did it change
Remove the noise from `404` healthcheck requests to the root of the app. This also enables us to implement Route53 healthcheck on the HZ and improve our Amazon Shield Advanced protection.

This has been run into my dev environment and the ECS task becomes stable as expected.

### Issue tracking

- [PYIC-1493](https://govukverify.atlassian.net/browse/PYIC-1493)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -184,8 +184,9 @@ Resources:
     Properties:
       HealthCheckEnabled: TRUE
       HealthCheckProtocol: HTTP
+      HealthCheckPath: /healthcheck
       Matcher:
-        HttpCode: 200-499
+        HttpCode: 200
       Port: 80
       Protocol: HTTP
       TargetType: ip


src/app/healthcheck/middleware.js
```diff
@@ -0,0 +1,10 @@
+const axios = require("axios");
```
Wynndow: Do you need this require?

Wynndow (commented):
Just that thing about the import, other wise 🫶

src/app/healthcheck/middleware.js
danworth: Thanks mate, nope  (I caught it before but didn't actually add it to the commit 🤦 , updated and pushed)

Wynndow (approved):

PR #272, opened 2022-06-27
[PYIC-758] Add cookie banner (and google analytics code)

wilsond-gds:
## Proposed changes

This PR implements the GOV.UK cookie banner code and interaction on core-front. It also uses the DI Auth front-end javascript code to implement page views and other google analytics events if the user consents to tracking.

The code uses the 

#### HTML code

`ipv-template.html` now includes a complete cookie banner, which pulls in the nunjucks banner code at `src/views/shared/banner.njk`. The content of the banner code is found in `src/locales/en/default.yml`. This code is added to all pages but hidden via CSS unless front-end javascript code is available.

#### Front-end JavaScript code

All front-end JS code for this repo is held in `src/assets/javascript/application.js`. This code shows and hides the cookie banner based on user interaction, controls the setting and detection of cookies, adds the google tag manager code if the user allows it, and sends page view tracking events back to the GDS google analytics account. 

The front-end code is `init`-ed from the script tag at the bottom of each HTML page, which looks like this:

`<script type="text/javascript" nonce='{{ scriptNonce }}'>window.GOVSignIn.appInit("{{ gtmId }}", "{{ analyticsCookieDomain }}", "{{ hmpoPageKey }}")</script>`

All of the nunjucks variables are created by the back-end app code. The `{{ hmpoPageKey}}` is required so that the state of the user’s journey (errors or important waypoints) can be sent to google analytics.

#### Back-end app code

The back-end code retrieves global variables and adds them to the nunjucks templates rendered in the app. `src/lib/config.js`, `src/lib/locals.js`, `src/lib/settings.js` and `src/lib/strings.js` process the global variables and add them to the session, so that `src/lib/middleware.js` can output the correct values for the cookie banner and google analytics tracking as in the `script` tag above.

### Why did it change

This code removes the testing code that could track users who have not opted in, and adds a cookie banner so the user controls the use of their data.

### Issue tracking

[The original DI Auth code is in this repository](https://github.com/alphagov/di-authentication-frontend/tree/main/src/assets/javascript).

- [PYIC-758](https://govukverify.atlassian.net/browse/PYIC-758)

### Other considerations

- [ ] If PR is accepted then the code can be reused for the passport CRI.


src/app.js
```diff
@@ -6,16 +6,17 @@ const DynamoDBStore = require("connect-dynamodb")(session);
...
-AWS.config.update({
-  region: "eu-west-2",
-});
-const dynamodb = new AWS.DynamoDB();
-
-const dynamoDBSessionStore = new DynamoDBStore({
-  client: dynamodb,
-  table: SESSION_TABLE_NAME,
-});
+// AWS.config.update({
```
dannyhayes-gds: This stuff is going to be needed uncommenting again

src/app.js
```diff
@@ -26,22 +27,24 @@ const loggerConfig = {
 const sessionConfig = {
   cookieName: "ipv_core_service_session",
   secret: SESSION_SECRET,
-  sessionStore: dynamoDBSessionStore,
+  //sessionStore: dynamoDBSessionStore,
```
dannyhayes-gds: Same here

src/app.js
```diff
@@ -26,22 +27,24 @@ const loggerConfig = {
...
 };
 
 const { router } = setup({
   config: { APP_ROOT: __dirname },
   port: PORT,
   logs: loggerConfig,
   session: sessionConfig,
-  redis: false,
+  //redis: false,
```
dannyhayes-gds: Same here

wilsond-gds: yep, missed that - will re-commit

src/lib/settings.js
```diff
@@ -0,0 +1,9 @@
+// const { generateNonce } = require("./strings");
```
dannyhayes-gds: I think we can delete this file

src/views/shared/ipv-template.html
```diff
@@ -4,49 +4,12 @@
...
 
 {% set hmpoPageTitle = "pages."+hmpoPageKey+".title" %}
 {% set hmpoPageTitleServiceNameRequired = "pages."+hmpoPageKey+".serviceNameRequired" %}
 {% set hmpoPageHtmlContent = "pages."+hmpoPageKey+".htmlContent" %}
 
-
-{% block head %}
-{# super copies over the code for this block from HMPO template #}
-{{ super() }}
-<script>
```
dannyhayes-gds, line 16: Is all of this stuff definitely not needed? Just wanted to double check in case we removed anything accidentally during the merge conflicts we had?

src/lib/settings.js
wilsond-gds: Can definitely remove it - it will mean the repos will be slightly out of sync but we probably need to refactor after MVP.

src/views/shared/ipv-template.html
wilsond-gds: This part of the commit is an earlier version of the GA code that we added to check the tracking code was working. It's all handled in application.js now and we can’t use the `<noscript>` code because it opts people into tracking without their consent. The nunjucks block removes the cookie banner from the source so we need to go back to the default behaviour there.

LaiTang-GDS (approved):

PR #277, opened 2022-06-28
PYIC-1511 Updates to the dev env deployment script

vam-gds:
### What changed
Updates to the dev env deployment script

Further updates to script, fixed all remaining actual issues

	modified:   utils/README.md
	modified:   utils/deploy_to_dev_env.sh

### Why did it change
This is just some quality of life improvements to the deployment shell script. 

### Issue tracking

- [PYIC-1511](https://govukverify.atlassian.net/browse/PYIC-1511)

### Other considerations


vam-gds:
This was originally pushed up under: https://github.com/alphagov/di-ipv-core-front/pull/274
This is basically the same thing with more cleanups, but assigned to a proper ticket for tracking purposes. (I renamed the branch which ended up creating a new PR only reason why this PR exists)

Shellcheck run on it - 2 warnings ignored as the warnings are not taking into account the comparison of a value in an array and assume its related to regex which it is not.

utils/deploy_to_dev_env.sh
```diff
@@ -4,55 +4,120 @@
...
+    exit 1
+  fi
+}
 
-echo "Logging into developement ECR registry"
-ecr_registry="130355686670.dkr.ecr.eu-west-2.amazonaws.com"
-aws ecr get-login-password --region eu-west-2 | \
-  docker login --username AWS --password-stdin "${ecr_registry}"
+function clone_config_repo() {
+  git clone git@github.com:alphagov/di-ipv-config.git config-repo
```
danworth: So at present if I do not provide the `-e` flag the error message isn't right because process_args is called before init and the `environment` array is not populated, not a huge deal but I imagine it'll happen frequently the first time people run it. This is what you get
```
GDS11321:utils dan.worth$ aws-vault exec di-ipv-dev -- ./deploy_to_dev_env.sh

Usage: ./deploy_to_dev_env.sh:
        [-e] Provide environment name from:
```


I get why you're cloning the repo but ideally I don't want to wait to clone it when I already have it (its also unexpected). For simplicity why not just add a link to the file in the help text, something like the following and add anything extra to the README? Or look for it on the dev's laptop, most people will clone it into the same parent directory. I feel like we're on the brink of over engineering this, despite my fondness for a bit of `bash` :)

```
echo -e "\t[-e] Provide environment name from those listed at https://github.com/alphagov/di-ipv-config/blob/main/core/ci/core-developer-pipelines/generate-pipelines/list_of_developers.txt\n"
```

utils/deploy_to_dev_env.sh
```diff
@@ -61,27 +126,35 @@ function isStackUpdateComplete() {
     | jq '.Stacks[].StackStatus' -r )"
 
   echo "Current status is: ${status}"
-  if [[ "$status" == "UPDATE_COMPLETE" ]]; then
+  if [[ " ${stack_status[*]} " =~ " ${status} " ]]; then
```
danworth, line 113: ah nice, haven't seen this done in a while :)

utils/deploy_to_dev_env.sh
danworth: If you want to be helpful and provide a list of configured envs then how about using something like the following to derive it from the available front end stacks?? This also means it'll make sure the stack actually exists which is necessary further on....
```
GDS11321:utils dan.worth$ aws-vault exec dev-admin -- aws cloudformation list-stacks --region eu-west-2 | jq '.StackSummaries[] | select(.StackName | contains("front")).StackName' -r | sort | uniq | cut -d '-' -f 3,4
dev-amrits
dev-aprilm
dev-chrisw
dev-danh
dev-danp
dev-danw
dev-lait
dev-patrickb
dev-vamh
dev-wojciecho
```

danworth: To be honest I'm leaning towards a link to the file and a note that if they're not listed to ask for their environment to be set up. That seems the most simple but open to other ideas.

vam-gds: so I have pushed up updates which sorts this out. 

We cant be sure they have the correct credentials that early on so aws cant be used. 

I have completely reworked the git clone it downloads only 448K:

```
du -sh config-repo/
488K    config-repo/
```

it runs before anything else quietly downloads a very sparse version of the repo assigns the contents of that file to an array and then removes itself and everything now works as expected (quietly in the background no one will be aware minimal interfernace) 

I do this because I cant be sure they are calling aws-vault or gds to correct env, where the ipv-config repo resides what branch it may be on, when they last updated it etc - this way its completely clean we get the latest update directly from what should be a source of truth and it should be very quick now considering its less than half a meg download. 

so for me locally:

```
$ time aws-vault exec core-dev -- ./deploy_to_dev_env.sh

Usage: ./deploy_to_dev_env.sh:
        [-e] Provide environment name from: dev-danw dev-danp dev-danh dev-chrisw dev-amrits dev-kerrr dev-lait dev-ianj dev-derrenw dev-patrickb dev-vamh dev-wojciecho dev-aprilm

        [-h] Print this usage guide


real    0m1.814s
user    0m0.085s
sys     0m0.055s
```
```
$ time aws-vault exec core-dev -- ./deploy_to_dev_env.sh -e test
        ./deploy_to_dev_env.sh: Error: 'test' does not match any of the valid known environments: dev-danw dev-danp dev-danh dev-chrisw dev-amrits dev-kerrr dev-lait dev-ianj dev-derrenw dev-patrickb dev-vamh dev-wojciecho dev-aprilm


Usage: ./deploy_to_dev_env.sh:
        [-e] Provide environment name from: dev-danw dev-danp dev-danh dev-chrisw dev-amrits dev-kerrr dev-lait dev-ianj dev-derrenw dev-patrickb dev-vamh dev-wojciecho dev-aprilm

        [-h] Print this usage guide


real    0m1.947s
user    0m0.085s
sys     0m0.067s
```

roughly about 2 seconds now.

utils/deploy_to_dev_env.sh
```diff
@@ -4,55 +4,104 @@
...
+    esac
+  done
+
+  [ -z "$ENVIRONMENT" ] || export ENVIRONMENT=$ENVIRONMENT
+  [ -z "$ENVIRONMENT" ] && print_error_exit "Environment Name: ${ENVIRONMENT} not provided (-e) :: exiting"
+}
 
+function check_dependencies() {
+  if ! which jq >/dev/null; then
+    print_error_exit "Please install jq to use this script - more info here: https://formulae.brew.sh/formula/jq"
```
danworth, line 53: this is a helpful addition 👍

utils/deploy_to_dev_env.sh
```diff
@@ -4,55 +4,104 @@
...
+  STACK_NAME="core-front-${ENVIRONMENT}"
+  DEV_IMAGE_TAG="${ENVIRONMENT}-$(date +%s)"
+  echo -e "\nCurrent Environment Configuration, as follows:"
+  echo -e "\t* Region:      '${region}'"
+  echo -e "\t* Account ID:  '${account_id}'"
+  echo -e "\t* Environment: '${ENVIRONMENT}'"
+  echo -e "\t* Stack Name:  '${STACK_NAME}'"
+  echo -e "\t* Image Tag:   '${DEV_IMAGE_TAG}'\n"
+  check_dependencies
+  check_connection
```
danworth, line 88: Useful summary 👍

danworth (approved):
🏅

PR #281, opened 2022-06-29
BAU Added a check for the core dev account id

vam-gds:
### What changed
BAU Added a check for the core dev account id (to make sure this is being run in core-dev)

### Why did it change
This just an additional check to make sure this not accidentally run in a different environment. 

### Issue tracking

BAU TASK

### Other considerations


danworth (approved):
That's a nice safety check, it would have failed to find the stack I would have thought if run against the wrong account but better to safe 👍 🛡️

PR #342, opened 2022-08-02
PYIC-1719: update core-front to use new renamed lambdas

dannyhayes-gds:
### What changed
Update core-front to use some of the renamed lambda endpoints.

### Why did it change
A lot of the core-back lambdas have been renamed with some getting new api endpoint paths. For these ones they were deployed as an additional copy of the lambda under the new path, while keeping a version accessible on the old api gateway path. This PR is part of the changes to move over to the new renamed lambdas allowing the old ones to be deleted.

### Issue tracking

- [PYIC-1719](https://govukverify.atlassian.net/browse/PYIC-1719)

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -1,7 +1,7 @@
 const axios = require("axios");
 const {
   API_BASE_URL,
-  API_CRI_RETURN_PATH,
+  API_CRI_ACCESS_TOKEN_PATH,
   EXTERNAL_WEBSITE_HOST,
 } = require("../../lib/config");
 const { generateAxiosConfig } = require("../shared/axiosHelper");
@@ -31,7 +31,7 @@ module.exports = {
     try {
       logger.info("calling cri return lambda", { req, res });
       const apiResponse = await axios.post(
-        `${API_BASE_URL}${API_CRI_RETURN_PATH}`,
+        `${API_BASE_URL}${API_CRI_ACCESS_TOKEN_PATH}`,
         evidenceParam,
         generateAxiosConfig(req.session.ipvSessionId)
       );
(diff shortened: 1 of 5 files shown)


src/app/credential-issuer/middleware.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
     try {
       logger.info("calling cri return lambda", { req, res });
       const apiResponse = await axios.post(
-        `${API_BASE_URL}${API_CRI_RETURN_PATH}`,
+        `${API_BASE_URL}${API_CRI_ACCESS_TOKEN_PATH}`,
```
Wynndow, line 34: Might be worth updating the wording of that logging just above?

src/app/debug/middleware.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
     try {
       logger.info("calling issued credentials lambda", { req, res });
       const apiResponse = await axios.get(
-        `${API_BASE_URL}${API_ISSUED_CREDENTIALS_PATH}`,
+        `${API_BASE_URL}${API_BUILD_DEBUG_CREDENTIAL_DATA_PATH}`,
```
Wynndow, line 31: Same here with the logging

src/app/oauth2/middleware.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
 
       logger.info("calling session start lambda", { req, res });
       const response = await axios.post(
-        `${API_BASE_URL}/session/start`,
+        `${API_BASE_URL}/session/initialise`,
```
Wynndow, line 38: Here too

src/app/credential-issuer/middleware.js
dannyhayes-gds: Yep good shout

dannyhayes-gds:
Forgot about the unit test as well. Will fix that with these changes

Wynndow (approved):

PR #392, opened 2022-08-26
PYIC-1870: Update core-front to have new callback endpoint with criId as a url path parameter

dannyhayes-gds:
### What changed
Setup a new callback endpoint for returning from a CRI that uses a URL path parameter for the criId instead of a query parameter.

### Why did it change
This new endpoint will be used to handle the redirect back from the app system. We have found that using a query param for the criId makes it slightly more complicated for a CRI to construct the redirect uri, so using a URL path parameter is a better alternative.

There will be future work to migrate the other CRI's to use this new endpoint as well.

### Issue tracking

- [PYIC-1870](https://govukverify.atlassian.net/browse/PYIC-1870)

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -18,13 +18,16 @@ module.exports = {
     next();
   },
   sendParamsToAPI: async (req, res, next) => {
+    const { id } = req.query;
+
+    const criId = req.params.criId || id;
+    const redirectUri = ["dcmaw, stubDcmaw"].includes(criId)
+      ? `${EXTERNAL_WEBSITE_HOST}/credential-issuer/callback/${criId}`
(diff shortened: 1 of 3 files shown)


Wynndow (dismissed):

src/app/credential-issuer/middleware.js
```diff
@@ -18,13 +18,16 @@ module.exports = {
     next();
   },
   sendParamsToAPI: async (req, res, next) => {
+    const { id } = req.query;
+
+    const criId = req.params.criId || id;
+    const redirectUri = ["dcmaw, stubDcmaw"].includes(criId)
```
dannyhayes-gds: I'm gonna change this bit slightly. I think it will be better if I just have 2 different `sendParamsToAPI` methods, one for each callback URL. This will make the future transitions of the other CRI's to start using the new callback URL a bit easier

Wynndow (dismissed):

Wynndow (approved):

PR #402, opened 2022-08-31
[PYIC-1867] Add Welsh translation yaml files

wilsond-gds:
## Proposed changes

This PR adds the Welsh translation of the English user interface content, ready for checking in early September.

### What changed

Welsh versions of the yaml files in `locales/en` were added to `locales/cy`. Pages can be viewed in Welsh by adding `?lang=cy` to the end of the URI.

### Why did it change

Welsh translation is required for the system to pass service evaluation.

### Issue tracking

- [PYIC-1867](https://govukverify.atlassian.net/browse/PYIC-1867)

### Other considerations

- [ ] Content will be reviewed in early September
- [ ] A variable will need to be added to sessions so the system is aware of the language the user has chosen earlier in the journey ([see RFC0002](https://github.com/alphagov/digital-identity-architecture/blob/785811fce3686677ddd20f4b5f3ccf39a25326ed/rfc/0002-crossteam-oidc-profile.md))

### DIFF
--- a/src/locales/cy/default.yml
+++ b/src/locales/cy/default.yml
@@ -1,0 +1,47 @@
+buttons:
+  authorizeAndReturn: Awdurdodi a Dychwelyd
+  next: Parhau
+  cancel: Canslo
+  credentialIssuer: Cyhoeddwr Cymhwyster
+  govUkHomepage: Ewch i hafan GOV.UK
+govuk:
+  serviceName: " "
+  backLink: Yn ôl
+  errorSummaryTitle: Mae problem
+  error: Gwall
+  warning: Rhybudd
+  skipLink: Neidio i’r prif gynnwys
+  betaBannerRequired: true
+  betaBannerContent:
+    betaBannerContentLabel: beta
+    betaBannerContentHTML: 'Mae hwn yn wasanaeth newydd – bydd eich <a class="govuk-link" rel="noopener" target="_blank" href="https://signin.account.gov.uk/contact-us">adborth (agor mewn tab newydd)</a> yn ein helpu i’w wella.'
+  cookie:
+    cookieBanner:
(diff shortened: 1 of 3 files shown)


LaiTang-GDS (approved):

src/locales/cy/default.yml
```diff
@@ -0,0 +1,47 @@
...
+      title: Cwcis ar gyfrif GOV.UK
+      heading: Cwcis ar gyfrif GOV.UK
+      paragraph1: Rydym yn defnyddio rhai cwcis hanfodol i wneud i’r gwasanaeth hwn weithio.
+      paragraph2: Hoffem osod cwcis ychwanegol er mwyn i ni allu cofio eich gosodiadau, deall sut rydych yn defnyddio’r gwasanaeth a gwneud gwelliannau.
+      changeCookiePreferencesLink: newid eich gosodiadau cwcis
+      cookieBannerAccept:
+        paragraph1: "Rydych wedi derbyn cwcis ychwanegol. Gallwch "
+        paragraph2: " unrhyw bryd."
+      cookieBannerReject:
+        paragraph1: "Rydych wedi gwrthod cwcis ychwanegol. Gallwch "
```
blakeyp, line 29: Just curious do some of the lines have to be wrapped in quote marks like this and others not?

wilsond-gds: hello Patrick - good question, in this case I think it's to force the content into a string for later processing -  this content is concatenated with a link in between the text. 

Sometimes quoting is also necessary to escape strings `paragraph1: "I have a 'single quote' in this sentence"` or to encourage transformation into HTML correctly - sometimes you'll see code like `    - "## What you can do"` which creates a `h2` instead of the default paragraph tag (or even worse, a `h2` wrapped in a paragraph tag).

PR #431, opened 2022-09-16
PYIC-1970: create app return and choose kbv route page

wilsond-gds:
## Proposed changes

Code to build the English version of the page presented to users who do not want to use the app route. The content has been sent for Welsh translation. The page renders correctly but hasn't yet been tested to ensure it returns the correct information to core back.

### What changed

Changes to the middleware to allow the page to be rendered; HTML file created which renders gov.uk components, and yaml file updated to hold the new content. 

### Why did it change

This page provides a route to users who do not wish to complete verification using the app journey

### Issue tracking

- [PYIC-1970](https://govukverify.atlassian.net/browse/PYIC-1970)

### DIFF
--- a/src/views/ipv/page-passport-doc-check.html
+++ b/src/views/ipv/page-passport-doc-check.html
@@ -1,0 +1,45 @@
+{% extends "shared/ipv-template.html" %}
+{% set hmpoPageKey = "pagePassportDocCheck" %}
+
+
+{% block cta %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+<form id="nextForm" action="/ipv/page/{{pageId}}" method="POST" onsubmit="return nextFormSubmit()">
+  <input type="hidden" name="_csrf" value="{{csrfToken}}">
+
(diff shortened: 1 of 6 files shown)


src/views/ipv/page-passport-doc-check.html
```diff
@@ -0,0 +1,43 @@
+{% extends "shared/ipv-template.html" %}
+{% set hmpoPageKey = "pagePassportDocCheck" %}
+
+{% block cta %}
+  {% from "govuk/components/button/macro.njk" import govukButton %}
+  {% from "govuk/components/radios/macro.njk" import govukRadios %}
+
+  <form id="passportDocCheckingForm" action="ipv/page/{{pageId}}" method="POST" onsubmit="return passportDocCheckingFormSubmit()">
```
dannyhayes-gds: Does this need a "/" at the start of the action? I spotted that we do in one of the njk files, but not 100% sure if its needed

LaiTang-GDS: I've changed it to be consistent with the other njk files

dannyhayes-gds (approved):

PR #492, opened 2022-10-21
PYIC-1078 set loadbalancer to use protected subnets which connects to the nat gateway

vam-gds:
### What changed
PYIC-1078 set loadbalancer to use protected subnets which connects to the nat gateway

### Why did it change
PYIC-1078 set loadbalancer to use protected subnets which connects to the nat gateway

### Issue tracking

### Other considerations


TobySaundersGDS (approved):
Worth a go - if this doesn't do it then VPC link maybe the way - I can have a dig into what's done in the dev platform vpc

PR #493, opened 2022-10-21
BAU: harmonise query string field name with cookie name for localisation

wilsond-gds:
## Proposed changes

This PR harmonises the query string field name to `lng`, which matches the cookie name `lng` to control the language of the user interface’s content.

### What changed

`app.js` has a new key added in the translation object.

### Why did it change

In the current state of the app, the language can be changed using the `lang` query string field name, which is different to the cookie name, `lng`. Using the same name reduces confusion and will make documenting this feature easier.

### Issue tracking

- This will be discussed further at the #di-frontend meeting

### Other considerations

- [ ] This behaviour should be documented as part of the Welsh translation work

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -54,6 +54,7 @@ const { router } = setup({
     allowedLangs: ["en", "cy"],
     fallbackLang: ["en"],
     cookie: { name: "lng" },
+    query: "lng",
   },
   dev: true,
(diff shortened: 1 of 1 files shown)


wilsond-gds:
This pull request was superseded by the migration to the non-hmpo version of the app - see `src/config/i18next.js`, lines 13 and 14.

PR #506, opened 2022-10-26
Added new alarm for External Gateway

TobySaundersGDS:
Added cloudwatch alarms  for load balancer and target group 5xx errors

## Proposed changes
This add a new cloudwatch alarm to trigger if there are load balancer, or ttarget group 5xx errors. This is configured per environment in a mapping

https://govukverify.atlassian.net/browse/PYIC-2091


TobySaundersGDS:
This is not ready - we need to enable cloudwatch metrics on the gateway first.

wpo-gds (approved):

PR #522, opened 2022-11-02
PYIC-2077: Get IP Address from Forwarded Header and set in session header

Danomanic:
### What changed
- Get IP Address from `forwarded` and set in session header for each Axios call.

### Issue tracking

- [PYIC-2077](https://govukverify.atlassian.net/browse/PYIC-2077)

### Other considerations

### DIFF
--- a/src/app/oauth2/middleware.js
+++ b/src/app/oauth2/middleware.js
@@ -15,6 +15,16 @@ module.exports = {
     next();
   },
 
+  setIpAddress: (req, res, next) => {
+    if (req.headers && req.headers["forwarded"]) {
+      const ipAddress = req.headers["forwarded"].split(";")[0].split("=")[1];
+      req.session.ipAddress = ipAddress || "unknown";
+    } else {
+      req.session.ipAddress = "unknown";
+    }
+    next();
+  },
+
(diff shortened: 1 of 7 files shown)


src/app/oauth2/middleware.js
```diff
@@ -15,6 +15,12 @@ module.exports = {
     next();
   },
 
+  setIpAddress: (req, res, next) => {
+    const ipAddress = req.headers["forwarded"].split(";")[0].split("=")[1];
+    req.session.ipAddress = ipAddress ? ipAddress : "unknown";
```
blakeyp: More idiomatic for js would be `ipAddress || "unknown"` 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation

blakeyp (commented):
To avoid any confusion could you update the PR description to reference the "forwarded" header not "x-forwarded-for" now?

src/app/oauth2/middleware.js
Danomanic: Good spot, fixed.

blakeyp (approved):

PR #601, opened 2022-12-13
Set evaluation periods and datapoints for alarms to 2

TobySaundersGDS:
This should prevent false positives from one session creating many 500 errors.


vam-gds (approved):
:+1:

PR #630, opened 2023-01-11
PYIC-2308: Handle new error type responses

AmritSidhu:
### What changed

Added an additional error handler to handle new error responses 

It checks to see if the error has come with a page, which it then sets the provided statusCode and renders the provided page 

Tests have been added for the all the handlers which was previously missing

### Why did it change

To reduce the proportion of journeys ending in an error and better handle unexpected journeys. 

### Issue tracking

- [PYIC-2308](https://govukverify.atlassian.net/browse/PYIC-2308)

[PYIC-2308]: https://govukverify.atlassian.net/browse/PYIC-2308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/journey-event-error-handler.js
+++ b/src/handlers/journey-event-error-handler.js
@@ -1,3 +1,5 @@
+const sanitize = require("sanitize-filename");
+
 module.exports = {
   journeyEventErrorHandler(err, req, res, next) {
     if (res.headersSent) {
@@ -8,7 +10,7 @@ module.exports = {
 
     if (res.err?.response?.data?.page) {
       res.status(res.err.response.data.statusCode);
-      return res.render(`ipv/${res.err.response.data.page}.njk`);
+      return res.render(`ipv/${sanitize(res.err.response.data.page)}.njk`);
     }
 
     next(err);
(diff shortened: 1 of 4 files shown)


src/handlers/journey-event-error-handler.js
```diff
@@ -0,0 +1,16 @@
...
+  journeyEventErrorHandler(err, req, res, next) {
+    if (res.headersSent) {
+      return next(err);
+    }
+
+    res.err = err; // this is required so that the pino logger does not log new error with a different stack trace
+
+    if (res.err?.response?.data?.page) {
+      res.status(res.err.response.data.statusCode);
+      return res.render(`ipv/${res.err.response.data.page}.njk`);
```
blakeyp: Feel like we should sanitise this input in some way? not sure what res.render does itself if anything

AmritSidhu: Yeah we sanitize the res.render input elsewhere so i'll look at adding that here to

blakeyp (approved):

PR #644, opened 2023-01-18
PYIC-2345 Remove obsolete front-end code

wilsond-gds:
## Proposed changes

Our front-end HTML source code attempts to serve assets to support very old browsers that don’t have modern CSS support or support for HTML5 block level elements such as `<nav>` or `<footer>`. This change removes this code to reduce complexity and increase performance.

Our current code is also wrapped in Internet Explorer specific conditional comments which are no longer necessary as all browsers we target will be required to parse the single file served. At the same time, the code that is served to modern browsers to fetch the stylesheet does not currently indicate which `media` the stylesheet supports, so this change adds the `media` attribute to the stylesheet `<link>`.

### What changed

In the global `base.njk` file the references to an IE specific stylesheet and the [HTML5 shim/shiv](https://github.com/aFarkas/html5shiv) have been removed. The `link` to the stylesheet is no longer surrounded by [conditional comments](https://www.htmldog.com/guides/html/advanced/conditionalcomments/) and has the [`media="all"` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media) added, to indicate to user agents that this stylesheet is applicable to print and screen styles. The code has been tested and print styles are applied if a user wishes to print out a page.

The nunjucks code also has dashes added to relevant tags to trim whitespace in the generated HTML source code.

### Why did it change

This will reduce the complexity of our `<head>` code and remove calls to files that are not part of our build pipeline.

#### Data on browser usage related to this change

Examining the statistics available shows that there is a very tiny percentage (0.06%) of users visiting DI with Internet Explorer 11, and no users at all visiting the site with IE10 or below. This follows trends seen elsewhere on GOV.UK. The situation will be monitored as the DI service is rolled out.

### Issue tracking

- [PYIC-2345](https://govukverify.atlassian.net/browse/PYIC-2345)

[PYIC-2345]: https://govukverify.atlassian.net/browse/PYIC-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assets/javascript/application.js
+++ b/src/assets/javascript/application.js
@@ -4,7 +4,7 @@
 // based on the code at
 // https://github.com/alphagov/di-authentication-frontend/blob/main/src/assets/javascript/cookies.js
 
-var cookies = function(trackingId, analyticsCookieDomain, journeyState) {
+var cookies = function cookies(trackingId, analyticsCookieDomain, journeyState) {
   var COOKIES_PREFERENCES_SET = "cookies_preferences_set";
   var cookiesAccepted = document.querySelector("#cookies-accepted");
   var cookiesRejected = document.querySelector("#cookies-rejected");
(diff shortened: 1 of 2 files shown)


kr8n3r:
> **no users at all visiting the site with IE10 or below.**

Would that be because there is ES6 syntax in your cookies code?

wilsond-gds:
Thanks @kr8n3r you’re right - the code is being uglified but not transpiled. We are seeing very very low numbers of IE11 users so it’d be surprising (but definitely not impossible) that there are some IE10 and lower users trying to use the service. I'll rewrite our JS code to use ES5 (and start investigating transpiling) and see if that changes the data we’re getting.

PR #689, opened 2023-02-16
BAU: Improve test coverage

AmritSidhu:
### What changed

By using the sonar test coverage you can see which lines in the code arent being covered. Tests have been added to hit these. 
Some tests where previously not working or producing false positive results which have now been rectified. 

### Why did it change

When fixing the sonarcloud dashboard connection it was discovered there where a few false positive tests along with gaps in the testing.


wilsond-gds (approved):
explained via video call; these tests remove some false positives and add coverage to important functionality in core-front.

PR #710, opened 2023-03-09
BAU NODE_ENV param added to deployment template

vam-gds:
### What changed
BAU NODE_ENV param added to deployment template

	modified:   deploy/template.yaml

### Why did it change
BAU NODE_ENV param added to deployment template

	modified:   deploy/template.yaml

### Issue tracking

### Other considerations

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -61,6 +61,7 @@ Mappings:
       fargateCPUsize: "256"
       fargateRAMsize: "512"
       nodeOldSpaceLimit: "384"
+      nodeEnv: "development"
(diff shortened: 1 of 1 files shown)


wilsond-gds (approved):
The aim of this change is so that we now have `NODE_ENV` set to `local` for running on a developer’s machine, `development` for Development environment, and `production` for build and higher environments. `production` should improve performance [by a factor of three](https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production:~:text=Setting%20NODE_ENV%20to,factor%20of%20three!) according to the express docs.

PR #712, opened 2023-03-10
PYIC-2537 Explicitly set cookie domain

wilsond-gds:
## Proposed changes

This commit adds the `cookieDomain` to the internationalisation code so that the cookie domain for the `lng` cookie can be categorically set.

### What changed

* `src/config/i18n.js` now sets `cookieDomain` to the value in the `process.env.ANALYTICS_DOMAIN` variable, or falls back to `.account.gov.uk`
* `.env.sample` has the `SESSION_SECRET`, `GTM_ID` and `ANALYTICS_DOMAIN`  constants added to ensure cookies are set correctly in local developer environments. Without setting these variables it will be impossible to change between languages correctly on the front-end as the JavaScript will not be able to set the `lng` cookie correctly. 

### Why did it change

Making these changes means that if the environment sets the `ANALYTICS_DOMAIN` then this setting is reflected in the node code which places the language cookie on the correct domain level, instead of setting it to a level which is too specific for CRIs to access.

### Issue tracking

- [PYIC-2537](https://govukverify.atlassian.net/browse/PYIC-2537)

[PYIC-2537]: https://govukverify.atlassian.net/browse/PYIC-2537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.sample
```diff
@@ -3,3 +3,6 @@ PORT=
 EXTERNAL_WEBSITE_HOST=
 BASE_URL=http://localhost:8080
 NODE_ENV=local
+SESSION_SECRET=secret123
+GTM_ID=GTM-TK92W68
```
AmritSidhu: We dont need to be putting the real values in the sample.env

AmritSidhu (approved):

chris-dco (approved):

PR #726, opened 2023-03-22
Adding definition of RUM to get snippet for adding to template

TobySaundersGDS:
This defines the RUM parameters to generate the snippet

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -870,6 +870,34 @@ Resources:
             Period: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, lambdaInvokeCompareWindow ]
             Stat: Sum
 
+
+  RUMAppMonitor:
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
wilsond-gds, line 1: I think ```Domain: !If [ IsProduction, "account.gov.uk", !Sub "${Environment}.account.gov.uk" ]``` might be better with a . in front of the account.gov.uk? that's what I think we need to do for cookies?

TobySaundersGDS: This one is without cookies - it may not do anything as that's the case - if it does impact the cookie I will update it.

wilsond-gds: yes, I've had a proper read and I think the dot issue is for cookie domains only. It's quite confusing when the spec and the actual behaviour don’t seem to match...

wilsond-gds (approved):

PR #754, opened 2023-04-06
[PYIC-2580] Allow core front to request callback with no session id

lwebb-iig:
Allow core front to call the cri callback without an ipvSessionId

### What changed

Middleware for cri updated to remove ipvSessionId null checks

### Why did it change

Support for mobile browser app change

### Issue tracking

- [PYIC-2580](https://govukverify.atlassian.net/browse/PYIC-2580)

### Other considerations

[PYIC-2580]: https://govukverify.atlassian.net/browse/PYIC-2580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kerrr:
I think this is overly broad. In most circumstances we still want frontend to check that we have an ipvSessionId. Perhaps we should have a way of specifying routes where the ipvSessionId check is not mandatory?

lwebb-iig:
> I think this is overly broad. In most circumstances we still want frontend to check that we have an ipvSessionId. Perhaps we should have a way of specifying routes where the ipvSessionId check is not mandatory?

Have limited this to only the CRI callback for now, as this seems the simplest way to achieve this

src/app/credential-issuer/middleware.js
```diff
@@ -32,15 +27,6 @@ module.exports = {
     }
 
     try {
-      if (!req.session?.ipvSessionId) {
```
AmritSidhu, line 35: I think we should still have some sort of validation incase neither ipvSessionId and the new clientOauthSessionId dont exist, if thats a scenario?

AmritSidhu (approved):

PR #783, opened 2023-04-21
RUM values conditional on non-dev environment

TobySaundersGDS:
made the RUM values not apply in the dev account as not relevant/available

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -62,7 +62,6 @@ Mappings:
       nodeOldSpaceLimit: "384"
       nodeEnv: "development"
       desiredTaskCount: 1
-      assetsCloudfrontArn: ""
     "457601271792": # Build
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
@@ -78,7 +77,6 @@ Mappings:
       nodeOldSpaceLimit: "1896"
       nodeEnv: "production"
       desiredTaskCount: 2
-      assetsCloudfrontArn: "somearn"
     "335257547869": # Staging
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
@@ -90,7 +88,6 @@ Mappings:
       nodeOldSpaceLimit: "256"
       nodeEnv: "production"
       desiredTaskCount: 1
-      assetsCloudfrontArn: "somearn"
     "991138514218": # Integration
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -63,7 +62,7 @@ Mappings:
       nodeOldSpaceLimit: "384"
       nodeEnv: "development"
       desiredTaskCount: 1
-      rumAppId: ""
+      assetsCloudfrontArn: ""
```
therealslimsladey: why the parameter name change?

deploy/template.yaml
```diff
@@ -448,14 +446,17 @@ Resources:
...
-                Fn::ImportValue: !Sub "AppMonitorId-${Environment}"
+              Value: !If [ IsNotDevelopment,
+                  Fn::ImportValue: !Sub "AppMonitorId-${Environment}",
+                  !Ref AWS::NoValue ]
             - Name: RUM_ROLE_ARN
-              Value:
-                Fn::ImportValue: !Sub "CognitoRoleARN-${Environment}"
+              Value: !If [ IsNotDevelopment,
+                  Fn::ImportValue: !Sub "CognitoRoleARN-${Environment}",
+                  !Ref AWS::NoValue ]
```
therealslimsladey, line 450: Presumably CognitoRoleARN-{env} and CognitoPoolID-{env} are also non-dev exports?

deploy/template.yaml
```diff
@@ -771,6 +775,14 @@ Resources:
             Period: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, tg500ErrorWindow ]
             Stat: Sum
 
+  #AssetsUploadInvoke:
```
therealslimsladey: Why's this commented out? Wouldn't deleting it be preferable?

deploy/template.yaml
TobySaundersGDS: I removed this - it was some WIP that I commited

TobySaundersGDS: The rumAppId is now an import rather than a map

deploy/template.yaml
TobySaundersGDS: yes - I've removed as it's not ready - will copy back in next week

therealslimsladey (approved):
looks good to me

deploy/template.yaml
TobySaundersGDS: yes

PR #833, opened 2023-05-19
don't use AWS::NoValue in env vars as removes var from definition

TobySaundersGDS:
The environment var being undefined was causing issues with the assets path. This defines it as an empty string to prevent the issue.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -460,24 +460,24 @@ Resources:
             - Name: RUM_APPLICATION_ID
               Value: !If [ IsNotDevelopment,
                   Fn::ImportValue: !Sub "AppMonitorId-${Environment}",
-                  !Ref AWS::NoValue ]
+                  ""]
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -459,20 +459,20 @@ Resources:
               Value: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, nodeEnv ]
             - Name: RUM_APPLICATION_ID
               Value: !If [ IsNotDevelopment,
-                  Fn::ImportValue: !Sub "AppMonitorId-${Environment}",
-                  !Ref AWS::NoValue ]
+                   Fn::ImportValue: !Sub "AppMonitorId-${Environment}",
```
AmritSidhu: will this complain about template format issues with the extra indent?

AmritSidhu (approved):

PR #842, opened 2023-05-23
Turn down the scaling cooldown period to scale faster

TobySaundersGDS:
This setting makes the ECS service wait for less time to bring in more tasks


TalatAydinGDS (approved):
LGTM

PR #843, opened 2023-05-23
Further tweaks for performance testing

TobySaundersGDS:
let ECS service bring in *5 number of targets
keep targets registered with load balancer for longer set separate scale in and out policies

## Proposed changes
Set scaling policy to handle scale out events faster and scale in more slowly

### What changed
Different policies for each type of scaling

### Why did it change
Improve the FE handling of spiky traffic

### Issue tracking
PYIC-1811

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -305,6 +305,9 @@ Resources:
       HealthCheckEnabled: TRUE
       HealthCheckProtocol: HTTP
       HealthCheckPath: /healthcheck
+      HealthCheckTimeoutSeconds: 2
+      HealthCheckIntervalSeconds: 5
+      HealthyThresholdCount: 2
       Matcher:
         HttpCode: 200
       Port: 80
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -440,6 +435,26 @@ Resources:
           - MetricIntervalLowerBound: 35
             ScalingAdjustment: 500
 
+  CoreFrontStepScaleInPolicy:
+    DependsOn: CoreFrontAutoScalingTarget
+    Type: AWS::ApplicationAutoScaling::ScalingPolicy
+    Properties:
+      PolicyName: CoreFrontStepScalingPolicy
```
lwebb-iig: Should we update this name property on both of the resources ie `CoreFrontStepScaleInPolicy`/`CoreFrontStepScaleOutPolicy` respectively?

TobySaundersGDS: yes - will do so

TobySaundersGDS: Gave the policies different names

lwebb-iig (approved):

PR #967, opened 2023-07-18
PYIC-2873 Show date of birth on ipv-reuse in GDS standard format

wilsond-gds:
## Proposed changes

This change shows birth dates in the [standard GDS style guide format](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates) which are both easier to read and are easier to parse when announced by a screenreader.

### What changed

A new nunjucks filter for dates is available, called `GDSDate`, and there is an updated sample date that tests a leap year is formatted correctly.

### Why did it change

Using this format makes it easier for users to identify if we have recorded their data correctly: see date example screenshot which transforms `1984-02-29` into `29 February 1984`.
Screenshot 2023-07-19 at 11 07 12

### Issue tracking

- [PYIC-2873](https://govukverify.atlassian.net/browse/PYIC-2873)

[PYIC-2873]: https://govukverify.atlassian.net/browse/PYIC-2873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/nunjucks.js
+++ b/src/config/nunjucks.js
@@ -14,6 +14,19 @@ module.exports = {
       return translate(key, options);
     });
 
+    nunjucksEnv.addFilter("GDSDate", function (formatDate) {
+      let dateTransform = new Date(formatDate);
+      let dateFormat = "en-GB"; // only using 'en' uses American month-first date formatting
+      if (this.ctx.i18n.language === "cy") {
+        dateFormat = "cy";
+      }
+      return dateTransform.toLocaleDateString(dateFormat, {
+        day: "numeric",
+        month: "long",
+        year: "numeric",
+      });
+    });
+
     return nunjucksEnv;
   },
 };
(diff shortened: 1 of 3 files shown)


shivanshuit914 (approved):

src/config/nunjucks.js
```diff
@@ -14,6 +14,19 @@ module.exports = {
       return translate(key, options);
     });
 
+    nunjucksEnv.addFilter("GDSDate", function (formatDate) {
+      let dateTransform = new Date(formatDate);
```
blakeyp, line 18: Very minor but best practice to use `const` if you don't need to reassign the var

blakeyp (approved):

src/config/nunjucks.js
wilsond-gds: Ah yes, I can see `dateFormat` could change but `dateTransform` doesn’t – I probably got confused because I’m using a method on `dateTransform`. I'll merge this and then update the code for the second part of `ipv-reuse`.

PR #1012, opened 2023-08-14
PYIC-3353: add security headers response

shivanshuit914:
### Issue tracking

- [PYIC-3353](https://govukverify.atlassian.net/browse/PYIC-3353)

### Other considerations

[PYIC-3353]: https://govukverify.atlassian.net/browse/PYIC-3353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/security-headers-handler.js
+++ b/src/handlers/security-headers-handler.js
@@ -2,7 +2,8 @@ module.exports = {
   securityHeadersHandler: (req, res, next) => {
     res.set({
       "Strict-Transport-Security": "max-age=31536000",
-      "Content-Security-Policy": "default-src 'self';",
+      "Content-Security-Policy":
+        "default-src 'self' identity.gov.uk *.identity.gov.uk",
       "X-Frame-Options": "DENY",
       "X-XSS-Protection": "0",
       "X-Content-Type-Options": "nosniff",
--- a/src/handlers/security-headers-handler.js
+++ b/src/handlers/security-headers-handler.js
@@ -4,7 +4,7 @@ module.exports = {
       "Strict-Transport-Security": "max-age=31536000",
       "Content-Security-Policy": "default-src 'self';",
       "X-Frame-Options": "DENY",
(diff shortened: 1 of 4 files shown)


src/handlers/security-headers-handler.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  securityHeadersHandler: (req, res, next) => {
+    res.set({
+      "Strict-Transport-Security": "max-age=31536000",
+      "Content-Security-Policy": "default-src 'self';",
```
blakeyp: Thinking this might block assets if they're hosted on a different domain?

src/handlers/security-headers-handler.js
```diff
@@ -0,0 +1,13 @@
...
+      "X-Frame-Options": "DENY",
+      "X-XSS-Protection": "0",
```
blakeyp, line 8: I think this should be 1 to enable it?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

shivanshuit914: I added 1 initially but in doc it says to set 0.

blakeyp: You're right
> The X XSS Protection header is designed to detect and assist in  protecting against cross-site scripting (“XSS”) attacks. However,  most modern browsers ignore this header now as having it on  can introduce additional security issues. The header should  therefore be set to off.

src/handlers/security-headers-handler.js
blakeyp: Confirmed with Toby we should add `*.identity.gov.uk` here so the assets don't get blocked on build and above where the assets are externally hosted

TobySaundersGDS (approved):

PR #1043, opened 2023-08-31
PYIC-3477: Update Content-Security-Policy to use nonce

Joe-Edwards-GDS:
### What changed

This change updates the CSP to follow the guidelines at https://csp.withgoogle.com/docs/strict-csp.html, i.e. using a nonce applied to all scripts, and more permissive fallbacks for older browsers.

### Why did it change

A previous change introduced a Content-Security-Policy with the value `default-src 'self'`, which blocks all inline scripts. This includes tracking scripts and various others.

### Issue tracking

- [PYIC-3477](https://govukverify.atlassian.net/browse/PYIC-3477)

### Other considerations

An alternative solution would be to add `unsafe-inline` (without `strict-dynamic`) to the CSP, allowing all inline scripts to run - but this significantly reduces the value of using a CSP.

[PYIC-3477]: https://govukverify.atlassian.net/browse/PYIC-3477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/locals.js
+++ b/src/lib/locals.js
@@ -7,9 +7,9 @@ const {
 const { generateNonce } = require("./strings");
 
 module.exports = {
-  getGTM: function (req, res, next) {
+  setLocals: function (req, res, next) {
     res.locals.gtmId = GTM_ID;
-    res.locals.scriptNonce = generateNonce();
+    res.locals.cspNonce = generateNonce();
     res.locals.analyticsCookieDomain = GTM_ANALYTICS_COOKIE_DOMAIN;
     res.locals.assetsCdnPath = CDN_PATH;
     res.locals.assetPath = CDN_DOMAIN + "/assets";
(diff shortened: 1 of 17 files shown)


src/lib/locals.js
```diff
@@ -7,9 +7,9 @@ const {
 const { generateNonce } = require("./strings");
 
 module.exports = {
-  getGTM: function (req, res, next) {
+  setLocals: function (req, res, next) {
     res.locals.gtmId = GTM_ID;
-    res.locals.scriptNonce = generateNonce();
+    res.locals.cspNonce = generateNonce();
```
Joe-Edwards-GDS, line 12: This value is used by the `govuk-frontend` template, and must be called `cspNonce` to allow scripts in the template to work properly.

src/views/ipv/page-f2f-multiple-doc-check.njk
Joe-Edwards-GDS, line 1: There are two changes to each of these files:
- Add the nonce to the `<script>` tag
- Move the submit handler into the script tag

Both of these count as inline javascript (and the easiest solution to the latter is to move it into a script).

Wynndow (approved):
Looks like it does the stuff we talked about

PR #1047, opened 2023-09-05
PYIC-3497 Handle Cimit Escape action form events

wilsond-gds:
## Proposed changes

To ensure the correct values are sent to core-back this PR intercepts the form submission for `pyi-suggest-other-options` and `pyi-suggest-other-options-no-f2f` to rewrite the `handleJourneyResponse` request. The parameters sent in `pyi-suggest-other-options-no-f2f` have been changed to match those used on other pages.

### What changed

The `router.js` file and the `middleware.js` files have a new `handleCimitEscapeAction` listener added, and the `router.js` file has specific `post` behaviour defined for the two CIMIT-related pages.

### Why did it change

These changes ensure `core-front` is sending the right information to `core-back` for processing.

### Issue tracking

- [PYIC-3497](https://govukverify.atlassian.net/browse/PYIC-3497)

[PYIC-3497]: https://govukverify.atlassian.net/browse/PYIC-3497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -50,12 +50,6 @@ router.post(
   handleCriEscapeAction
 );
 router.post(
-  "/page/pyi-suggest-other-options",
-  parseForm,
-  csrfProtection,
-  handleCimitEscapeAction
-);
-router.post(
   "/page/pyi-suggest-other-options-no-f2f",
   parseForm,
   csrfProtection,
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -50,12 +50,11 @@ router.post(
   handleCriEscapeAction
 );
 router.post(
-  "/page/pyi-suggest-other-options-no-f2f",
+  "/page/pyi-suggest-other-options",
   parseForm,
   csrfProtection,
   handleCimitEscapeAction
 );
-
 router.post("/page/:pageId", parseForm, csrfProtection, handleJourneyAction);
 router.get("/*", updateJourneyState);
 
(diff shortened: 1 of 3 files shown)


src/app/ipv/router.js
```diff
@@ -48,6 +49,19 @@ router.post(
   csrfProtection,
   handleCriEscapeAction
 );
+router.post(
+  "/page/pyi-suggest-other-options",
+  parseForm,
+  csrfProtection,
+  handleCimitEscapeAction
+);
+router.post(
```
AmritSidhu: You dont need this for this page, this can be handled with the next/end events. If you take a look at pyi-escape-page & pyi-escape-no-f2f it will make a bit more sense. We can handle a next & end events in core-back as standard. With the other page where its handling more than those two possibilities it requires the code you've added

src/views/ipv/pyi-suggest-other-options-no-f2f.njk
```diff
@@ -24,13 +24,13 @@
     name: "journey",
     items: [
               {
-                value: "next",
+                value: "next/dcmaw",
```
AmritSidhu: Based on my previous comment this can stay as just next & end

src/app/ipv/router.js
wilsond-gds: That makes sense - I've removed this code and also reverted the change to the nunjucks file for `pyi-suggest-other-options-no-f2f.njk`

AmritSidhu (approved):

PR #1048, opened 2023-09-05
BAU: Validate JSON files as part of yarn lint

wilsond-gds:
## Proposed changes

As part of the `yarn lint` pre-commit step, these changes check the `translation.json` files are valid.

### What changed

`package.json` and `.eslintrc.js` now use a new npm package, [eslint-plugin-jsonc](https://www.npmjs.com/package/eslint-plugin-jsonc).

### Why did it change

Some errors had crept into the translation files over time, so this change checks that a valid file structure is preserved as different team members work on the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,9 @@
   "devDependencies": {
     "chai": "4.3.7",
     "chai-as-promised": "7.1.1",
-    "eslint": "8.46.0",
+    "eslint": "^8.46.0",
     "eslint-config-prettier": "8.8.0",
+    "eslint-plugin-jsonc": "^2.9.0",
     "eslint-plugin-prettier": "4.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,8 +35,9 @@
   "devDependencies": {
     "chai": "4.3.7",
     "chai-as-promised": "7.1.1",
-    "eslint": "8.46.0",
+    "eslint": "^8.46.0",
```
merlinc: Can you reinstall these packages with an exact version? All the other packages in this project are installed that way.

Joe-Edwards-GDS:
Closing stale PR. If we see further inconsistencies in the translation files, we should write a test that checks the structure matches what we expect (not just that it's valid JSON).

PR #1084, opened 2023-09-26
PYIC-3591: Set up catalogue of core-front pages

AmritSidhu:
### What changed

Set up catalogue of core-front pages

Use the NODE_ENV to determine whether we are in a non-production environment to display the new route
Repurposed the existing unused data we already had defined
Moved the reuse code into the helper as it is now used in two places
Screenshot 2023-09-28 at 09 49 19

### Why did it change

We’d like devs/testers/UCD/any non-technical stakeholders to be able to view a catalogue of all the pages served by core-front, labelled by their path slug/file name with each page clickable to view. This is also important so we develop a common language around page names with UCD

### Issue tracking

- [PYIC-3591](https://govukverify.atlassian.net/browse/PYIC-3591)

- [x] No environment variables or secrets were added or changes

[PYIC-3591]: https://govukverify.atlassian.net/browse/PYIC-3591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -195,7 +195,10 @@ module.exports = {
       const currentEnvironment = process.env.NODE_ENV;
       const { pageId } = req.params;
 
-      if (currentEnvironment === "development") {
+      if (
+        currentEnvironment === "development" &&
+        req.session.visitedAllTemplates
+      ) {
         if (pageId === "page-ipv-reuse") {
           let userDetailsResponse = samplePersistedUserDetails;
           const i18n = req.i18n;
@@ -412,6 +415,8 @@ module.exports = {
         const templatesWithoutExtension = files.map(
           (file) => path.parse(file).name
         );
+
+        req.session.visitedAllTemplates = true;
 
         res.render("ipv/all-templates.njk", {
           allTemplates: templatesWithoutExtension,
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -3,6 +3,7 @@ const sanitize = require("sanitize-filename");
 const {
   API_BASE_URL,
   API_BUILD_PROVEN_USER_IDENTITY_DETAILS,
+  DEVELOPMENT_ENVIRONMENT,
 } = require("../../lib/config");
 const {
   buildCredentialIssuerRedirectURL,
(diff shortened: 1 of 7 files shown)


src/app/ipv/router.js
```diff
@@ -27,6 +28,11 @@
   next();
 }
 
+const currentEnvironment = process.env.NODE_ENV;
+if (currentEnvironment === "development") {
+  router.get("/all-templates", allTemplates);
```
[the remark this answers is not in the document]
AmritSidhu: Dismissed as this route is only accessible in dev/build

src/views/ipv/all-templates.njk
```diff
@@ -0,0 +1,24 @@
+{% extends "shared/base.njk" %}
+{% set pageTitleName = "All templates" %}
+
+{% block content %}
+  <table class="govuk-table">
+    <caption class="govuk-table__caption govuk-table__caption--l"></caption>
+    <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">List of all pages in di-ipv-core-front</th>
+        <th scope="col" class="govuk-table__header"></th>
```
RogVanGDS: Is it worth having this as a larger header than the content rows below? (Know that this isn't user facing, but might be better for readability nonetheless

src/app/ipv/middleware.js
```diff
@@ -187,7 +192,31 @@ module.exports = {
   },
   handleJourneyPage: async (req, res, next) => {
     try {
+      const currentEnvironment = process.env.NODE_ENV;
       const { pageId } = req.params;
+
+      if (
+        currentEnvironment === "development" &&
```
RogVanGDS: Might be worth having "development" as a constant or an Enum value to avoid accidental changes in the future?

src/app/ipv/middleware.js
```diff
@@ -187,7 +192,31 @@ module.exports = {
...
+        req.session.visitedAllTemplates
+      ) {
+        if (pageId === "page-ipv-reuse") {
+          let userDetailsResponse = samplePersistedUserDetails;
```
RogVanGDS: Does this need to be `let`? (I may have missed the reassignment in all fairness)

src/app/ipv/middleware.js
```diff
@@ -187,7 +192,31 @@ module.exports = {
...
+          const i18n = req.i18n;
+          const userDetails = generateUserDetails(userDetailsResponse, i18n);
```
RogVanGDS: Do we lose much readability by just having `const userDetails = generateUserDetails(samplePersistedUserDetails, req.i18n);` here and removing the above two lines?

RogVanGDS (dismissed):
Looks good to me, feel free to ignore me being a nit-pick if you want 🙂

RogVanGDS (approved):
LGTM

PR #1102, opened 2023-10-09
PYIC-3540: use render instead of redirect to handle technical error page

shivanshuit914:
## Proposed changes

- Replace technical error page redirection with render method.

### What changed

- Replaced redirection with page render method.

### Why did it change

- At the moment front is not sending correct http status code to Analytics tools, because of redirections front is sending 302 and 200 status code instead of error codes like (400, 500 etc)

### Issue tracking

- [PYIC-3540](https://govukverify.atlassian.net/browse/PYIC-3540)

### Environment variables or secrets

- [x] No environment variables or secrets were added or changed

- [ ] Documented in the [README](./blob/main/README.md)
- [ ] Added to deployment repository
- [ ] Added to local startup repository

### Other considerations

- [ ] Update [README](./blob/main/README.md) with any new instructions or tasks

[PYIC-3540]: https://govukverify.atlassian.net/browse/PYIC-3540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1104, opened 2023-10-09
BAU: Save session before internal redirect

Wynndow:
### What changed

Save session before internal redirect

### Why did it change

While getting core-back and core-front to run locally it was noticed that it's possible for core-front to issue a redirect, and it be followed, before the session has finished writing. This resulted in weird errors where the expected page didn't match the request page.

Whilst we've not knowingly seen this happen in production, probably due to the longer request times involved, it's a possibility and we should address it.

This updates core front to issue the redirect in the callback function of the session.save() method. This will mean the redirect will only be issued once the session is successfully stored.

This has only been applied to internal redirects - redirects that take the user to a different page on core front. We don't need to worry about this issue if we're sending the user off to do something elsewhere.

I had initially worried that implementing this would cause the users session to be saved twice - once when we explicitly call it with the changes, and once when the request-response cycle ends and express saves it automatically. I've dug through express-sessions and the lib does a check of the hash of the current state of the session against the hash the last time it saved it, and doesn't save again if they match. So that's good.

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -36,6 +36,7 @@ const axios = require("axios");
 const { getIpAddress } = require("../shared/ipAddressHelper");
 const fs = require("fs");
 const path = require("path");
+const { saveSessionAndRedirect } = require("../shared/redirectHelper");
 
 async function journeyApi(action, req) {
   if (action.startsWith("/")) {
@@ -119,7 +120,11 @@ async function handleBackendResponse(req, res, backendResponse) {
       requestId: req.requestId,
     });
     req.session.currentPage = backendResponse.page;
-    return res.redirect(`/ipv/page/${backendResponse.page}`);
+    return await saveSessionAndRedirect(
+      req,
+      res,
+      `/ipv/page/${backendResponse.page}`
+    );
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/app/ipv/middleware.js
```diff
@@ -245,7 +250,11 @@ module.exports = {
         );
 
         req.session.currentPage = "pyi-attempt-recovery";
-        return res.redirect(req.session.currentPage);
+        return await saveSessionAndRedirect(
+          req,
+          res,
+          `/ipv/page/pyi-attempt-recovery`
```
Joe-Edwards-GDS, line 256: This might conflict with https://github.com/govuk-one-login/ipv-core-front/pull/1102/files - once that's been merged, you might find you can simplify this, as we'll only redirect in a couple of places

cc @shivanshuit914

Joe-Edwards-GDS (approved):
Will any of the Selenium tests check the URL for error pages? Might be worth keeping an eye on the tests just in case

Wynndow:
@Joe-Edwards-GDS I don't think so? Shiv's changes seem to have been applied successfully. This change won't make any difference to the system from an external point of view (I hope), so not expecting any tests to fail because of it.

PR #1160, opened 2023-11-13
PYIC-3429: Add validation to radio buttons

AmritSidhu:
### What changed

Added radio button form validation to all core-front pages which have a radio button forms
 - Added nunjucks filter to push more keys into an object  
 - Added a check function to determine whether journey is undefined and set an errorState
 
Added routes for each individual form POST which has the new form validation check
Used the latest translations provided by content for the error messages

Regarding `page-ipv-bank-account-start.njk form validation` I have added a placeholder message which is from the other error messages, this will need to be updated when we get finalised content

### Why did it change

Users where previously able to submit a form without selecting an option which would default to using the first item in the form. 

### Issue tracking

- [PYIC-3429](https://govukverify.atlassian.net/browse/PYIC-3429)

[PYIC-3429]: https://govukverify.atlassian.net/browse/PYIC-3429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -283,6 +283,7 @@ module.exports = {
         case "pyi-technical-unrecoverable":
           return res.render(`ipv/${sanitize(pageId)}.njk`, {
             pageId,
+            pageErrorState: req.query.errorState,
             csrfToken: req.csrfToken(),
           });
         case "page-ipv-reuse": {
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -281,11 +281,15 @@ module.exports = {
         case "pyi-f2f-technical":
         case "pyi-technical":
         case "pyi-technical-unrecoverable":
-          return res.render(`ipv/${sanitize(pageId)}.njk`, {
+          const renderOptions = {
             pageId,
-            pageErrorState: req.query.errorState,
             csrfToken: req.csrfToken(),
-          });
+          };
+          if (req.query?.errorState !== undefined) {
+            renderOptions.pageErrorState = req.query.errorState;
+          }
+
+          return res.render(`ipv/${sanitize(pageId)}.njk`, renderOptions);
         case "page-ipv-reuse": {
           const userDetailsResponse = await axios.get(
             `${API_BASE_URL}${API_BUILD_PROVEN_USER_IDENTITY_DETAILS}`,
@@ -443,10 +447,27 @@ module.exports = {
     }
   },
   formRadioButtonChecked: async (req, res, next) => {
-    if (req.method === "POST" && req.body.journey === undefined) {
-      return res.redirect(req.originalUrl + "?errorState=true");
-    } else {
+    const allowedPathRule = { pattern: /^\/ipv\/page\/[^\/]+$/, methods: ['GET', 'POST'] };
+
+    const isPathAllowed = (url, method) => {
+      const path = new URL(url, `https://${req.headers.host}`).pathname.toLowerCase();
+      if (!method || allowedPathRule.methods.includes(method)) {
+        if (path.match(allowedPathRule.pattern)) {
+          return true;
+        }
+        throw new Error('Path is not allowed');
+      }
+      return false;
+    };
+    try {
+      if (req.method === 'POST' && req.body.journey === undefined && isPathAllowed(req.originalUrl, req.method)) {
+        const redirectUrl = `${req.baseUrl}${req.path}?errorState=true`;
+        return res.redirect(redirectUrl);
+      }
+
       next();
+    } catch (error) {
+      next(error);
     }
   },
   validateFeatureSet: async (req, res, next) => {
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -449,35 +449,16 @@ module.exports = {
     }
   },
   formRadioButtonChecked: async (req, res, next) => {
-    const allowedPathRule = {
-      pattern: /^\/ipv\/page\/[^/]+$/,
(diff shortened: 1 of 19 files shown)


src/views/ipv/page-ipv-identity-document-start.njk
```diff
@@ -1,72 +1,100 @@
...
+{% set errorState = pageErrorState %}
 {% set pageTitleName = 'pages.pageIpvIdentityDocumentStart.title' | translate %}
 {% set googleTagManagerPageId = "pageIpvIdentityDocumentStart" %}
 
 {% block content %}
-  <h1 class="govuk-heading-l" id="header" data-page="{{googleTagManagerPageId}}">{{'pages.pageIpvIdentityDocumentStart.header' | translate }}</h1>
-  <p class="govuk-body">{{'pages.pageIpvIdentityDocumentStart.content.paragraph' | translate }}</p>
+
+  {% if errorState %}
+  <p class="govuk-body">{{ 'pages.pageIpvIdentityDocumentStart.content.paragraph' | translate }}</p>	    {{ govukErrorSummary({
```
DanCorderIPV: Don't think this \<p> should be here

src/app/ipv/middleware.js
```diff
@@ -441,6 +448,40 @@ module.exports = {
...
+      return false;
+    };
+    try {
+      if (
+        req.method === "POST" &&
+        req.body.journey === undefined &&
+        isPathAllowed(req.originalUrl, req.method)
+      ) {
+        const redirectUrl = `${req.baseUrl}${req.path}?errorState=true`;
+        return res.redirect(redirectUrl);
```
DanCorderIPV: It doesn't seem right to redirect the user to another URL - can't we just show the error message on the page we are going to render?

src/views/ipv/page-f2f-multiple-doc-check.njk
```diff
@@ -2,53 +2,76 @@
...
 {% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% set errorState = pageErrorState %}
 {% set pageTitleName = 'pages.pageF2FMultipleDockCheck.title' | translate %}
 {% set googleTagManagerPageId = "pageF2FMultipleDockCheck" %}
 
 {% block content %}
-  <h1 class="govuk-heading-l" id="header" data-page="{{googleTagManagerPageId}}">{{'pages.pageF2FMultipleDockCheck.header' | translate }}</h1>
-  <p class="govuk-body">{{'pages.pageF2FMultipleDockCheck.content.paragraph1' | translate | safe }}</p>
+  {% if errorState %}
```
DanCorderIPV: Can we move this block to `base.njk` and somehow pass it the text entries to display?
Maybe in the validation check create a list of errors (currently maximum length of 1 :) ) that contain the relevant bits of text and then have the shared template display the error box if there are any validation errors.

DanCorderIPV:
Did you consider using something like https://express-validator.github.io/docs/ ? From Googling around that seems to be the most common way of adding some validation to an Express app.
Not sure what the exact rules are on adding more libraries/dependencies to the project, but we are using Express already.

src/views/ipv/page-f2f-multiple-doc-check.njk
AmritSidhu: Yes this can be moved into the base.njk, however I think it would still be best to keep the errors in the internationalisation files for consistency and makes it easier to maintain. What I can do is move it into the base and then in the html template set the necessary fields. This would reduce it to half the amount of lines in each template

AmritSidhu:
> Did you consider using something like https://express-validator.github.io/docs/ ? From Googling around that seems to be the most common way of adding some validation to an Express app. Not sure what the exact rules are on adding more libraries/dependencies to the project, but we are using Express already.

No longer necessary with using render over the redirect, we can just define and sanitize the url like we have done elsewhere

src/app/ipv/middleware.test.js
```diff
@@ -735,4 +735,47 @@ describe("journey middleware", () => {
...
+      };
+    });
+
+    it("should render if method is POST, journey is not defined", async function () {
+      req.body.journey = undefined;
+      req.method = "POST";
+      await middleware.formRadioButtonChecked(req, res, next);
+
+      expect(res.render).to.not.have.been.called;
+      expect(next).to.have.been.calledOnce;
```
blakeyp: Shouldn't this be the opposite?
render called and next not called

blakeyp: Ah I think the `called` needs to be called 😅  i.e. `expect(res.render).to.have.been.called()` ?

src/views/ipv/page-f2f-multiple-doc-check.njk
```diff
@@ -5,50 +5,64 @@
...
+        {
+          value: "end",
+          text: 'pages.pageF2FMultipleDockCheck.content.formRadioButtons.otherWayButtonText' | translate
+        }
+      ]
+    } %}
+
+    {% if errorState %}
+      {% set errorMessageObject = { 'text': 'pages.pageF2FMultipleDockCheck.content.formErrorMessage.errorRadioMessage' | translate } %}
+      {% set radiosConfig = radiosConfig | setAttribute('errorMessage', errorMessageObject) %}
```
blakeyp, line 60: Nice ⚡

blakeyp (approved):

PR #1164, opened 2023-11-15
PYIC-3860: Propagated reset user details event to backend 

RogVanGDS:
### What changed

Propagated reset user details event from frontend to backend

### Why did it change

Added so that we can complete the user flow for delete user details and trigger the delete-details lambda.

### Issue tracking

- [PYIC-3860](https://govukverify.atlassian.net/browse/PYIC-3860)

[PYIC-3860]: https://govukverify.atlassian.net/browse/PYIC-3860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -254,6 +254,9 @@ module.exports = {
       }
 
       switch (pageId) {
+        case "pyi-new-details":
+        case "pyi-confirm-delete-details":
+        case "pyi-details-deleted":
         case "page-ipv-identity-document-start":
         case "page-ipv-identity-postoffice-start":
         case "page-ipv-bank-account-start":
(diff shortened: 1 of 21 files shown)


src/app/ipv/middleware.js
```diff
@@ -389,6 +392,18 @@ module.exports = {
       next(error);
     }
   },
+  handleConfirmDeleteDetailsAction: async (req, res, next) => {
+    try {
+      if (req.body?.journey === "reset") {
+        await handleJourneyResponse(req, res, "journey/reset");
```
AmritSidhu: this can just be journey/next as there are only two options meaning we can reuse the `handleJourneyAction`. The subsequent change will need to be made in the journey map to use next. We've done that elsewhere for example on the `PYI_F2F_TECHNICAL`

RogVanGDS: Ah yep good spot

blakeyp:
Could you just pop a title on the PR please

blakeyp (approved):

PR #1177, opened 2023-11-29
Pyic 3884: Added dynamic content framework.

RogVanGDS:
### What changed

Added dynamic content framework.

Added dynamic translation for pyi-suggest-other-options

### Why did it change

So in future we can use more dynamic content in pages, instead of having to add a new page every time.

### Issue tracking

- [PYIC-3884](https://govukverify.atlassian.net/browse/PYIC-3884)

[PYIC-3884]: https://govukverify.atlassian.net/browse/PYIC-3884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV (dismissed):

src/views/ipv/pyi-suggest-other-options.njk
```diff
@@ -8,13 +8,14 @@
 {% set errorTitle = 'pages.pyiSuggestOtherOptions.content.formErrorMessage.errorSummaryTitleText' | translate %}
 {% set errorText = 'pages.pyiSuggestOtherOptions.content.formErrorMessage.errorSummaryDescriptionText' | translate %}
 {% set errorHref = "#pyiSuggestOtherOptionsForm" %}
+{% set pageContext = context %}
```
blakeyp: Is this required, we can't pass in context directly?

src/locales/en/translation.json
```diff
@@ -185,6 +185,7 @@
         "paragraph1": "This is because you did not answer some of the security questions correctly.",
         "subHeading": "What you can do",
         "paragraph2": "Based on what you’ve told us, you have a photo ID you can use to prove your identity.",
+        "paragraph2NoPhotoId": "If you have a photo ID instead you may be able to use it to prove your identity",
```
blakeyp: I would add `TEMP` here too as this isn't finalised UCD content

src/app/shared/loggerHelper.js
```diff
@@ -23,11 +23,15 @@ module.exports = {
...
+    { logCommunicationType, type, path, info, context = "" },
+  ) => {
     const message = {
       logCommunicationType,
       path,
       info,
+      context,
     };
     if (type) {
       message.type = type;
```
blakeyp, line 37: Is there a way we can include context in the request/response logging too? I think we might be able to if we store context in res.locals similar to sessionId - see: https://github.com/govuk-one-login/ipv-core-front/blob/eac294d0f93f75afcb2280f99d25fd92aeea54e3/src/lib/logger.js#L22

blakeyp (approved):

PR #1189, opened 2023-12-07
PYIC-4149: remove yarn and use npm for build

shivanshuit914:
## Proposed changes

- Remove yarn and use NPM.

### What changed

- Removed yarn set up and replaced with NPM in package file.
- Replaced build and tests pipeline steps using NPM.
- Replaced docker yarn steps to NPM.
- Updated readme file.

### Why did it change

- NPM is more standard and widely used compare to yarn.

### Issue tracking

- [PYIC-4149](https://govukverify.atlassian.net/browse/PYIC-4149)

[PYIC-4149]: https://govukverify.atlassian.net/browse/PYIC-4149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -1,24 +1,22 @@
...
-COPY yarn.lock ./
+COPY package-lock.json ./
 
-RUN yarn install
-RUN yarn build
+RUN npm install
+RUN npm run build
 
-# 'yarn install --production' does not prune test packages which are necessary
+# 'npm install --production' does not prune test packages which are necessary
```
Joe-Edwards-GDS: npm actually can do this - `npm install --omit=dev` will prune dev dependencies :)

shivanshuit914: aah thanks

shivanshuit914: Updated it :)

Dockerfile
```diff
@@ -1,24 +1,22 @@
...
-COPY yarn.lock ./
+COPY package-lock.json ./
 
-RUN yarn install
-RUN yarn build
+RUN npm install
+RUN npm run build
 
-# 'yarn install --production' does not prune test packages which are necessary
+# 'npm install --omit=dev' does not prune test packages which are necessary
```
Joe-Edwards-GDS, line 10: Sorry, I meant that this command _does_ prune dev dependencies - so you don't need to delete `node_modules` explicitly.

package.json
```diff
@@ -5,13 +5,13 @@
   "main": "src/server.js",
   "engines": {
     "node": "18.x",
-    "yarn": "1.22.x"
+    "npm": "9.5.x"
```
Joe-Edwards-GDS: This feels quite restrictive - I've got `npm` version 9.6.7, and the latest is 10.2.5

Perhaps just `9.x` if you're worried about breaking changes? If anyone wants to use 10.x then they can test it and add it to the specification :)

Joe-Edwards-GDS: At some point we should upgrade to node 20 as well (it's now LTS), but that's a separate change!

shivanshuit914: Yes makes sense, will update it to `9.x` thanks

Dockerfile
shivanshuit914: Done

Joe-Edwards-GDS (approved):

PR #1191, opened 2023-12-07
PYIC-4151: Bug fix - double-submit prevention now working for all forms

AmritSidhu:
### What changed

Updated the form check to look for all forms 

### Why did it change

Double-submit prevention is not working, allowing users in some browsers (e.g. Firefox) to submit twice from pages in ipv core.

### Issue tracking

- [PYIC-4151](https://govukverify.atlassian.net/browse/PYIC-4151)

[PYIC-4151]: https://govukverify.atlassian.net/browse/PYIC-4151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assets/javascript/app/init.js
+++ b/src/assets/javascript/app/init.js
@@ -14,13 +14,12 @@ window.DI = window.DI || {};
 // Globally disable second clicks on the continue button for radio button pages,
 // But allow onSubmit scripts to overwrite this global behaviour
 
-const checkForOnSubmit = document.querySelector("form[onsubmit]");
-const checkForRadio = document.querySelector('input[type="radio"]');
 let disableSubmit = false;
 
-if (checkForOnSubmit === null && checkForRadio !== null) {
-  const manageFormClicks = document.querySelector("form");
-  manageFormClicks.addEventListener("submit", () => {
+const manageFormClicks = document.querySelectorAll("form");
+
+manageFormClicks.forEach((form) => {
+  form.addEventListener("submit", () => {
     if (!disableSubmit) {
       disableSubmit = true;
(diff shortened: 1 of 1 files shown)


src/assets/javascript/app/init.js
```diff
@@ -18,7 +18,7 @@ const checkForOnSubmit = document.querySelector("form[onsubmit]");
 const checkForRadio = document.querySelector('input[type="radio"]');
 let disableSubmit = false;
 
-if (checkForOnSubmit === null && checkForRadio !== null) {
+if (checkForOnSubmit === null || checkForRadio !== null) {
```
Wynndow: Could we just drop this if statement completely? And let the JS manage the submit button on any form that's found?

src/assets/javascript/app/init.js
```diff
@@ -14,18 +14,17 @@ window.DI = window.DI || {};
 // Globally disable second clicks on the continue button for radio button pages,
 // But allow onSubmit scripts to overwrite this global behaviour
 
-const checkForOnSubmit = document.querySelector("form[onsubmit]");
```
Joe-Edwards-GDS, line 17: Were there any forms that did this? I can't think why they would, but someone obviously thought we needed it originally!

Joe-Edwards-GDS (approved):

src/assets/javascript/app/init.js
AmritSidhu: It looks like two forms use onSubmit, not sure why though, might be some old lingering code

PR #1196, opened 2023-12-11
PYIC-4125 Increase health-check timeout

Joe-Edwards-GDS:
In the high-volume reuse load-test, tasks are being killed due to health-check timeouts when under load. This means the service is stuck in a task recycle loop without being able to scale up sufficiently to handle the volume.


TobySaundersGDS (approved):
Looks sensible

blakeyp (approved):

PR #1206, opened 2023-12-19
PYIC-4125: Optimise core-front

MikeCollingwood:
### What changed

- Reduced Target tracking autoscaling CPU threshold
- Added in nunjuck caching
- Remove an unused helper file

### Why did it change

- To reduce CPU overuse, the major issue spotted with stress test results

### Issue tracking

- [PYIC-4125](https://govukverify.atlassian.net/browse/PYIC-4125)

### Env vars

- [ ] Added in processing for TEMPLATE_CACHING boolean - which should be set to true for production at least

[PYIC-4125]: https://govukverify.atlassian.net/browse/PYIC-4125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config/nunjucks.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     const nunjucksEnv = nunjucks.configure(viewsPath, {
       autoescape: true,
       express: app,
-      noCache: true,
+      noCache: process.env.NODE_ENV !== "production",
```
Joe-Edwards-GDS: Makes sense - it doesn't look like this was a calculated decision to have enabled in prod, as far as I can tell.

One note of caution - by tying to `NODE_ENV`, it's not possible to override this on a per-environment basis. For example, it'll be enabled in all environments together. In general it's nicer to have explicit config for this sort of thing - `TEMPLATE_CACHING` or something.

deploy/template.yaml
```diff
@@ -487,7 +487,7 @@ Resources:
       TargetTrackingScalingPolicyConfiguration:
         PredefinedMetricSpecification:
           PredefinedMetricType: ECSServiceAverageCPUUtilization
-        TargetValue: 60
+        TargetValue: 50
```
Joe-Edwards-GDS, line 496: We might want to tune this back up once we've got a passing test, but should be safe enough for now.

If we reach a steady-state of 50%, we're paying for twice as much capacity as we're actually using! I assume our production usage is currently well below this for now anyway?

src/lib/strings.js
Joe-Edwards-GDS, line 1: Why is this deleted - isn't it used in `locals.js`?

deploy/template.yaml
MikeCollingwood: That makes sense, if it looks good then I will revert 👍

src/lib/strings.js
MikeCollingwood: Ah, I rely on the IDE too much for this sort of logic. VS Code didn't show usages for this method on clicking on it... will be more careful next time

Joe-Edwards-GDS: Slightly concerning that this isn't being linked up properly - are usages working properly elsewhere?

MikeCollingwood: It works for the exported logging methods, but trying the same on generateNonce yields nothing.... (???)
Screenshot 2024-01-02 at 11 23 22
Screenshot 2024-01-02 at 11 23 55

Right well after repeatedly clicking it, it now works. It was tempermentally taking me there, but now it works most of the time!? Is it VS code

MikeCollingwood: Found an issue which sums it up: https://github.com/microsoft/vscode/issues/117642. I'll ask for a WebStorm license when Chris Clayson's back

src/config/nunjucks.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     const nunjucksEnv = nunjucks.configure(viewsPath, {
       autoescape: true,
       express: app,
-      noCache: true,
+      noCache: !process.env.TEMPLATE_CACHING,
```
Joe-Edwards-GDS: I think you probably need to add this to
- `config.js` (so all the config can be found in one place)
- `template.yaml` (so it's set appropriately in the deployed environments)

src/lib/strings.js
Joe-Edwards-GDS: Occasionally it'll be dodgy if you're flicking through branches - it'll cache some of the references and you need to revisit the file (or reload the project) for it to notice. 🤷

That issue sounds very specific, so hard to know if it's actually related.

src/config/nunjucks.js
MikeCollingwood: Ah, ok I didn't realise environment variables could be set in template.yaml, will make edits 👍

src/lib/strings.js
MikeCollingwood: Could imagine it's caching but while I was repeatedly asking for references I didn't revisit the file and at times it'd alternate in giving references or none 🤔

deploy/template.yaml
```diff
@@ -65,6 +65,7 @@ Mappings:
       assetsDomain: ""
       assetsPath: ""
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
+      templateCaching: false
```
Joe-Edwards-GDS: I think this can be true in all deployed environments - just not local (especially if you might be load testing against dev!)

I was actually thinking potentially the opposite - if you think there are any potential concerns about this, you could leave it off in staging & production. Should be safe enough if you've tested yourself...

MikeCollingwood: I was thinking if it's being worked on (and tested in dev envs) and nunjucks templates are getting edited regularly then it could be irritating - but is it being worked on locally rather than in the dev env? (A question for Hilux possibly)

I can make an edit for my dev, but I guess you're right that it would be more wanted if we're expecting people to perf test in dev more often

Joe-Edwards-GDS (approved):
Looks good (bearing in mind the code freeze runs until tomorrow)

PR #1207, opened 2023-12-19
PYIC-4141: Add option to set page context on all-templates page.

DanCorderIPV:
### What changed

There is now an option on the all-templates page to set a `context` that is used by some pages to change what they display.

Additional refactorings:
- To allow capturing of the context the all-templates page has switched to being a form
- The all-templates code has been moved into a separate folder with router and middleware files. (With the added benefit that the all-templates file no longer shows up on the all-templates page)
- The new dev router is only instantiated in the app when `ENABLE_PREVIEW` is set. This should make it easier to keep dev features disabled in production.
- The all-templates page has moved `/ipv/all-templates` to `/dev/all-templates` to make it clearer that it is a development tool.

### Why did it change

Some pages now display differently depending on a supplied `context` value. We would like to be able to see all the different versions of pages from the `all-templates` page.

### Issue tracking

- [PYIC-4141](https://govukverify.atlassian.net/browse/PYIC-4141)

- [ x ] No environment variables or secrets were added or changed

[PYIC-4141]: https://govukverify.atlassian.net/browse/PYIC-4141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app.js
```diff
@@ -165,6 +166,9 @@ router.use((req, res, next) => {
 router.use("/oauth2", require("./app/oauth2/router"));
 router.use("/credential-issuer", require("./app/credential-issuer/router"));
 router.use("/ipv", require("./app/ipv/router"));
+if (ENABLE_PREVIEW) {
+  router.use("/dev", require("./app/development/router"));
```
DanCorderIPV, line 170: This change of URL may be controversial as we'll need to let the UX team know and they'll have to update bookmarks. We could revert this if required.

blakeyp: Might be best if we can avoid this - it will affect support folk that have been using this too
and I guess it will break the journey map links too?

RogVanGDS (approved):
LGTM

PR #1218, opened 2024-01-02
PYIC-4125: Keep HTTPS connections alive

MikeCollingwood:
### What changed

- Use HTTP keep alive connections

### Why did it change

- I/O performance optimisation

### Issue tracking

- [PYIC-4125](https://govukverify.atlassian.net/browse/PYIC-4125)

[PYIC-4125]: https://govukverify.atlassian.net/browse/PYIC-4125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/credential-issuer/middleware.js
+++ b/src/app/credential-issuer/middleware.js
@@ -1,22 +1,9 @@
-const {
-  API_BASE_URL,
-  API_CRI_CALLBACK,
-  EXTERNAL_WEBSITE_HOST,
-} = require("../../lib/config");
-const { generateJsonAxiosConfig } = require("../shared/axiosHelper");
+const { API_CRI_CALLBACK, EXTERNAL_WEBSITE_HOST } = require("../../lib/config");
 const { handleBackendResponse } = require("../ipv/middleware");
 const { logCoreBackCall, transformError } = require("../shared/loggerHelper");
 const { LOG_COMMUNICATION_TYPE_REQUEST } = require("../shared/loggerConstants");
-const http = require("http");
-const https = require("https");
-const axios = require("axios");
-
-const httpAgent = new http.Agent({ keepAlive: true });
-const httpsAgent = new https.Agent({ keepAlive: true });
-const axiosInstance = axios.create({
-  httpAgent,
-  httpsAgent,
-});
+const { CoreBackService } = require("../../services/coreBackService");
+const coreBackService = new CoreBackService();
 
 module.exports = {
   sendParamsToAPI: async (req, res, next) => {
@@ -41,10 +28,10 @@ module.exports = {
         path: API_CRI_CALLBACK,
       });
 
-      const apiResponse = await axiosInstance.post(
-        `${API_BASE_URL}${API_CRI_CALLBACK}`,
-        { ...body, ...errorDetails },
-        generateJsonAxiosConfig(req),
+      const apiResponse = await coreBackService.postCriCallback(
+        req,
+        body,
+        errorDetails,
       );
       if (apiResponse?.status) {
         res.status(apiResponse.status);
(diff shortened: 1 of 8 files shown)


src/app/credential-issuer/middleware.js
```diff
@@ -7,8 +7,17 @@ const { generateJsonAxiosConfig } = require("../shared/axiosHelper");
 const { handleBackendResponse } = require("../ipv/middleware");
 const { logCoreBackCall, transformError } = require("../shared/loggerHelper");
 const { LOG_COMMUNICATION_TYPE_REQUEST } = require("../shared/loggerConstants");
+const http = require("http");
+const https = require("https");
 const axios = require("axios");
 
+const httpAgent = new http.Agent({ keepAlive: true });
```
Joe-Edwards-GDS: Do we make any HTTP requests (non-HTTPS)?

src/app/credential-issuer/middleware.js
Joe-Edwards-GDS, line 1: I think it'd be good to test how much difference this makes, but if it's effective then there are probably a couple of enhancements we could make:
- There are other Axios requests being made in `oauth2/middleware.js`, which could benefit from the same change
- We could share an instance across all three, since they're calling the same API. I'd suggest pulling out a shared `coreBackService` or similar which has methods for each core back API we call. This also makes it easier to test :)

src/app/credential-issuer/middleware.js
MikeCollingwood: Good point!

src/services/coreBackService.js
```diff
@@ -0,0 +1,83 @@
+const {
+  API_BASE_URL,
+  API_BUILD_PROVEN_USER_IDENTITY_DETAILS,
+  API_CRI_CALLBACK,
+  API_SESSION_INITIALISE
+} = require("../lib/config");
+const https = require("https");
+const axios = require("axios");
+
+class CoreBackService {
```
Joe-Edwards-GDS: Does this need to be a class - I think we could just make this a module? (i.e. just export `postSessionInitialise`, `postCriCallback` and `getProvenIdentityDetails` directly)

The only instance variable is the axios instance, and I think we _do_ want to share it between different consumers. Plus it'll probably make jest mocking easier.

Joe-Edwards-GDS (commented):
Do we need to update tests as well?

src/services/coreBackService.js
```diff
@@ -0,0 +1,83 @@
...
+      `${API_BASE_URL}/${action}`,
+      {},
+      req.session?.clientOauthSessionId
+        ? this.generateAxiosConfigWithClientSessionId(req)
+        : this.generateAxiosConfig(req),
+    )
+  }
+
+  postSessionInitialise(req, authParams) { return this.axiosInstance.post(
+      `${API_BASE_URL}${API_SESSION_INITIALISE}`,
```
Joe-Edwards-GDS: Not necessary, but you can actually set a baseURL on the axios instance, which means you only need to specify the path part here

src/app/credential-issuer/middleware.js
```diff
@@ -1,13 +1,8 @@
...
-  API_CRI_CALLBACK,
-  EXTERNAL_WEBSITE_HOST,
-} = require("../../lib/config");
-const { generateJsonAxiosConfig } = require("../shared/axiosHelper");
+const { API_CRI_CALLBACK, EXTERNAL_WEBSITE_HOST } = require("../../lib/config");
 const { handleBackendResponse } = require("../ipv/middleware");
 const { logCoreBackCall, transformError } = require("../shared/loggerHelper");
 const { LOG_COMMUNICATION_TYPE_REQUEST } = require("../shared/loggerConstants");
-const axios = require("axios");
+const CoreBackService = require("../../services/coreBackService");
```
Joe-Edwards-GDS: I think I'd normally camelCase this (i.e. coreBackService), or even just destructure the imports directly?

Hopefully we'll move to TS and ES6 modules soon anyway...

MikeCollingwood: I was getting struggling to avoid errors when destructing. I'll change it to camelCase

Joe-Edwards-GDS (approved):

PR #1245, opened 2024-01-26
PYIC-4431: Make page pyi-technical dynamic

RogVanGDS:
### What changed

Made page pyi-technical dynamic

### Why did it change

Now that we have dynamic pages, we can combine some templates that are very similar to remove code duplication

### Issue tracking

- [PYIC-4431](https://govukverify.atlassian.net/browse/PYIC-4431)

[PYIC-4431]: https://govukverify.atlassian.net/browse/PYIC-4431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/ipv/middleware.js
```diff
@@ -185,7 +216,7 @@ module.exports = {
   },
   handleJourneyPage: async (req, res, next) => {
     try {
-      const { pageId } = req.params;
+      let { pageId } = req.params;
```
blakeyp: Any reason for changing to `let`? I can't see it being reassigned

src/locales/en/translation.json
```diff
@@ -274,7 +274,8 @@
       "content": {
         "paragraph1": "We cannot prove your identity right now.",
         "subHeading": "What you can do",
-        "paragraph2": "Continue to the service you were trying to use and look for other ways to prove your identity."
+        "paragraph2": "Continue to the service you were trying to use and look for other ways to prove your identity.",
+        "paragraph2Unrecoverable": "Go back to the service you were trying to use and start again. You can look for the service using your search engine or find it from the GOV.UK homepage."
```
blakeyp, line 278: cy translations file needs updating too?

src/locales/en/translation.json
```diff
@@ -274,7 +274,8 @@
...
       }
     },
     "pyiTechnicalUnrecoverable": {
```
blakeyp: Can we remove this now?

blakeyp (commented):
I'm guessing there's no change needed for core-back because it never returns `pyi-technical-unrecoverable` itself?

RogVanGDS:
> I'm guessing there's no change needed for core-back because it never returns `pyi-technical-unrecoverable` itself?

Yep, it doesn't look like we route from core-back to pyi-technical-unrecoverable at all, it's only used when catching front-end errors.

I did have a fail-safe routing option to route anything trying to reach `pyi-technical-unrecoverable` to `pyi-technical` with the appropriate context, but I don't think it's needed on second thought.

blakeyp (approved):

PR #1265, opened 2024-02-01
PYIC-4432: Refactored timeout screens to use context

RogVanGDS:
### What changed

Refactored timeout screens to use context

### Why did it change

Now that we have dynamic pages, we can condense similar view templates into one to avoid code duplication.

### Issue tracking

- [PYIC-4432](https://govukverify.atlassian.net/browse/PYIC-4432)

[PYIC-4432]: https://govukverify.atlassian.net/browse/PYIC-4432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/ipv/middleware.js
```diff
@@ -238,9 +238,9 @@ module.exports = {
         return res.render(`ipv/${req.session.currentPage}.njk`, {
           context: "unrecoverable",
         });
-      } else if (pageId === "pyi-timeout-unrecoverable") {
```
DanCorderIPV, line 241: As discussed, we probably need to leave this section in temporarily until the back end is updated to use the new target.

DanCorderIPV (changes requested):

src/app/ipv/middleware.js
```diff
@@ -238,9 +238,9 @@ module.exports = {
...
-        req.session.currentPage = "pyi-timeout-unrecoverable";
-        return res.render(`ipv/${req.session.currentPage}.njk`);
+      } else if (pageId === "pyi-timeout" && context === "unrecoverable") {
+        req.session.currentPage = "pyi-timeout";
+        return res.render(`ipv/${req.session.currentPage}.njk`, { context });
```
blakeyp, line 243: Maybe I'm not reading this right, but why can't this just be handled in the switch below which already passes context to the template?

RogVanGDS: This early escape route to the pyi-timeout-unrecoverable page has been there since I joined, so not sure on the original requirement to have this.

It looks like these early escapes before the switch statement are for handling forced redirects when there is a backend error, as opposed to the switch statement which handles the rendering from the usual journey-map routing

RogVanGDS: I think removing this early-escape, would result in the pyi-timeout-unrecoverable redirect being caught in the else if statement below, resorting in the wrong page being rendered

blakeyp: I see ok, thanks

PR #1280, opened 2024-02-07
PYIC-4613: Improve front-end logging 

RogVanGDS:
### What changed

Changed the structure of some of the front-end logs that get sent to cloudwatch. 

### Why did it change

This is to improve readability of the logs when debugging in cloudwatch/splunk and hide possibly sensitive information from being logged.

### Issue tracking

- [PYIC-4613](https://govukverify.atlassian.net/browse/PYIC-4613)

[PYIC-4613]: https://govukverify.atlassian.net/browse/PYIC-4613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -1,3 +1,4 @@
+/* istanbul ignore file */
 const fs = require("fs");
 const path = require("path");
 const sanitize = require("sanitize-filename");
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -1,4 +1,3 @@
-/* istanbul ignore file */
 const fs = require("fs");
 const path = require("path");
 const sanitize = require("sanitize-filename");
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -26,8 +26,8 @@ const axiosErrorHandler = (error) => {
       const message = {
         description: "Error response received in coreBackService",
         errorMessage: error.response.data,
-        endpoint: `${error.err?.config?.method} ${error.err?.config?.url}`,
-        status: error.err?.config?.status,
+        endpoint: `${error.request?.method} ${error.request?.url}`,
+        request: error.request,
         cri
       }
 
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -25,6 +25,7 @@ const axiosErrorHandler = (error) => {
         description: "Error response received in coreBackService",
         errorMessage: error.response.data,
         endpoint: `${error.request?.method} ${error.request?.path}`,
+        status: request.status
       };
 
       if (cri) {
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -5,7 +5,7 @@ const { getCriFromErrorResponse } = require("./loggerHelper");
 
 const createAxiosInstance = () => {
   const instance = axios.create({
-    baseURL: API_BASE_URL + "invalid",
+    baseURL: API_BASE_URL,
     httpsAgent: new https.Agent({ keepAlive: true }),
   });
 
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -27,7 +27,8 @@ const axiosErrorHandler = (error) => {
         description: "Error response received in coreBackService",
         errorMessage: error.response.data,
         endpoint: `${error.request?.method} ${error.request?.path}`,
-        cri
+        cri,
+        error: error.response
       }
 
       logger.error({ message, level: "ERROR" });
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -25,7 +25,7 @@ const axiosErrorHandler = (error) => {
         description: "Error response received in coreBackService",
         errorMessage: error.response.data,
         endpoint: `${error.request?.method} ${error.request?.path}`,
-        status: request.status
+        status: error.request.status
       };
 
       if (cri) {
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -13,7 +13,7 @@ const createAxiosInstance = (baseUrl) => {
   return instance;
 };
 
-const axiosErrorHandler = (error) => {
+const axiosErrorHandler = async (error) => {
   const logger = error.config.logger;
 
   if (axios.isAxiosError(error)) {
@@ -49,7 +49,7 @@ const axiosErrorHandler = (error) => {
     }
   }
 
-  return Promise.reject(error);
+  return error;
 };
 
 module.exports = {
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -1,7 +1,7 @@
 const axios = require("axios");
 const https = require("https");
 const { API_BASE_URL } = require("../../lib/config");
-const { getMiddlewareErrorHandlerMessage } = require("./loggerHelper");
+const { getCriFromErrorResponse } = require("./loggerHelper");
 
 const createAxiosInstance = (onRejected) => {
   const instance = axios.create({
@@ -19,7 +19,7 @@ const axiosErrorHandler = (error) => {
 
   if (axios.isAxiosError(error)) {
     if (error.response) {
-      const cri = getMiddlewareErrorHandlerMessage(
+      const cri = getCriFromErrorResponse(
         error.response
       );
 
@@ -34,6 +34,7 @@ const axiosErrorHandler = (error) => {
     } else if (error.request) {
       const message = {
         request: error.request,
+        error: error,
         description: "Error occured making request in coreBackService",
       };
 
@@ -74,7 +75,7 @@ module.exports = {
       headers: {
         "Content-Type": "application/x-www-form-urlencoded",
         "ipv-session-id": req.session?.ipvSessionId,
-        "client-session-id": req?.session?.clientOauthSessionId,
+        "client-session-id": req?.session?.clientOauthSessionId + "invalid",
         "x-request-id": req.id,
         "ip-address": req.session.ipAddress,
         "feature-set": req.session.featureSet,
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -25,7 +25,7 @@ const axiosErrorHandler = (error) => {
 
       const message = {
         description: "Error response received in coreBackService",
-        errorMessage: error.response,
+        errorMessage: error.response.data,
         endpoint: `${error.err?.config?.method} ${error.err?.config?.url}`,
         status: error.err?.config?.status,
         cri
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -35,7 +35,6 @@ const axiosErrorHandler = (error) => {
       logger.error({ message, level: "ERROR" });
     } else if (error.request) {
       const message = {
-        request: error.request,
         error: error,
         description: "Error occurred making request in coreBackService",
       };
--- a/src/app/shared/axiosHelper.js
+++ b/src/app/shared/axiosHelper.js
@@ -28,7 +28,6 @@ const axiosErrorHandler = (error) => {
         errorMessage: error.response.data,
         endpoint: `${error.request?.method} ${error.request?.path}`,
         cri,
-        error: error.response
       }
 
       logger.error({ message, level: "ERROR" });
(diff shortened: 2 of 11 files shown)


src/app/shared/loggerHelper.js
```diff
@@ -38,4 +39,24 @@ module.exports = {
     }
     req.log.info({ message, level: "INFO", requestId: req.id });
   },
+  getMiddlewareErrorHandlerMessage: (err, description) => {
+    const { config, request, response, ...desiredErrorProperties } = err;
+
+    const requestDataString = config?.data;
+
+    const credentialIssuerId =
+      requestDataString && JSON.parse(requestDataString)?.credentialIssuerId;
```
blakeyp: Do we know what would happen if JSON.parse() blew up (if `requestDataString` was invalid json for some reason)? Maybe wrap it to be super safe?

src/app/shared/loggerHelper.js
```diff
@@ -38,4 +39,24 @@ module.exports = {
...
+
+    const message = {
+      err: desiredErrorProperties,
+      response: err?.response?.data,
+      description: description,
```
blakeyp: I'm not sure how useful any of `err` is in the context of a core-back error response
```
err: {
       code: ERR_BAD_REQUEST
       config: {}
       message: Request failed with status code 400
       name: AxiosError
       stack: AxiosError: Request failed with status code 400
    at settle (/app/node_modules/axios/dist/node/axios.cjs:1967:12)
    at IncomingMessage.handleStreamEnd (/app/node_modules/axios/dist/node/axios.cjs:3066:11)
    at /opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2167:48
    at t.runInContext (/opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2278:24)
    at IncomingMessage.handleStreamEnd (/opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2167:18)
    at IncomingMessage.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at /opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2167:48
    at t.runInContext (/opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2278:24)
    at endReadableNT (/opt/dynatrace/oneagent/agent/bin/1.283.132.20240205-143805/any/nodejs/nodejsagent.js:2167:18)
    at Axios.request (/app/node_modules/axios/dist/node/axios.cjs:3877:41)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async sendParamsToAPI (/app/src/app/credential-issuer/middleware.js:30:27)
       status: 400
     }
```
I know people on call have seen this and based on the stack trace thought there was a dynatrace related issue 😅 

The response actually tells us what went wrong
```
response: {
       code: 1030
       journey: /journey/error
       message: Missing OAuth state in callback request
       statusCode: 400
     }
```

This certainly helps but as the ticket suggests maybe we could make core-back response logging part of the `coreBackService`?

blakeyp (commented):
This looks good but be good to get @Joe-Edwards-GDS's thoughts

src/app/shared/loggerHelper.js
```diff
@@ -38,4 +39,24 @@ module.exports = {
     }
     req.log.info({ message, level: "INFO", requestId: req.id });
   },
+  getMiddlewareErrorHandlerMessage: (err, description) => {
+    const { config, request, response, ...desiredErrorProperties } = err;
```
Joe-Edwards-GDS: The intent here is actually to strip out the `config`, `request` and `response` from the error? I think it might be worth saying that in a comment - it confused me a little to start with!

src/app/shared/loggerHelper.js
Joe-Edwards-GDS: If we catch Axios errors in the `coreBackService` then I think we can probably be more controlled about what's being logged and where

E.g. in `coreBackService` we could have something like https://axios-http.com/docs/handling_errors:
```
if (axios.isAxiosError(err)) {
  if (err.response) {
    // Error response from backend
    // Log useful request and response details - e.g. URL, status code, response body (if present)
  } else if (err.request) {
    // Other error making request (e.g. connection failed)
    // Log useful request details and error info - e.g. URL and error message
  } else {
    // Something went wrong setting up the request
    throw err;
  }
  throw new CoreBackError(err.response?.data?.page); // Throw error that the overall error handlers can use
}
```
You can use an axios 'interceptor' to do this across all API calls without needing to duplicate the logic.

Then in the overall error handlers you can identify `CoreBackError`s and handle them appropriately (safe in the knowledge that the API call details have been logged).

Joe-Edwards-GDS (commented):
I wouldn't object to this going in as-is and improved in a separate PR, but left some suggestions

src/app/shared/loggerHelper.js
RogVanGDS: It throws an error which is caught in the general error handling on the server

src/app/shared/axiosHelper.js
```diff
@@ -1,5 +1,56 @@
+const axios = require("axios");
+const https = require("https");
+const { API_BASE_URL } = require("../../lib/config");
```
Joe-Edwards-GDS: (minor, but I'd probably prefer to make this agnostic to this, and have it be passed as an argument to `createAxiosInstance` - what if we need to add another API with a different URL?)

src/app/shared/axiosHelper.js
```diff
@@ -1,5 +1,56 @@
...
+const { getMiddlewareErrorHandlerMessage } = require("./loggerHelper");
+
+const createAxiosInstance = (onRejected) => {
```
Joe-Edwards-GDS: I think `onRejected` is always `axiosErrorHandler` - does it need to be an argument?

src/app/shared/axiosHelper.js
```diff
@@ -1,5 +1,56 @@
...
+  }
+
+  return Promise.reject(error);
+};
+
 module.exports = {
+  createAxiosInstance,
+  axiosErrorHandler,
   generateAxiosConfig: (req) => {
+    const logger = req.log;
```
Joe-Edwards-GDS: (any reason you've declared a variable for this rather than inlining?)

src/app/shared/axiosHelper.js
```diff
@@ -1,5 +1,56 @@
...
+};
+
+const axiosErrorHandler = (error) => {
+  const logger = error.config.logger;
+
+  if (axios.isAxiosError(error)) {
+    if (error.response) {
+      const message = getMiddlewareErrorHandlerMessage(
+        error.response,
+        "Error response received in coreBackService",
```
Joe-Edwards-GDS: This is not (necessarily) true now you've pulled this into a separate module - I'd either move it back into `coreBackService` or make it more generic (e.g. 'Error response received from API')

src/app/shared/axiosHelper.js
```diff
@@ -1,5 +1,56 @@
...
+      );
+
+      logger.error({ message, level: "ERROR" });
+    } else if (error.request) {
+      const message = {
+        request: error.request,
+        description: "Error occured making request in coreBackService",
```
Joe-Edwards-GDS: I think this is losing the actual error message?

I think we can be specific about what you want to log here and above, since we know exactly what shape the error has. Perhaps something like:
```
if (axios.isAxiosError(error)) {
  const { request, response } = error;
  if (response) {
    const message = {
      description: "Error response received in coreBackService",
      errorMessage: error.message,
      endpoint: `${request.method} ${request.url}`,
      status: response.status,
      response: response.data,
    };

    logger.error({ message, level: "ERROR" });
  } else if (request) {
    const message = {
      description: "Error occured making request in coreBackService",
      errorMessage: error.message,
      endpoint: `${request.method} ${request.url}`,
    };

    logger.error({ message, level: "ERROR" });
  } else {
    const message = {
      description: "Error occurred setting up request in coreBackService",
      errorMessage: error.message,
    };

    logger.error({ message, level: "ERROR" });
  }
```
Is there anything else in the axios object that you think will be useful? If the url doesn't include the CRI ID, then it might also be worth pulling that explicitly out of the `params`.

Joe-Edwards-GDS (commented):
Can you give some example log messages that this produces? I think that's the most useful way to review this, in a way!

src/app/shared/axiosHelper.js
RogVanGDS: It does so we can test the onRejected handler logic in isolation

src/app/shared/axiosHelper.js
RogVanGDS: So that if it's undefined then it won't include it as a config field, rather than having logger: undefined in the config object

RogVanGDS:
Example error message: 
Screenshot 2024-02-22 at 10 46 56

src/app/development/middleware.js
```diff
@@ -1,3 +1,4 @@
+/* istanbul ignore file */
```
blakeyp: Is this for sonar? could we add it to the exclusions in `sonar-project.properties` instead

src/app/shared/axiosHelper.js
```diff
@@ -1,4 +1,60 @@
...
+      const message = {
+        error,
+        description: "Something went wrong setting up an API request",
+      };
+
+      logger.error({ message, level: "ERROR" });
+    }
+  }
+
+  return Promise.reject(error);
```
blakeyp, line 52: Does this need to be wrapped in a promise? I think if you make this an async function you can just return error and it will implicitly wrap in a promise

src/services/coreBackService.js
```diff
@@ -23,11 +25,13 @@ function postAction(req, action) {
 }
 
 function postSessionInitialise(req, authParams) {
+  const logger = req.log;
   return axiosInstance.post(API_SESSION_INITIALISE, authParams, {
     headers: {
       "ip-address": req.session.ipAddress,
       "feature-set": req.session.featureSet,
     },
+    logger,
```
blakeyp: What is the need to pass in the logger here but not in the other axios calls?

blakeyp (commented):
This is really nice 🚀

src/services/coreBackService.js
RogVanGDS: Think I just missed this when refactoring

src/app/shared/axiosHelper.js
RogVanGDS: Yeah it does unfortunately, otherwise the request times out

src/app/shared/axiosHelper.js
```diff
@@ -1,4 +1,60 @@
...
+
+      if (cri) {
+        message.cri = cri;
+      }
+
+      logger.error({ message, level: "ERROR" });
+    } else if (error.request) {
+      const message = {
+        error: error,
+        description: "Error occurred making request to API",
```
Joe-Edwards-GDS, line 38: Can we still include the endpoint here as well? It's less likely to be relevant, but might still be useful

src/app/shared/axiosHelper.js
```diff
@@ -1,4 +1,60 @@
...
+      };
+
+      logger.error({ message, level: "ERROR" });
+    } else {
+      const message = {
+        error,
```
Joe-Edwards-GDS, line 44: Minor detail, but this is slightly inconsistent with the structure in `internal-server-error-handler` - `error` vs `err`. Might be nice to make them the same?

src/app/shared/axiosHelper.js
```diff
@@ -1,4 +1,60 @@
...
+
+      const message = {
+        description: "Error response received from API",
+        errorMessage: error.message,
+        endpoint: `${error.request?.method} ${error.request?.path}`,
+        data: error.response.data,
+      };
+
+      if (cri) {
+        message.cri = cri;
```
Joe-Edwards-GDS, line 31: If you want a 'clever' one-liner, you can do:
```
description: "Error response received from API",
errorMessage: error.message,
endpoint: `${error.request?.method} ${error.request?.path}`,
data: error.response.data,
...(cri && { cri }),
```
...but I think the explicit `if` is a lot clearer!

Joe-Edwards-GDS (approved):
A couple of very minor comments that you can address if you like, but I think this is looking good, thanks for sorting it out!

src/app/shared/axiosHelper.js
blakeyp: I prefer the explicit if, more readable 😃

blakeyp (approved):

PR #1325, opened 2024-02-29
PYIC-4946: Implemented new GA4 library

RogVanGDS:
### What changed

Implemented new GA4 library node module and ripped out the old implementation.

### Why did it change

Moving to the new library to make it as easy as possible for the various pods that make up the One Login journey to upgrade their analytics while having as minimal an impact as possible on the dev teams time and effort.

We need to remove our current GA4 implementation to avoid duplicating logic.

### Issue tracking

- [PYIC-4946](https://govukverify.atlassian.net/browse/PYIC-4946)

### Other considerations

[PYIC-4946]: https://govukverify.atlassian.net/browse/PYIC-4946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/views/shared/banner.njk
```diff
@@ -76,5 +76,5 @@
       ],
       hidden: true
     }
-  ],  hidden: true
+  ],  hidden: false
```
blakeyp, line 79: was this intentionally changed?

src/views/shared/ga4/on-page-load-script.njk
```diff
@@ -0,0 +1,22 @@
...
+            .DI
+            .analyticsGa4
+            .pageViewTracker
+            .trackOnPageLoad({
+                language: document.querySelector("html")?.getAttribute("lang"),
+                location: document.location.href?.toLowerCase(),
+                englishPageTitle: '{{params.englishPageTitle}}',
+                referrer: document.referrer?.toLowerCase(),
+                content_id: "",
+                dynamic: '{{params.isPageDynamic}}' ?? false,
```
blakeyp: should this be wrapped in a string or can just pass the bool as is?

deploy/template.yaml
```diff
@@ -168,6 +178,8 @@ Mappings:
       nodeOldSpaceLimit: "1792"
       nodeEnv: "production"
       desiredTaskCount: 2
+      ga4Disabled: true
+      uaDisabled: true
```
blakeyp: Based on https://govukverify.atlassian.net/browse/DCP-3156?focusedCommentId=126209 this should initially be true, false in prod (and integration)?

src/views/shared/banner.njk
RogVanGDS: Yes, the banner needs to be displayed by default. The logic in the analytics package is to hide it if the user has opted out, not display it if the user has opted in so we want display by default.

At least that is my understanding and the logic is the same for the mobile-app but I suggest we test this before going live

src/views/shared/ga4/on-page-load-script.njk
RogVanGDS: We want to be passing in a boolean here

blakeyp: Sorry I mean the `'{{params.isPageDynamic}}'` is wrapped in `''` when it shouldn't be?

RogVanGDS: Ah of course 🤦

src/views/shared/ga4/on-page-load-script.njk
```diff
@@ -0,0 +1,22 @@
...
+            .DI
+            .analyticsGa4
+            .pageViewTracker
+            .trackOnPageLoad({
+                language: document.querySelector("html")?.getAttribute("lang"),
+                location: document.location.href?.toLowerCase(),
+                englishPageTitle: '{{ params.englishPageTitle }}',
+                referrer: document.referrer?.toLowerCase(),
+                content_id: "",
+                dynamic: '{{ params.isPageDynamic }}' === 'true',
```
RogVanGDS, line 13: Turns out it does need to be wrapped in quotation marks after all 🤷

shivanshuit914 (approved):

blakeyp (approved):

PR #1336, opened 2024-03-12
PYIC-5343: Override status code if provided

Wynndow:
### What changed

Override status code if provided

### Why did it change
In the case of an error page, core-back can provide a status code in the body of the response. Core-front should respect that status code.

### Issue tracking

- [PYIC-5343](https://govukverify.atlassian.net/browse/PYIC-5343)

[PYIC-5343]: https://govukverify.atlassian.net/browse/PYIC-5343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -105,6 +105,7 @@ async function handleBackendResponse(req, res, backendResponse) {
 
     req.session.currentPage = backendResponse.page;
     req.session.context = backendResponse?.context;
+    req.session.currentPageStatusCode = backendResponse?.statusCode;
 
     return await saveSessionAndRedirect(
       req,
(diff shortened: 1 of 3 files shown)


src/app/ipv/middleware.js
```diff
@@ -274,6 +275,11 @@ module.exports = {
             renderOptions.pageErrorState = req.query.errorState;
           }
 
+          if (req.session.currentPageStatusCode !== undefined) {
+            res.status(req.session.currentPageStatusCode);
+            delete req.session.currentPageStatusCode;
```
Joe-Edwards-GDS: Do you think we need to clear this anywhere else as well - e.g. if there's an unrecoverable timeout, or you end up on the wrong page for some reason?

It should never happen, but I wonder if we want to clear this status code on _any_ page request (e.g. using a `finally`)?

Joe-Edwards-GDS (commented):
I forgot you'd have to faff around with redirects like this...

src/app/ipv/middleware.js
Wynndow: Yeah, that's a good shout. I'll move it to a finally block.

Joe-Edwards-GDS (approved):

PR #1338, opened 2024-03-14
PYIC-5247: added pyi-confirm-name-dob page, refactored ipv middleware

RogVanGDS:
### What changed

- Added pyi-confirm-name-dob page
- refactored ipv middleware to make adding new pages easier

### Issue tracking

- [PYIC-5247](https://govukverify.atlassian.net/browse/PYIC-5247)

[PYIC-5247]: https://govukverify.atlassian.net/browse/PYIC-5247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2024-03-14 at 16 01 22


src/app/ipv/middleware.js
```diff
@@ -26,6 +26,12 @@ const path = require("path");
 const { saveSessionAndRedirect } = require("../shared/redirectHelper");
 const coreBackService = require("../../services/coreBackService");
 
+const directoryPath = __dirname + "/../../views/ipv";
```
Joe-Edwards-GDS: It's good habit to use the `path` helpers for this - it's not _necessary_, but it helps avoid issues with path separators, particularly cross-platform. i.e. `path.join(__dirname, '../../views/ipv');`

(e.g. it would otherwise break if running on Windows)

src/app/ipv/middleware.js
```diff
@@ -233,71 +247,35 @@ module.exports = {
...
-            renderOptions.pageErrorState = req.query.errorState;
-          }
-
-          if (req.session.currentPageStatusCode !== undefined) {
-            res.status(req.session.currentPageStatusCode);
-          }
-
-          return res.render(`ipv/${sanitize(pageId)}.njk`, renderOptions);
+      if (!isValidPage(pageId)) {
+        return res.render("ipv/pyi-technical.njk");
```
Joe-Edwards-GDS: Not related to your change, but do you think we should render the not found page, with a status code 404 here?

test/utils/mocha-helpers.js
```diff
@@ -2,6 +2,7 @@ const chai = require("chai");
 const sinon = require("sinon");
 const sinonChai = require("sinon-chai");
 const chaiAsPromised = require("chai-as-promised");
+process.env.ENABLE_PREVIEW = "test";
```
RogVanGDS, line 5: For local running we set this to 'development' in a .env file, but that causes tests to fail as we are then testing the development-only router.

This will now ensure that local tests properly replicate the pipeline runners

src/views/ipv/pyi-confirm-name-dob.njk
```diff
@@ -0,0 +1,74 @@
...
+{% set isPageDynamic = false %}
+
+{% set errorState = pageErrorState %}
+{% set errorTitle = 'pages.pyiConfirmNameDob.content.formErrorMessage.errorSummaryTitleText' | translate %}
+{% set errorText = 'pages.pyiConfirmNameDob.content.formErrorMessage.errorSummaryDescriptionText' | translate %}
+{% set errorHref = "#pyiConfirmNameDobActionForm" %}
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}"
+      xmlns="http://www.w3.org/1999/html">{{ 'pages.pyiConfirmNameDob.header' | translateWithContext(context) }}</h1>
```
RogVanGDS: No need for dynamic translate here

src/app/ipv/middleware.js
```diff
@@ -233,71 +247,35 @@ module.exports = {
...
-        case "pyi-another-way":
-        case "pyi-timeout-recoverable":
-        case "pyi-timeout-unrecoverable":
-        case "pyi-f2f-technical":
-        case "pyi-escape-m2b":
-        case "pyi-kbv-escape-m2b":
-        case "pyi-continue-with-passport":
-        case "pyi-driving-licence-no-match":
-        case "pyi-driving-licence-no-match-another-way":
-        case "pyi-technical": {
```
Joe-Edwards-GDS, line 267: Have you checked how this list compares to the directory contents? At a glance, I can see `page-featureset` that isn't included here.

Doesn't feel like a massive deal, but I wonder if we should have separate directories for pages to render here, vs. special cases like the feature set.

src/app/ipv/middleware.js
```diff
@@ -233,71 +247,35 @@ module.exports = {
...
-          }
-
-          if (req.session.currentPageStatusCode !== undefined) {
-            res.status(req.session.currentPageStatusCode);
-          }
-
-          return res.render(`ipv/${sanitize(pageId)}.njk`, renderOptions);
+      if (!isValidPage(pageId)) {
+        return res.render("ipv/pyi-technical.njk");
+      } else if (pageRequiresUserDetails(pageId)) {
```
Joe-Edwards-GDS: Should this condition actually be nested a bit deeper? Most of the renderOptions are identical, so I wonder if it's actually more like:
```
if (!isValidPage(pageId)) {
  // error
}

const renderOptions = { <common options> };

if (pageRequiresUserDetails(pageId)) {
  renderOptions.userDetails = ...
}

// render page
```

src/views/ipv/pyi-confirm-name-dob.njk
```diff
@@ -0,0 +1,74 @@
...
+
+        {% set rows = [
+        {
```
Joe-Edwards-GDS: Not a big deal, but the indentation looks off to me here (I'd expect this to be indented one level further)?

Joe-Edwards-GDS:
It would be really helpful if you could provide a screenshot on frontend reviews - so it's easy to see at a glance that it matches the designs :)

(obviously this will be covered later as well, but picking up sooner is better than later!)

src/app/ipv/middleware.js
```diff
@@ -164,6 +170,14 @@ async function handleEscapeAction(req, res, next, actionType) {
   }
 }
 
+function pageRequiresUserDetails(pageId) {
+  return ["page-ipv-reuse", "pyi-confirm-name-dob"].includes(pageId);
```
Joe-Edwards-GDS, line 174: Not critical - I wonder if these should be in a constant somewhere, so they're a bit more obvious.

(I could see in future we might want some kind of 'template metadata' module, but seems unnecessary for now!)

Joe-Edwards-GDS (commented):
This looks good - and I like the refactoring to simplify the rendering middleware!

src/app/ipv/middleware.js
RogVanGDS: Yeah I noticed that as well, I'm not sure what the original decision was behind rendering the pyi-technical page as a pseudo 404 page here.

Definitely feels like we should be consistent between the two, the only difference being that this is just a 404 page for anything within the /ipv/page/* subdomain, and not just a generic 404 page for anything else

src/app/ipv/middleware.js
RogVanGDS: Yeah the other refactoring I was eying up was separating out the ipv/views into more meaningful subdirectories, perhaps that is the answer to this then

Joe-Edwards-GDS: Sounds sensible to me, an obvious choice is to reflect the URL structure (i.e. move these under `/ipv/page`), but we could break up further if there are meaningful buckets to put them in.

src/app/ipv/middleware.js
RogVanGDS: Yeah I think I/Sammy will be touching this in one of the next pages we make so I might leave it for now as this refactor has slightly balooned

src/app/ipv/middleware.js
RogVanGDS: Again may park this for a subsequent refactor

src/app/development/middleware.js
```diff
@@ -5,10 +5,11 @@ const {
   samplePersistedUserDetails,
   generateUserDetails,
 } = require("../shared/reuseHelper");
+const { pageRequiresUserDetails } = require("../ipv/middleware");
 
 async function allTemplatesGet(req, res, next) {
   try {
-    const directoryPath = __dirname + "/../../views/ipv";
+    const directoryPath = path.join(__dirname + "/../../views/ipv/page");
```
Joe-Edwards-GDS: These need to be separate arguments - i.e. `path.join(__dirname, "/../../views/ipv/page")`

src/app/ipv/middleware.js
Joe-Edwards-GDS: Ah, I've figured it out - because we only reach here if `req.session.currentPage == pageId`, if it's invalid then it probably does indicate something has gone wrong (how did it end up as the currentPage).

In any case, I've raised https://govukverify.atlassian.net/browse/PYIC-5435

Joe-Edwards-GDS (commented):
One small comment, but otherwise should be good to approve (assuming all the integration tests pass)

Joe-Edwards-GDS (approved):

PR #1339, opened 2024-03-14
PYIC-4809: Add device select page

MikeCollingwood:
### What changed

- Added the device select page

### Why did it change

- To allow a user to specify what device they're using

### Issue tracking

- [PYIC-4809](https://govukverify.atlassian.net/browse/PYIC-4809)

[PYIC-4809]: https://govukverify.atlassian.net/browse/PYIC-4809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2024-03-14 at 16 07 22
Screenshot 2024-03-14 at 16 07 36


src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,43 @@
+{% extends "shared/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% set pageTitleName = 'pages.pyiTriageSelectDevice.title' | translate %}
```
RogVanGDS: This needs to be just the pageTitleKey without being translated.

We set the key here, and then the base.njk file that we extend in every template uses the key to set the page title, and passes the key into GA4 for analytics events.

src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,43 @@
...
+{% set errorState = pageErrorState %}
+{% set errorTitle = 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorSummaryTitleText' | translate %}
+{% set errorText = 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorSummaryDescriptionText' | translate %}
+{% set errorHref = "#pyiTriageSelectDeviceForm" %}
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiTriageSelectDevice.header' | translate }}</h1>
+  <p class="govuk-body">{{ 'pages.pyiTriageSelectDevice.content.paragraph1' | translate }}</p>
+
+  <form id="pyiTriageSelectDeviceForm" action="/ipv/page/{{ pageId }}" method="POST" onsubmit="return pyiTriageSelectDeviceFormSubmit()">
```
RogVanGDS: I don't think we need the `onsubmit` handler here anymore.

I believe we previously would add in this on each page so that we prevent the user from double-submitting the form by pressing continue twice.

But we now have a global script that should reproduce this behaviour automatically.

@blakeyp is that correct?

src/locales/en/translation.json
```diff
@@ -314,6 +314,22 @@
...
+      "header": "Are you on a computer or tablet right now?",
+      "content": {
+        "paragraph1": "Some ways of confirming your identity require a smartphone. Let us know what you're using right now.",
+        "formRadioButtons": {
+          "onComputerOrTabletButtonText": "Yes, I am on a computer or tablet",
+          "onSmartphoneButtonText": "No, I am on a smartphone"
+        },
+        "formErrorMessage": {
+          "errorSummaryTitleText": "There is a problem",
+          "errorSummaryDescriptionText": "Select what you would like to do",
```
DanCorderIPV: Has this error text come from UCD? I would expect it to be something like `Select whether you are on a smartphone or computer`. We shouldn't just use the same error message for everything.
(https://design-system.service.gov.uk/components/error-message/)

src/views/ipv/pyi-triage-select-device.njk
DanCorderIPV: Relatedly `pyiTriageSelectDeviceFormSubmit()` doesn't actually appear to be defined anywhere?

DanCorderIPV: The script Roger mentions appears to be in `init.js` which gets built into `application.js` which is included in `base.njk`

src/locales/cy/translation.json
```diff
@@ -314,6 +314,22 @@
         }
       }
     },
+    "pyiTriageSelectDevice": {
+      "title": "Ydych chi ar gyfrifiadur neu dabled ar hyn o bryd?",
```
blakeyp: Where are these translations from? the exact content is still TBC so we probably want to just copy over the English here for now (so it doesn't look like these are done and finalised already)

blakeyp: There's a follow up ticket for final content and Welsh - [PYIC-5254](https://govukverify.atlassian.net/browse/PYIC-5254)

src/locales/en/translation.json
blakeyp: Validation is out of scope of this ticket really and we haven't had any error content from UCD yet (though probably fine to leave this as is for now - there's a follow up ticket [PYIC-5254](https://govukverify.atlassian.net/browse/PYIC-5254) to update finalised content)

src/locales/cy/translation.json
MikeCollingwood: These came from DCMAW directly, thankfully it was easy to spot which page this relates to

src/locales/en/translation.json
MikeCollingwood: In general, the text comes from DCMAW, since we haven't received confirmation from UCD. I picked this up from another page template but I see your suggestion is a better fit 👍

src/locales/cy/translation.json
blakeyp: Usually when we don't have the final content/Welsh translations confirmed we just set the English content here for the time being (which is fine since this page won't be going live yet) - @MikeCollingwood could you do that for now? otherwise when we come back to this it may look like we've already done the Welsh translations

MikeCollingwood: These were the welsh translations from DCMAW but I can modify to english even so?

blakeyp: Yeah I think best for now since this content is TBC

src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,43 @@
...
+
+    {% if errorState %}
+      {% set errorMessageObject = { 'text': 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorRadioMessage' | translate } %}
+      {% set radiosConfig = radiosConfig | setAttribute('errorMessage', errorMessageObject) %}
+    {% endif %}
+
+    {{ govukRadios(radiosConfig) }}
+    <button name="submitButton" class="govuk-button" data-module="govuk-button" id="submitButton">
+      {{ 'general.buttons.next' | translate }}
+    </button>
```
blakeyp: We're importing `govukButton` but not using it here - can we? think just need to ensure the `id` is set on it

DanCorderIPV (approved):

src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,43 @@
...
+{% set googleTagManagerPageId = "pyiTriageSelectDevice" %}
+
+{% set errorState = pageErrorState %}
+{% set errorTitle = 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorSummaryTitleText' | translate %}
+{% set errorText = 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorSummaryDescriptionText' | translate %}
+{% set errorHref = "#pyiTriageSelectDeviceForm" %}
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiTriageSelectDevice.header' | translate }}</h1>
+  <p class="govuk-body">{{ 'pages.pyiTriageSelectDevice.content.paragraph1' | translate }}</p>
```
AmritSidhu: From the image in the ticket I believe this is suppose to be a hint in the form rather than a paragraph, you may want to double check that

src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,43 @@
...
+      ]
+    } %}
+
+    {% if errorState %}
+      {% set errorMessageObject = { 'text': 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorRadioMessage' | translate } %}
+      {% set radiosConfig = radiosConfig | setAttribute('errorMessage', errorMessageObject) %}
+    {% endif %}
+
+    {{ govukRadios(radiosConfig) }}
+    {{ govukButton({
```
AmritSidhu: I dont think this is the button we want here. We use this on pages where it is unrecoverable and direct the user to gov.uk

DanCorderIPV: Do you mean the actual button or just the text in it?

DanCorderIPV (changes requested):
Amrit's right with his comments that I missed

src/views/ipv/pyi-triage-select-device.njk
AmritSidhu: I was referring to the text sorry. But also to keep consistency with the buttons we have on other pages where we use html over the nunjucks. Seems to be a bit of inconsistency overall but we majoritly use html buttons for these type of form pages

MikeCollingwood: Ok, I have updated the button to reference `general.buttons.next` which is "Continue". The use of `govukButton` was based on [this comment](https://github.com/govuk-one-login/ipv-core-front/pull/1339#discussion_r1528933932). I don't know, but would've thought `govukButton` was designed to replace html buttons. I also noticed an inconsistency, so it might be a good tech-debt ticket to standardise buttons whichever way we choose

src/views/ipv/pyi-triage-select-device.njk
MikeCollingwood: Do you mean the styling should be the `govuk-hint` CSS class? I see the grey fits the DCMAW screenshot better so will go with that. I can rename the reference `.paragraph1` as well

MikeCollingwood: Screenshot 2024-03-19 at 11 26 42

src/views/ipv/pyi-triage-select-device.njk
blakeyp: Yeah I'd suggested using the macro - be good to gradually refactor towards using this everywhere. But @MikeCollingwood I think you'll need to add `id: submitButton` (required for the event listener which disables the button on click)

src/views/ipv/pyi-triage-select-device.njk
AmritSidhu: Yeah I think thats okay for now.
This is the first scenario where we have a simple question page with a hint. We've not set a standard approach for this yet on another page.
Otherwise we would put the question as a legend along with a hint inside the radiosConfig. But not sure how that will interfere with the `<h1>` element or the GA4  where we are setting `data-page="{{ googleTagManagerPageId }}`

src/views/ipv/pyi-triage-select-device.njk
```diff
@@ -11,7 +11,7 @@
 
 {% block content %}
   <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiTriageSelectDevice.header' | translate }}</h1>
-  <p class="govuk-body">{{ 'pages.pyiTriageSelectDevice.content.paragraph1' | translate }}</p>
+  <p class="govuk-hint">{{ 'pages.pyiTriageSelectDevice.content.hint' | translate }}</p>
```
DanCorderIPV: There's a hint parameter you can set in radiosConfig rather than creating a paragraph.

As a general rule whenever you're creating some HTML on page ask the question whether it's already done for you in a template :)

e.g. to find out about the hint go to this page: https://design-system.service.gov.uk/components/radios/ click on `Nunjucks` and then click on `Nunjucks macro options`

DanCorderIPV: In this case you could also consider using the fieldset legend to replace the `h1` tag but I don't know if you'd be able to preserve the extra `data-page` attribute we have.

DanCorderIPV (changes requested):

src/views/ipv/pyi-triage-select-device.njk
AmritSidhu: Looks like @DanCorderIPV has had the same thought below*

src/views/ipv/pyi-triage-select-device.njk
MikeCollingwood: This is a helpful reference I haven't seen before - I'll use this to also replace my hint implementation

I couldn't manage to add the data-page attribute to the legend, only to the fieldset, so I'm going to leave the `h1` tag in for now

src/views/ipv/pyi-triage-select-device.njk
MikeCollingwood: Apparently the govukButton has a method of stopping double clicks: [link](https://design-system.service.gov.uk/components/button/#:~:text=Stop%20users%20from%20accidentally%20sending%20information%20more%20than%20once). Would it be worth setting this flag to true instead?

AmritSidhu: So that prevents double clicks but doesnt prevent multiple clicks if i remember correctly which is why we implemented our own javascript to prevent it

src/views/ipv/page/pyi-triage-select-device.njk
```diff
@@ -0,0 +1,46 @@
...
+    } %}
+
+    {% if errorState %}
+      {% set errorMessageObject = { 'text': 'pages.pyiTriageSelectDevice.content.formErrorMessage.errorRadioMessage' | translate } %}
+      {% set radiosConfig = radiosConfig | setAttribute('errorMessage', errorMessageObject) %}
+    {% endif %}
+
+    {{ govukRadios(radiosConfig) }}
+    {{ govukButton({
+      text: button_text|default('general.buttons.next' | translate, true),
```
DanCorderIPV: What is `button_text` here?

Searching through core-front I can't actually see it being set anywhere, maybe we can delete it in general?

DanCorderIPV (approved):

PR #1341, opened 2024-03-15
PYIC-5402 Added 3 new pages for CI at passport mitigation with english and welsh translation.

rpayal:
## Proposed changes
Add 3 new pages for CI at passport mitigation with english and welsh translation for same and separate session mitigation.

### Issue tracking

- [PYIC-5402](https://govukverify.atlassian.net/browse/PYIC-5402)

### Other considerations

[PYIC-5402]: https://govukverify.atlassian.net/browse/PYIC-5402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/ipv/page/pyi-continue-with-driving-licence.njk
+++ b/src/views/ipv/page/pyi-continue-with-driving-licence.njk
@@ -11,13 +11,13 @@
 
 {% block content %}
   <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiContinueWithDrivingLicence.header' | translate }}</h1>
-  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph1' | translate | safe }}</p>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph1' | translate }}</p>
   <ul class="govuk-list govuk-list--bullet">
-    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet1' | translate | safe }}</li>
-    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet2' | translate | safe }}</li>
+    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet1' | translate }}</li>
+    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet2' | translate }}</li>
   </ul>
-  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph2' | translate | safe }}</p>
-  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph3' | translate | safe }}</p>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph2' | translate }}</p>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph3' | translate }}</p>
 
   <form id="pyiContinueWithDrivingLicenceForm" action="/ipv/page/{{ pageId }}" method="POST">
     <input type="hidden" name="_csrf" value="{{ csrfToken }}">
--- a/src/views/ipv/page/pyi-continue-with-driving-licence.njk
+++ b/src/views/ipv/page/pyi-continue-with-driving-licence.njk
@@ -1,0 +1,56 @@
+{% extends "shared/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% set pageTitleName = 'pages.pyiContinueWithDrivingLicence.title' | translate %}
+{% set googleTagManagerPageId = "pyiContinueWithDrivingLicence" %}
+
+{% set errorState = pageErrorState %}
+{% set errorTitle = 'pages.pyiContinueWithDrivingLicence.content.formErrorMessage.errorSummaryTitleText' | translate %}
+{% set errorText = 'pages.pyiContinueWithDrivingLicence.content.formErrorMessage.errorSummaryDescriptionText' | translate %}
+{% set errorHref = "#pyiContinueWithDrivingLicenceForm" %}
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiContinueWithDrivingLicence.header' | translate }}</h1>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph1' | translate | safe }}</p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet1' | translate | safe }}</li>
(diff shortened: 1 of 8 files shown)


src/views/ipv/page/pyi-continue-with-driving-licence.njk
```diff
@@ -0,0 +1,56 @@
+{% extends "shared/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% set pageTitleName = 'pages.pyiContinueWithDrivingLicence.title' | translate %}
```
RogVanGDS: This actually needs to be setting `pageTitleKey` to the key string and without translating.

We refactored the base.njk file for the new GA4 analytics implementation, and the `base.njk` component is now handling the translation of the `pageTitleName`

src/views/ipv/page/pyi-passport-no-match-another-way.njk
```diff
@@ -0,0 +1,56 @@
+{% extends "shared/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% set pageTitleName = 'pages.pyiPassportNoMatchAnotherWay.title' | translate %}
```
RogVanGDS: Same deal here.

Do you mind also fixing this in `pyi-continue-with-passport`, `pyi-driving-licence-no-match-another-way` and `pyi-driving-licence-no-match` as part of this review?

Just realised that they also need to change

src/views/ipv/page/pyi-passport-no-match.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "shared/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% set pageTitleName = 'pages.pyiPassportNoMatch.title' | translate %}
```
RogVanGDS: Likewise here

src/views/ipv/page/pyi-passport-no-match.njk
```diff
@@ -0,0 +1,14 @@
...
+{% set googleTagManagerPageId = "pyiPassportNoMatch" %}
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiPassportNoMatch.header' | translate }}</h1>
+  <p class="govuk-body">{{ 'pages.pyiPassportNoMatch.content.paragraph1' | translateWithContext(context) }}</p>
```
RogVanGDS: What context are we supporting here? I can't seem to find it in the translation files.

We should avoid using translateWithContext unless we specifically need this page to be dynamic, as somebody could accidentally render the wrong content by changing the journey-map in the backend and it would be quite tricky to spot.

src/views/ipv/page/pyi-continue-with-driving-licence.njk
```diff
@@ -0,0 +1,56 @@
...
+
+{% block content %}
+  <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pyiContinueWithDrivingLicence.header' | translate }}</h1>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph1' | translate | safe }}</p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet1' | translate | safe }}</li>
+    <li>{{ 'pages.pyiContinueWithDrivingLicence.content.bullet2' | translate | safe }}</li>
+  </ul>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph2' | translate | safe }}</p>
+  <p class="govuk-body">{{ 'pages.pyiContinueWithDrivingLicence.content.paragraph3' | translate | safe }}</p>
```
RogVanGDS: Do we need the `safe` nunjucks filter here?

I thought we only needed it to escape special characters in the translation strings (i.e. if we need to add an anchor tag/bold halfway through a string)

Possibly I've misunderstood, but I'm not sure we need it 😃

src/views/ipv/page/pyi-continue-with-driving-licence.njk
rpayal: done

src/views/ipv/page/pyi-passport-no-match-another-way.njk
rpayal: done

src/views/ipv/page/pyi-passport-no-match.njk
rpayal: done

src/views/ipv/page/pyi-passport-no-match.njk
rpayal: done

src/views/ipv/page/pyi-continue-with-driving-licence.njk
rpayal: We are saying this output is safe so Nunjucks not to escape this output.

src/views/ipv/page/pyi-continue-with-driving-licence.njk
```diff
@@ -0,0 +1,56 @@
...
+      {% set errorMessageObject = { 'text': 'pages.pyiContinueWithDrivingLicence.content.formErrorMessage.errorRadioMessage' | translate } %}
+      {% set radiosConfig = radiosConfig | setAttribute('errorMessage', errorMessageObject) %}
+    {% endif %}
+
+    {{ govukRadios(radiosConfig) }}
+    <button name="submitButton" class="govuk-button" data-module="govuk-button" id="submitButton">
+      {{ 'general.buttons.next' | translate }}
+    </button>
+  </form>
+{% endblock %}
```
Wynndow: Looks like this page is missing the "Contact the GOV.UK One Login team..." link.
Same for the `pyi-continue-with-passport` page.
I think it might just be a case of adding the below
```suggestion
  </form>
  <p class="govuk-body"><a target="_blank" rel="noopener noreferrer" href="{{ contactUsUrl }}" class="govuk-link">{{ 'general.shared.contactLinkText' | translate }}</a></p>

{% endblock %}
```

src/views/ipv/page/pyi-continue-with-driving-licence.njk
RogVanGDS: Ah yes sorry think I didn't explain that well. 

Nunjucks will escape special characters by default to prevent us from writing html instead of strings. Putting `| safe` in is essentially telling Nunjucks that the string contains special characters and we want to render it as html.

But I'm not sure it's necessary here, because our string doesn't contain any special characters that we would need to escape, so we don't need to use the safe filter.

src/views/ipv/page/pyi-continue-with-driving-licence.njk
rpayal: Updated

src/views/ipv/page/pyi-continue-with-driving-licence.njk
rpayal: @RogVanGDS  seems we may need it for welsh translations e.g â'ch

rpayal: As discussed removed the safe filter from both the pages.

RogVanGDS (approved):

PR #1347, opened 2024-03-19
PYIC-4821: Add desktop app download page

DanCorderIPV:
### What changed

Added desktop version of the app download page

### Why did it change

New page for new app flow

### Issue tracking

- [PYIC-4821](https://govukverify.atlassian.net/browse/PYIC-4821)

[PYIC-4821]: https://govukverify.atlassian.net/browse/PYIC-4821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

image

### DIFF
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -66,7 +66,8 @@ async function templatesDisplayGet(req, res) {
   }
 
   if (templateId === "pyi-triage-desktop-download-app") {
-    renderOptions.qrCode = await qrCodeHelper.generateQrCodeImageData("https://bbc.co.uk");
+    renderOptions.qrCode =
+      await qrCodeHelper.generateQrCodeImageData("https://bbc.co.uk");
   }
 
   return res.render(`ipv/page/${sanitize(templateId)}.njk`, renderOptions);
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -6,6 +6,7 @@ const {
   generateUserDetails,
 } = require("../shared/reuseHelper");
 const { pageRequiresUserDetails } = require("../ipv/middleware");
+const qrCodeHelper = require("../shared/qrCodeHelper");
 
 async function allTemplatesGet(req, res, next) {
   try {
@@ -62,6 +63,10 @@ async function templatesDisplayGet(req, res) {
       samplePersistedUserDetails,
       req.i18n,
     );
+  }
+
+  if (templateId === "pyi-triage-desktop-download-app") {
+    renderOptions.qrCode = await qrCodeHelper.generateQrCodeImageData("https://bbc.co.uk");
   }
 
   return res.render(`ipv/page/${sanitize(templateId)}.njk`, renderOptions);
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -5,8 +5,9 @@ const {
   samplePersistedUserDetails,
   generateUserDetails,
 } = require("../shared/reuseHelper");
-const { pageRequiresUserDetails } = require("../ipv/middleware");
+const { pageRequiresUserDetails, CONSTANTS } = require("../ipv/middleware");
 const qrCodeHelper = require("../shared/qrCodeHelper");
+const { SERVICE_URL } = require("../../lib/config");
 
 async function allTemplatesGet(req, res, next) {
   try {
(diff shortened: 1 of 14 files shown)


src/assets/scss/application.scss
```diff
@@ -99,3 +99,25 @@
 .govuk-btn-link:hover {
   background: none;
 }
+
+// Mostly copied from F2F CRI styling https://github.com/govuk-one-login/ipv-cri-f2f-front/blob/8c05ac2b93cbe92113de5877f29b4101cc4e2b15/src/assets/scss/application.scss#L129
```
DanCorderIPV, line 103: Ideally we wouldn't have to do this, but there doesn't seem to be a cross-identity set of shared styles and this isn't a GOV.UK Design System component.

src/app/shared/qrCodeHelper.js
```diff
@@ -0,0 +1,14 @@
+const qrCode = require("qrcode");
+
+async function generateQrCodeImageData(url) {
+  return qrCode.toDataURL(url, {
+    type: "image/jpeg",
+    margin: 0,
+    errorCorrectionLevel: "H",
+    version: 10,
+  });
```
MikeCollingwood, line 9: Could we set the scale/ width to maker the rendered QR code smaller like in DCMAW? Additionally I think [this stackoverflow post](https://stackoverflow.com/a/64939039) looks like it could help in adding the icon to the centre of the QR code and this method would be a good one to take those changes

MikeCollingwood (commented):
Looks good but had a thought about the rendered QR code

src/app/development/middleware.js
```diff
@@ -64,6 +65,11 @@ async function templatesDisplayGet(req, res) {
     );
   }
 
+  if (templateId === "pyi-triage-desktop-download-app") {
+    renderOptions.qrCode =
+      await qrCodeHelper.generateQrCodeImageData("https://bbc.co.uk");
```
blakeyp: Is there a way to test this as a standalone page? (can we not pass the `app-redirect` url here?)

src/app/shared/qrCodeHelper.js
DanCorderIPV: Good point - I asked about the icon in the middle in UCD slack channel (as having an icon there is a bit hacky) and then forgot I hadn't implemented it.
I can shrink the QR code to match the mock-up if we think it's worth it.

src/app/development/middleware.js
DanCorderIPV: Do you mean the redirect route? I wasn't sure how/if that would work on `build` from the `all-templates` page. I was assuming that this would be covered later on once we have the routing in and some way of getting to these pages with a feature flag.
Also, I think on `build` the redirect URL will be read from the environment and so will be `get real value` at the moment.
We could update the template with some dummy values as long as we don't forget to update them later :)

blakeyp: Yep the redirect route, I was hoping we could at least test the scanning of the QR code opening the app store link (even if it's just hardcoded for an iPhone for now)

src/app/shared/qrCodeHelper.js
blakeyp: The existing DCMAW download screen doesn't have the icon in the middle of the QR code, I think we probably don't want to bother with it if it's non-standard - I'll comment on the ticket

src/lib/config.js
```diff
@@ -17,6 +17,10 @@ module.exports = {
   API_CRI_CALLBACK: "/cri/callback",
   API_SESSION_INITIALISE: "/session/initialise",
   API_BUILD_PROVEN_USER_IDENTITY_DETAILS: "/user/proven-identity-details",
+  APP_STORE_URL_ANDROID:
+    process.env.APP_STORE_URL_ANDROID || "https://play.google.com/store/apps",
+  APP_STORE_URL_APPLE:
+    process.env.APP_STORE_URL_APPLE || "https://www.apple.com/uk/app-store/",
```
blakeyp: Could these be the links for the GOV.UK ID Check app for now for testing: 
https://apps.apple.com/gb/app/gov-uk-id-check/id1629050566
https://play.google.com/store/apps/details?id=uk.gov.documentchecking

src/app/development/middleware.js
DanCorderIPV: Ok, I've done that, we'll have to try it on `build` to see what actually happens - will the phone need to be on the VPN to access `build`?

blakeyp: That's fine we can merge and test this in build given it's not live - phone shouldn't need to be on VPN

blakeyp (approved):

PR #1348, opened 2024-03-19
PYIC-5256: Added redirect to GovUK contact us page for 6MFC flow

RogVanGDS:
### What changed

Added redirect to GovUK contact us page for 6MFC flow

### Why did it change

If the user needs to change their name or dob, we don't currently support this in OneLogin, so the user is redirected to the GOVUK Contact us screen

### Issue tracking

- [PYIC-5256](https://govukverify.atlassian.net/browse/PYIC-5256)

[PYIC-5256]: https://govukverify.atlassian.net/browse/PYIC-5256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/ipv/page/pyi-confirm-address.njk
+++ b/src/views/ipv/page/pyi-confirm-address.njk
@@ -46,7 +46,7 @@
             },
             items: [
                     {
-                    value: "fraud",
+                    value: "addressCurrent",
                     text: 'pages.pyiConfirmAddress.content.formRadioButtons.yes' | translate
                     },
                     {
--- a/src/views/ipv/page/pyi-confirm-address.njk
+++ b/src/views/ipv/page/pyi-confirm-address.njk
@@ -46,11 +46,11 @@
             },
             items: [
                     {
-                    value: "next",
+                    value: "fraud",
                     text: 'pages.pyiConfirmAddress.content.formRadioButtons.yes' | translate
                     },
                     {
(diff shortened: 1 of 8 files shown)


src/views/ipv/page/pyi-confirm-address.njk
```diff
@@ -46,11 +46,11 @@
             },
             items: [
                     {
-                    value: "next",
+                    value: "fraud",
```
Joe-Edwards-GDS: I'm not sure I _love_ `fraud` as the event here - since it describes what you think the next state will be (which is the concern of the journey map), rather than the event itself.

I guess better events might be more like `address-current` and `address-changed` or something.

This will be easier once we do the refactoring we discussed, and we can be a bit more flexible about our event names.

RogVanGDS: Yep, that makes sense.

Annoyingly the "next" event has to be a "next" event, as the Address and Fraud CRI only accepts certain input journey events into the nested subjourney.

But address-current is much better than fraud, so that will be clearer at least

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1351, opened 2024-03-20
Pyic 4822 mobile app download page

DanCorderIPV:
### What changed

Add mobile app download page

### Why did it change

New page

### Issue tracking

- [PYIC-4822](https://govukverify.atlassian.net/browse/PYIC-4822)

image

[PYIC-4822]: https://govukverify.atlassian.net/browse/PYIC-4822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -7,8 +7,8 @@ const {
 } = require("../shared/reuseHelper");
 const { pageRequiresUserDetails } = require("../ipv/middleware");
 const qrCodeHelper = require("../shared/qrCodeHelper");
-const { SERVICE_URL, APP_STORE_URL_APPLE } = require("../../lib/config");
 const PHONE_TYPES = require("../../constants/phone-types");
+const appDownloadHelper = require("../shared/appDownloadHelper");
 
 async function allTemplatesGet(req, res, next) {
   try {
@@ -69,10 +69,12 @@ async function templatesDisplayGet(req, res) {
 
   if (templateId === "pyi-triage-desktop-download-app") {
     renderOptions.qrCode = await qrCodeHelper.generateQrCodeImageData(
-      SERVICE_URL + "/ipv/app-redirect/" + PHONE_TYPES.IPHONE,
+      appDownloadHelper.getAppStoreRedirectUrl(PHONE_TYPES.IPHONE),
     );
   } else if (templateId === "pyi-triage-mobile-download-app") {
-    renderOptions.appDownloadUrl = APP_STORE_URL_APPLE;
+    renderOptions.appDownloadUrl = appDownloadHelper.getAppStoreRedirectUrl(
+      PHONE_TYPES.IPHONE,
+    );
   }
 
   return res.render(`ipv/page/${sanitize(templateId)}.njk`, renderOptions);
--- a/src/app/development/middleware.js
+++ b/src/app/development/middleware.js
@@ -5,7 +5,7 @@ const {
   samplePersistedUserDetails,
   generateUserDetails,
 } = require("../shared/reuseHelper");
-const { pageRequiresUserDetails, CONSTANTS } = require("../ipv/middleware");
+const { pageRequiresUserDetails } = require("../ipv/middleware");
 const qrCodeHelper = require("../shared/qrCodeHelper");
 const { SERVICE_URL, APP_STORE_URL_APPLE } = require("../../lib/config");
 const PHONE_TYPES = require("../../constants/phone-types");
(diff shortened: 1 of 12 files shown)


src/app/ipv/middleware.js
```diff
@@ -303,10 +302,12 @@ module.exports = {
...
       } else if (pageId === "pyi-triage-desktop-download-app") {
         // PYIC-4816: Use the actual device type selected on a previous page.
-        const qrCodeUrl =
-          SERVICE_URL + "/app-redirect/" + CONSTANTS.PHONE_TYPES.IPHONE;
+        const qrCodeUrl = SERVICE_URL + "/app-redirect/" + PHONE_TYPES.IPHONE;
         renderOptions.qrCode =
           await qrCodeHelper.generateQrCodeImageData(qrCodeUrl);
+      } else if (pageId === "pyi-triage-mobile-download-app") {
+        // PYIC-4816: Use the actual device type selected on a previous page and/or the current request's sniffed device type
+        renderOptions.appDownloadUrl = getAppStoreUrl(PHONE_TYPES.IPHONE);
```
DanCorderIPV: I'm starting to get a bit concerned about this function and how it handles so many different pages. If we stick with the idea that refreshing the `pyi-triage-desktop-download-app` page should do a check with core-back for the VC then that will probably need to be split out of here.

src/constants/phone-types.js
```diff
@@ -0,0 +1,4 @@
+module.exports = Object.freeze({
```
DanCorderIPV, line 1: If people are happy with this pattern I might move other constants I find to similar files

RogVanGDS: Yeah I think this would be a big improvement. There are quite a lot of places where we manually type out the journey event or page name which should probably be constants in reality

src/app/ipv/middleware.js
```diff
@@ -196,25 +190,30 @@ function isValidPage(pageId) {
...
-      error,
-      `error redirecting to app store for specified phone type ${specifiedPhoneType}`,
-    );
+    transformError(error, `Error redirecting to app store`);
     next(error);
   }
 }
 
+function getAppStoreUrl(phoneType) {
+  if (phoneType === PHONE_TYPES.IPHONE) {
```
MikeCollingwood: Could this suit being a switch statement?

src/constants/phone-types.js
MikeCollingwood: Would it be preferable to have one constants file as opposed to one for each type? E.g. to put this into app.constants.js

DanCorderIPV: Maybe, I think we'll find out once we start putting other constants here.

src/app/ipv/middleware.js
DanCorderIPV: It's getting there, I think it's actually more lines as a switch so I'll leave it like this for now, but if we ever add more branches we should probably change it.

DanCorderIPV:
I've added some code for setting Android intents. It's currently untested as don't have an android app to test with, but I think it's worth keeping the code as the structure should be right even if not all the details are 100%

src/app/ipv/middleware.js
```diff
@@ -211,22 +205,19 @@ function isValidPage(pageId) {
...
-    } else {
-      throw new Error("Unrecognised phone type: " + specifiedPhoneType);
     }
+    throw new Error("Unrecognised phone type: " + specifiedPhoneType);
   } catch (error) {
-    transformError(
-      error,
-      `error redirecting to app store for specified phone type ${specifiedPhoneType}`,
-    );
+    transformError(error, `Error redirecting to app store`);
```
MikeCollingwood, line 220: V minor: probably doesn't want the backticks

src/app/development/middleware.js
```diff
@@ -68,7 +69,11 @@ async function templatesDisplayGet(req, res) {
 
   if (templateId === "pyi-triage-desktop-download-app") {
     renderOptions.qrCode = await qrCodeHelper.generateQrCodeImageData(
-      SERVICE_URL + "/ipv/app-redirect/" + CONSTANTS.PHONE_TYPES.IPHONE,
+      appDownloadHelper.getAppStoreRedirectUrl(PHONE_TYPES.IPHONE),
+    );
+  } else if (templateId === "pyi-triage-mobile-download-app") {
+    renderOptions.appDownloadUrl = appDownloadHelper.getAppStoreRedirectUrl(
+      PHONE_TYPES.IPHONE,
```
MikeCollingwood, line 76: In ipv/middleware it's PHONE_TYPES.ANDROID for this one, I assume that was just for testing the android link works

MikeCollingwood (approved):
Looks good to merge 👍

PR #1363, opened 2024-03-28
DFC-493 Add language toggle

di-fabs:
Add language toggle into all views.

https://govukverify.atlassian.net/browse/DFC-493
### What changed
- All pages contain now new links: one for switching language to English, the other one for switching to Welsh
- Feature flag added as an env variable: 1 to enable the language toggle, 0 to disable

- Install a new node module "language toggle"
- import related css file into application.scss
- add a new env variable.
- Edit deployment template and add new env variable (default value is 0)
### Issue tracking

https://govukverify.atlassian.net/browse/DFC-493

- [x] Documented in the [README](./blob/main/README.md)

### Other considerations

- [x] Update [README](./blob/main/README.md) with any new instructions or tasks


deploy/template.yaml
```diff
@@ -707,6 +707,8 @@ Resources:
                   - !If [IsDev02, !Sub "https://${Environment}.02.dev.identity.account.gov.uk", !Sub "https://identity.${Environment}.account.gov.uk"]
             - Name: SESSION_SECRET
               Value: "no-secret"
+            - Name: LANGUAGE_TOGGLE
+              Value: "0"
```
blakeyp: for existing env vars we use false/true - could we switch to that instead of 0/1 ?

src/lib/config.js
```diff
@@ -36,4 +36,5 @@ module.exports = {
   SERVICE_URL: process.env.SERVICE_URL,
   TEMPLATE_CACHING: process.env.TEMPLATE_CACHING,
   SERVICE_DOMAIN: process.env.SERVICE_DOMAIN || "localhost",
+  LANGUAGE_TOGGLE_ENABLED: process.env.LANGUAGE_TOGGLE === true || false,
```
blakeyp: I believe the values of the env vars are actually strings (as aws interprets them) so this should be `LANGUAGE_TOGGLE_ENABLED: process.env.LANGUAGE_TOGGLE === "true"` ?

blakeyp (approved):

PR #1368, opened 2024-04-02
PYIC-5537: Refactored journey event logic to allow for backend event validation

RogVanGDS:
### What changed

Refactored the journey event posting from core-front to core-back to include a current page search parameter

### Why did it change

We need to make this change to give core-back the information required to apply some journey event validation. This change will not validate anything by itself, but should let us start validating without breaking in-flight journeys once the backend changes are ready.

### Issue tracking

- [PYIC-5537](https://govukverify.atlassian.net/browse/PYIC-5537)

[PYIC-5537]: https://govukverify.atlassian.net/browse/PYIC-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -59,7 +59,18 @@ async function journeyApi(action, req) {
   return coreBackService.postAction(req, action);
 }
 
+function getIpvPageFromRequest(req) {
+  const pathParts = req.path.split('/')
+
+  return pathParts.slice(2).join("/")
+}
+
 async function handleJourneyResponse(req, res, action) {
+  const backendResponse = (await journeyApi(action, req)).data;
+  return await handleBackendResponse(req, res, backendResponse);
+}
+
+async function handleJourneyResponseTest(req, res, action) {
   const backendResponse = (await journeyApi(action, req)).data;
   return await handleBackendResponse(req, res, backendResponse);
 }
@@ -132,6 +143,8 @@ async function handleBackendResponse(req, res, backendResponse) {
       `/ipv/page/${req.session.currentPage}`,
     );
   }
+
+  return res.render("ipv/page/pyi-technical.njk");
 }
 
 function tryValidateCriResponse(criResponse) {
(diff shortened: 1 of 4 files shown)


src/app/oauth2/middleware.js
```diff
@@ -3,47 +3,72 @@ const { logCoreBackCall, transformError } = require("../shared/loggerHelper");
...
+  next();
+}
+
+async function handleOAuthJourneyAction(req, res, next) {
+  try {
+    checkForIpvAndOauthSessionId(req, res);
+    await handleOAuthResponse(req, res, "next");
+  } catch (error) {
+    transformError(error, "error invoking handleOAuthJourneyAction");
+    next(error);
```
RogVanGDS, line 60: I think this logic merited being extracted into its own middleware function for the oauth route, as it was very hard to figure out what would happen in this use case, and it felt a bit too tightly coupled to the user-driven journey event logic.

Happy to revert if people feel differently though

src/app/ipv/middleware.js
```diff
@@ -132,11 +142,13 @@ async function handleBackendResponse(req, res, backendResponse) {
       `/ipv/page/${req.session.currentPage}`,
     );
   }
+
+  return res.render("ipv/page/pyi-technical.njk");
```
Joe-Edwards-GDS, line 146: This has been superseded by Laurence's last PR

src/app/ipv/middleware.js
```diff
@@ -59,8 +59,18 @@ async function journeyApi(action, req) {
   return coreBackService.postAction(req, action);
 }
 
+function getIpvPageFromRequest(req) {
+  const pathParts = req.path.split("/");
+
+  return pathParts.slice(2).join("/");
+}
```
Joe-Edwards-GDS, line 66: Would it be possible to pass this in more precisely (e.g. using the pageId from the router?) I'm a bit nervous that this makes assumptions about our URL structure and might be a bit fragile.

src/app/ipv/middleware.js
```diff
@@ -187,11 +199,11 @@ async function handleEscapeAction(req, res, next, actionType) {
     checkForSessionId(req, res);
 
     if (req.body?.journey === "next/f2f") {
-      await handleJourneyResponse(req, res, "journey/f2f");
+      await handleJourneyResponse(req, res, "f2f");
```
Joe-Edwards-GDS, line 202: (This will conflict with my PR, sorry :( https://github.com/govuk-one-login/ipv-core-front/pull/1376 - hopefully it's just identical though...)

src/app/oauth2/router.js
```diff
@@ -4,8 +4,13 @@ const router = express.Router();
 
 const { setIpvSessionId, setIpAddress } = require("./middleware");
 
-const { handleJourneyAction } = require("../ipv/middleware");
+const { handleOAuthJourneyAction } = require("../oauth2/middleware");
```
Joe-Edwards-GDS, line 7: I think this should be combined with the import above?

src/app/oauth2/middleware.js
Joe-Edwards-GDS: I think having a separate handler here is fine, but I wonder if we should still be using the same `handleJourneyResponse` function?

I think ideally we'd make the `pageId` an optional parameter to `handleJourneyResponse` and pass it explicitly from callers (tying into the comment above - we can then pass it explicitly, rather than deconstructing the request URL)

src/app/ipv/middleware.js
RogVanGDS: Ah so we can (and probably should) do this for the generic /page/:pageId route that calls handleJourneyAction. 

But we won't be able to do this for the handleEscapeAction routes (or any static route that doesn't use a path parameter)

This is probably even more of a reason to collapse those routes into one so we can access the path parameter everywhere.

Happy to do this as part of this ticket, although don't we have a further refactoring ticket ready to deal with this?

Joe-Edwards-GDS: Yeah, that's what I was expecting @sammyiyke 's ticket to cover (PYIC-5436)

For the static routes, I was assuming you could just pass through the fixed path (I think some of them already pass the 'action', which could be clarified to 'pageId')

PR #1369, opened 2024-04-03
BAU: Fixed user details not showing when RFC pages error

RogVanGDS:
### What changed

Fixed user details not showing when RFC pages error

### Why did it change

We didn't store a fetched user details anywhere when displaying the detail confirmation pages as part of the Repeat Fraud Check flow. 

Fixed by storing user details on the request session, so that we can populate the frontend with those details when the page reloads due to a user not selecting a radio option.

### Issue tracking

- [PYIC-5256](https://govukverify.atlassian.net/browse/PYIC-5256)

[PYIC-5256]: https://govukverify.atlassian.net/browse/PYIC-5256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -432,7 +432,7 @@ module.exports = {
   formRadioButtonChecked: async (req, res, next) => {
     try {
       const { context } = req?.session || "";
-      const pageId = req.params.currentPage;
+      const pageId = req.session.currentPage;
 
       const renderOptions = {
         pageId,
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -311,10 +311,10 @@ module.exports = {
         const userDetailsResponse =
           await coreBackService.getProvenIdentityUserDetails(req);
 
-        const userDetails = generateUserDetails(userDetailsResponse, req.i18n);
-
-        renderOptions.userDetails = userDetails;
-        req.session.userDetails = userDetails;
+        renderOptions.userDetails = generateUserDetails(
+          userDetailsResponse,
+          req.i18n,
+        );
       } else if (pageId === "pyi-triage-desktop-download-app") {
         // PYIC-4816: Use the actual device type selected on a previous page.
         const qrCodeUrl = appDownloadHelper.getAppStoreRedirectUrl(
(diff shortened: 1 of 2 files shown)


src/app/ipv/middleware.js
```diff
@@ -310,10 +310,11 @@ module.exports = {
...
           await coreBackService.getProvenIdentityUserDetails(req);
-        renderOptions.userDetails = generateUserDetails(
-          userDetailsResponse,
-          req.i18n,
-        );
+
+        const userDetails = generateUserDetails(userDetailsResponse, req.i18n);
+
+        renderOptions.userDetails = userDetails;
+        req.session.userDetails = userDetails;
```
Joe-Edwards-GDS: How long do the frontend sessions last? Do you think there's any risk to us storing/using this (PII) data inappropriately?

src/app/ipv/middleware.js
```diff
@@ -425,14 +426,30 @@ module.exports = {
...
+      };
+
+      if (pageRequiresUserDetails(pageId)) {
+        const userDetailsResponse =
+          await coreBackService.getProvenIdentityUserDetails(req);
+
+        renderOptions.userDetails = generateUserDetails(
+          userDetailsResponse,
+          req.i18n,
+        );
```
Joe-Edwards-GDS: Should this be pulled out into a method?

src/app/ipv/middleware.js
```diff
@@ -425,14 +426,30 @@ module.exports = {
   formRadioButtonChecked: async (req, res, next) => {
     try {
       const { context } = req?.session || "";
+      const { pageId } = req.params;
```
Joe-Edwards-GDS: I think this will be undefined for the pages with special routing (e.g. pyi-cri-escape) - is that likely to cause an issue? Should we be using `req.session.currentPage` (or parsing it from the URL properly)?

src/app/ipv/middleware.js
```diff
@@ -425,14 +432,22 @@ module.exports = {
   formRadioButtonChecked: async (req, res, next) => {
     try {
       const { context } = req?.session || "";
+      const pageId = req.params.currentPage;
```
Joe-Edwards-GDS: Is this right - the old code has `req.session.currentPage`

src/app/ipv/middleware.js
```diff
@@ -336,7 +338,10 @@ module.exports = {
         }
       }
 
-      return res.render(`ipv/page/${sanitize(pageId)}.njk`, renderOptions);
+      const sanitzedPageId = sanitize(pageId);
+
+      req.session.currentPage = sanitzedPageId;
```
Joe-Edwards-GDS: What cases is this intended to catch? Earlier in this method we have an if block that sets this to something else if `req.session.currentPage !== pageId`, so I'm not sure when it would be different.

(also typo in `sanitzed`)

src/app/ipv/middleware.js
```diff
@@ -425,14 +430,24 @@ module.exports = {
...
 
       if (req.method === "POST" && req.body.journey === undefined) {
-        res.render(`ipv/page/${sanitize(req.session.currentPage)}.njk`, {
-          pageId: req.session.currentPage,
-          csrfToken: req.csrfToken(),
-          pageErrorState: true,
-          context,
-        });
+        res.render(
+          `ipv/page/${sanitize(req.session.currentPage)}.njk`,
```
Joe-Edwards-GDS: (use pageId here as well?)

Joe-Edwards-GDS (commented):
We can probably just get this merged if it's blocking testing, but I'm not sure about the overwriting the session page id?

src/app/ipv/middleware.js
RogVanGDS: Ah this was left over from testing some changes out, I'll revert

Joe-Edwards-GDS (dismissed):
Sorry for making this go round so much!

Joe-Edwards-GDS (approved):

RogVanGDS:
> Sorry for making this go round so much!

Nah it's my fault for juggling too many things 😄

PR #1371, opened 2024-04-04
SPT-700: Add SPoT signing key bucket to `template.taml`

sam-leen-gds:
### What changed

- Add an S3 bucket (`SpotKeysS3Bucket`) to cloudformation template.
- Define access permissions on `SpotKeysS3Bucket` for ipv-core, and spot accounts.
- Add a KMS key (`S3KmsKey`) to encrypt contents of `SpotKeysS3Bucket`.

### Why did it change

We will be using IPV Core to host our signing keys, we should create the document in a bucket in their account in order to avoid any issues that may occur from cross-account permissions being removed.

### Issue tracking

- [SPT-700](https://govukverify.atlassian.net/browse/SPT-700)

[SPT-700]: https://govukverify.atlassian.net/browse/SPT-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -86,6 +86,7 @@ Mappings:
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountArn: arn:aws:iam::738810260032:root
     "175872367215": # core-dev02
       lb400ErrorLimit: 10
       lb500ErrorLimit: 2
@@ -105,6 +106,7 @@ Mappings:
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountArn: arn:aws:iam::738810260032:root
     "457601271792": # Build
       lb400ErrorLimit: 10
       lb500ErrorLimit: 2
@@ -128,6 +130,7 @@ Mappings:
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountArn: arn:aws:iam::429671060046:root
     "335257547869": # Staging
       lb400ErrorLimit: 10
       lb500ErrorLimit: 2
@@ -147,6 +150,7 @@ Mappings:
       contactUrl: "https://home.staging.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountArn: arn:aws:iam::444977453444:root
     "991138514218": # Integration
       lb400ErrorLimit: 10
       lb500ErrorLimit: 2
@@ -166,6 +170,7 @@ Mappings:
       contactUrl: "https://home.integration.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountArn: arn:aws:iam::298945768017:root
     "075701497069": # Production
       lb400ErrorLimit: 20
       lb500ErrorLimit: 2
@@ -185,6 +190,7 @@ Mappings:
       contactUrl: "https://home.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceProductionVariables
+      spotAccountArn: arn:aws:iam::101836073728:root
   SecurityGroups:
     PrefixListIds:
       dynamodb: "pl-b3a742da"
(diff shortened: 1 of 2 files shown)


deploy/template.yaml
```diff
@@ -105,6 +106,7 @@ Mappings:
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountNumber: 738810260032"arn:aws:iam::738810260032:root"
```
chris-dco:
```suggestion
      spotAccountNumber: "arn:aws:iam::738810260032:root"
```

deploy/template.yaml
```diff
@@ -347,6 +353,51 @@ Resources:
...
+        RestrictPublicBuckets: true
+
+  SpotKeysBucketPolicy:
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref SpotKeysS3Bucket
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: AllowAccount1Access
```
chris-dco: Should this be `AllowAccountAccess`?

deploy/template.yaml
```diff
@@ -347,6 +353,51 @@ Resources:
...
+            Resource: !Sub "arn:aws:s3:::${SpotKeysS3Bucket}/*"
+          - Sid: AllowSpotAccountAccess
+            Effect: Allow
+            Principal:
+              AWS: !FindInMap
+                - EnvironmentConfiguration
+                - !Ref AWS::AccountId
+                - spotAccountNumber
+            Action:
+              - "s3:*"
```
chris-dco: This access feels too broad. I think we only need `s3:PutObject` and `s3:GetObject`.

deploy/template.yaml
```diff
@@ -86,6 +86,7 @@ Mappings:
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
       templateCaching: true
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+      spotAccountNumber: "arn:aws:iam::738810260032:root"
```
chris-dco: May be more accurate to call this `spotAccountArn` as it's an ARN rather than a number. Also, the rest of the file does not add quotes around ARNs.

deploy/template.yaml
```diff
@@ -1220,6 +1271,46 @@ Resources:
...
+            Principal:
+              AWS: !FindInMap
+                - EnvironmentConfiguration
+                - !Ref AWS::AccountId
+                - spotAccountNumber
+            Action:
+              - "kms:Encrypt*"
+              - "kms:Decrypt*"
+            Resource: "*"
+          - Effect: Allow
```
chris-dco: Why do we need a DynamoDb policy?

deploy/template.yaml
```diff
@@ -1220,6 +1271,46 @@ Resources:
...
+    Properties:
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
+            Action:
+              - kms:*
```
chris-dco: Do we need such broad access?

deploy/template.yaml
```diff
@@ -1220,6 +1271,46 @@ Resources:
                   - "dynamodb.amazonaws.com"
                   - "ecs-tasks.amazonaws.com"
 
+  S3KmsKey:
```
chris-dco, line 1259: I think you may need to add SPOT to be able to use this KMS key so that it's able to encrypt and decrypt contents of the S3 bucket.

deploy/template.yaml
```diff
@@ -1220,6 +1271,46 @@ Resources:
...
+              - kms:*
+            Resource: "*"
+          - Effect: Allow
+            Principal:
+              AWS: !FindInMap
+                - EnvironmentConfiguration
+                - !Ref AWS::AccountId
+                - spotAccountNumber
+            Action:
+              - "kms:Encrypt*"
```
chris-dco: I think just `kms:Encrypt` and `kms:Decrypt` is sufficient.

deploy/template.yaml
chrisclayson: Agreed ... we should only grant what we need.

deploy/template.yaml
chrisclayson: Probably needs `kms:GenerateDataKey` too.

deploy/template.yaml
```diff
@@ -1203,22 +1255,31 @@ Resources:
...
+                - EnvironmentConfiguration
+                - !Ref AWS::AccountId
+                - spotAccountArn
             Action:
               - "kms:Encrypt*"
               - "kms:Decrypt*"
-              - "kms:ReEncrypt*"
-              - "kms:GenerateDataKey*"
-              - "kms:Describe*"
             Resource: "*"
```
chrisclayson, line 1281: Might be worth adding a `kms:ViaService` condition of `s3.amazonaws.com` here ... this will ensure we can _only_ use the key for encrypting/decryption from S3.

deploy/template.yaml
```diff
@@ -1203,22 +1255,31 @@ Resources:
...
+    Properties:
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: 2012-10-17
+        Statement:
           - Effect: Allow
             Principal:
-              Service: "dynamodb.amazonaws.com"
+              AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
+            Action:
```
chrisclayson, line 1269: I think the owning account probably needs `kms:*` otherwise CloudFormation won't be able to manage the key.

deploy/template.yaml
```diff
@@ -347,6 +353,52 @@ Resources:
...
+      Bucket: !Ref SpotKeysS3Bucket
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: AllowAccountAccess
+            Effect: Allow
+            Principal:
+              AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
+            Action:
+              - "s3:GetObject"
```
chrisclayson: I _think_ the owning account probably needs `s3:*` otherwise CloudFormation won't be able to manage the bucket.

deploy/template.yaml
chrisclayson: Actually, it turns out we don't need `kms:Encrypt` in this instance, see the [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html).

`kms:GenerateDataKey` is required for encrypting the data to upload to the bucket using `s3:PutObject`, `kms:Decrypt` is required to download using `s3:GetObject`.

deploy/template.yaml
sam-leen-gds: Fixed

deploy/template.yaml
sam-leen-gds: Yes, it should, it now is.

deploy/template.yaml
sam-leen-gds: Number -> Arn

deploy/template.yaml
sam-leen-gds: We don't, this has been removed.

deploy/template.yaml
sam-leen-gds: 👇 See Chris's comment below.

deploy/template.yaml
sam-leen-gds: Access policy for Spot on Kms key has been added.

deploy/template.yaml
sam-leen-gds: Spot account perms are now decrypt, and generateDataKey

deploy/template.yaml
sam-leen-gds: Condition to restrict usage to s3 crypto has been added.

deploy/template.yaml
sam-leen-gds: Owning account has been given blanket perms.

deploy/template.yaml
sam-leen-gds: Owning account has been given blanket perms.

chris-dco (approved):

blakeyp (approved):

deploy/template.yaml
chrisclayson: Minor point, you don't need the `*` in `kms:Decrypt*` ... it doesn't really matter because, at the moment, there is only one KMS API action that begins with `Decrypt` and that is `Decrypt` itself! However, if AWS added a new method called `kms:DecryptAllTheSecrets` you would be inadvertently be giving access to that method by including the `*` here. I think its best to avoid the `*` and be explicit.

PR #1373, opened 2024-04-04
PYIC-5600: Modify datastructures

MikeCollingwood:
### What changed

- Added for loops to Nunjucks
- Added PAGES enum constant
- Used path library for path-wrangling

Use path whenever possible, but if there are queries to add: use URL library, and if the path has no host defined, use URI.

### Why did it change

- DRY
- simplicity
- consistency

### Issue tracking

- [PYIC-5600](https://govukverify.atlassian.net/browse/PYIC-5600)

### Notes

- `safe` is a built-in Nunjucks filter: [safe](https://mozilla.github.io/nunjucks/templating.html#safe), [autoescaping](https://mozilla.github.io/nunjucks/templating.html#autoescaping)
_"Mark the value as safe which means that in an environment with automatic escaping enabled this variable will not be escaped."_
NB: None of the content edited is escapable

[PYIC-5600]: https://govukverify.atlassian.net/browse/PYIC-5600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2024-04-12 at 09 10 32
Screenshot 2024-04-12 at 09 10 25


src/app/credential-issuer/middleware.js
```diff
@@ -50,7 +60,7 @@ module.exports = {
     const body = {
       authorizationCode: req.query?.code,
       credentialIssuerId: criId,
-      redirectUri: `${EXTERNAL_WEBSITE_HOST}/credential-issuer/callback/${criId}`,
+      redirectUri: callbackPath,
```
blakeyp: is this missing `criId` on the end? (and should be `.href`?)

src/lib/locals.js
```diff
@@ -20,8 +21,15 @@ module.exports = {
...
-    res.locals.assetPath = CDN_DOMAIN + "/assets";
-    res.locals.contactUsUrl = `${CONTACT_URL}?fromURL=${SERVICE_URL}${req.originalUrl}`;
+    res.locals.assetPath = path.join(CDN_DOMAIN, "assets");
+
+    const contactUsUrl = new URL(CONTACT_URL);
+    contactUsUrl.searchParams.set(
+      "fromUrl",
+      `${SERVICE_URL}${req.originalUrl}`,
+    );
+    res.locals.contactUsUrl = contactUsUrl;
```
blakeyp: Does this need converting to a string?

src/lib/paths.js
```diff
@@ -0,0 +1,24 @@
...
+function getIpvPageTemplatePath(page) {
+  return path.join("ipv", "page", addNunjucksExt(page));
+}
+
+function addNunjucksExt(path) {
+  return `${path}.njk`;
+}
+
+module.exports = {
+  getRoutePath: getPostEndpoint,
```
blakeyp: any reason not to just call the function getRoutePath?

blakeyp (commented):
Nice 🚀

src/app/credential-issuer/middleware.js
MikeCollingwood: It shouldn't need `.href` since callbackPath is a string rather than URL type. Good point on the `criId`!

src/app/development/router.js
```diff
@@ -7,13 +7,14 @@
...
   templatesDisplayGet,
 } = require("./middleware");
+const path = require("path");
 
 const csrfProtection = csrf({});
 
-router.post("/all-templates", csrfProtection, allTemplatesPost);
-router.get("/all-templates", csrfProtection, allTemplatesGet);
+router.post(path.join("/", "all-templates"), csrfProtection, allTemplatesPost);
+router.get(path.join("/", "all-templates"), csrfProtection, allTemplatesGet);
```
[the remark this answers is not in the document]
MikeCollingwood: @blakeyp Do you know if these CodeQL errors can be solved?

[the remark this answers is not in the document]
blakeyp: I think we can safely ignore this, rate limiting for us should be done at the infra level not in code

MikeCollingwood:
E2E tests _only_ breaking due to not seeing VCs supposed to be produced by process-async-cri-credential lambda, which isn't getting invocated probably because of orch-stub F2F_QUEUE_NAME=stubQueue_F2FQueue_dev-mikec breakage. This means this PR works, but will also seek to get access to orch-stub account to fix

Dan C had access, so we went through it but it wasn't clear the difference between my dev account and the build account's setup.

blakeyp (dismissed):

blakeyp (approved):

PR #1374, opened 2024-04-04
PYIC-5483: Handle error on unknown backend response

lwebb-iig:
### What changed

- Adds error handling in `handleBackendResponse` for unknown responses

### Why did it change

- Prevents timeouts / doing nothing when receiving unrecognised responses

### Issue tracking

- [PYIC-5483](https://govukverify.atlassian.net/browse/PYIC-5483)

### Other considerations

[PYIC-5483]: https://govukverify.atlassian.net/browse/PYIC-5483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -134,7 +134,15 @@ async function handleBackendResponse(req, res, backendResponse) {
     );
   }
 
-  throw new Error(`Unknown backend response received ${backendResponse}`);
+  const err = new Error(`Unknown backend response received ${backendResponse}`);
+  err.status = HTTP_STATUS_CODES.UNAUTHORIZED;
+  logError(req, err);
+
+  req.session.currentPage = "pyi-technical";
+  res.status(HTTP_STATUS_CODES.UNAUTHORIZED);
+  return res.render("ipv/page/pyi-technical.njk", {
+    context: "unrecoverable",
+  });
 }
 
 function tryValidateCriResponse(criResponse) {
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -136,8 +136,8 @@ async function handleBackendResponse(req, res, backendResponse) {
 
   const message = {
     description: "Unexpected backend response",
-    data: backendResponse
-  }
+    data: backendResponse,
+  };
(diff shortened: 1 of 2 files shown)


src/app/ipv/middleware.js
```diff
@@ -133,6 +133,16 @@ async function handleBackendResponse(req, res, backendResponse) {
       `/ipv/page/${req.session.currentPage}`,
     );
   }
+
+  const err = new Error(`Unknown backend response received ${backendResponse}`);
+  err.status = HTTP_STATUS_CODES.UNAUTHORIZED;
```
Joe-Edwards-GDS: Is this status code right? I think it's probably a 500 if the backend is returning a success response with an unexpected payload.

Can we just throw the `Error` here and allow the `internal-server-error-handler` deal with it? I don't think we should need special logic for this case (and that'll default to a 500).

src/app/ipv/middleware.js
```diff
@@ -133,6 +133,16 @@ async function handleBackendResponse(req, res, backendResponse) {
       `/ipv/page/${req.session.currentPage}`,
     );
   }
+
+  const err = new Error(`Unknown backend response received ${backendResponse}`);
```
Joe-Edwards-GDS: Can we include the `backendResponse` as JSON? I think this will likely end up as `Unknown backend response received [object Object]`, which isn't very useful!

I think we probably either log a dedicated message for this, e.g.
```
const message = {
  description: "Unexpected backend response",
  data: backendResponse
}
req.log.error({ message, level: "ERROR" });
throw new Error(message.description);
```
or we can try including the response as a property on the error itself, and rely on the error handler to include it:
```
const err = new Error("Unexpected backend response");
err.data = backendResponse;
throw err;
```

(Either way it'd be nice to test, and see how comprehensible the log messages are 🙂)

src/app/ipv/middleware.js
lwebb-iig: Agreed - I'll try out throwing an error directly instead

Joe-Edwards-GDS (approved):

PR #1376, opened 2024-04-04
PYIC-5403: Use application/json for core-back APIs

Joe-Edwards-GDS:
### What changed

- Commonised all core-back API calls to use `application/json`
- Changed 'action' API into 'journey event'

### Why did it change

There were a mixture of JSON and x-www-formurlencoded endpoints. Mostly they don't send a request body anyway, so the content-type didn't matter.

### Issue tracking

- [PYIC-5403](https://govukverify.atlassian.net/browse/PYIC-5403)

### Other considerations

This must not go live until https://github.com/govuk-one-login/ipv-core-back/pull/1804 is merged

[PYIC-5403]: https://govukverify.atlassian.net/browse/PYIC-5403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -44,13 +44,17 @@ async function journeyApi(action, req) {
     action = action.substr(1);
   }
 
+  if (action.startsWith("journey/")) {
+    action = action.substr(8);
+  }
+
   logCoreBackCall(req, {
     logCommunicationType: LOG_COMMUNICATION_TYPE_REQUEST,
     type: LOG_TYPE_JOURNEY,
     path: action,
   });
 
-  return coreBackService.postAction(req, action);
+  return coreBackService.postJourneyEvent(req, action);
 }
 
 async function fetchUserDetails(req) {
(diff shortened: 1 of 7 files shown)


src/services/coreBackService.js
```diff
@@ -3,42 +3,52 @@ const {
...
-  generateAxiosConfigWithClientSessionId,
-  generateJsonAxiosConfig,
-  createAxiosInstance,
-} = require("../app/shared/axiosHelper");
+const { createAxiosInstance } = require("../app/shared/axiosHelper");
 
 const axiosInstance = createAxiosInstance(API_BASE_URL);
 
-function postAction(req, action) {
+function generateAxiosConfig(req) {
```
Joe-Edwards-GDS, line 13: One small consequence of this - we will now send the two session ids to all endpoints (including `initialiseIpvSession`) if they are available. I don't think this is a big deal, but we could explicitly control it if we wish.

src/app/ipv/middleware.js
```diff
@@ -44,13 +44,17 @@ async function journeyApi(action, req) {
     action = action.substr(1);
   }
 
+  if (action.startsWith("journey/")) {
+    action = action.substr(8);
+  }
```
Joe-Edwards-GDS, line 49: We should really be able to get rid of this, but I think we'll still get a `/journey/` prefix on journey events coming from the backend. (In future we should be able to eliminate these entirely - it's only really error scenarios that do this...)

src/services/coreBackService.js
Wynndow: Why move this out of axiosHelper?

Wynndow (dismissed):

src/services/coreBackService.js
Joe-Edwards-GDS: This config feels quite specific to core-back requests, rather than generic axios configuration.

(E.g. if we started calling another API, we probably wouldn't be able to reuse this)

Wynndow (approved):

PR #1380, opened 2024-04-05
BAU: Added missing context translation

RogVanGDS:
### What changed

Added missing context translation

### Why did it change

Missed from previous commit

### Issue tracking

### DIFF
--- a/src/views/ipv/page/pyi-no-match.njk
+++ b/src/views/ipv/page/pyi-no-match.njk
@@ -11,7 +11,7 @@
     <p class="govuk-body">{{ 'pages.pyiNoMatch.content.paragraph2' | translate }}</p>
   {% endif %}
   <h2 class="govuk-heading-m">{{ 'pages.pyiNoMatch.content.subHeading' | translate }}</h2>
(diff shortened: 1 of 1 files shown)


SreeBandla01 (approved):

src/views/ipv/page/pyi-no-match.njk
```diff
@@ -11,7 +11,7 @@
     <p class="govuk-body">{{ 'pages.pyiNoMatch.content.paragraph2' | translate }}</p>
   {% endif %}
   <h2 class="govuk-heading-m">{{ 'pages.pyiNoMatch.content.subHeading' | translate }}</h2>
-  <p class="govuk-body">{{ 'pages.pyiNoMatch.content.paragraph3' | translate }}</p>
+  <p class="govuk-body">{{ 'pages.pyiNoMatch.content.paragraph3' | translateWithContext(context) }}</p>
```
Joe-Edwards-GDS, line 14: Do we need to provide a fallback here as well? I don't think we have a variant for all the contexts

RogVanGDS: Ah yeah good spot 🤦

PR #1384, opened 2024-04-10
PYIC-5537: Refactored journey event logic for event validation

RogVanGDS:
Opened a fresh PR to make it easier given the amount of merge conflicts

### What changed

Refactored the journey event posting from core-front to core-back to include a current page search parameter

### Why did it change

We need to make this change to give core-back the information required to apply some journey event validation. This change will not validate anything by itself, but should let us start validating without breaking in-flight journeys once the backend changes are ready.

### Issue tracking

- [PYIC-5537](https://govukverify.atlassian.net/browse/PYIC-5537)

[PYIC-5537]: https://govukverify.atlassian.net/browse/PYIC-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/ipv/middleware.js
```diff
@@ -64,8 +64,16 @@ async function fetchUserDetails(req) {
   return generateUserDetails(userDetailsResponse, req.i18n);
 }
 
-async function handleJourneyResponse(req, res, action) {
-  const backendResponse = (await journeyApi(action, req)).data;
+async function handleJourneyResponse(req, res, action, currentPage = "") {
+
+  if (currentPage) {
+    action += `?currentPage=${currentPage}`
+  }
```
RogVanGDS: I'm hoping that Mike's changes can incorporate this, or if he merges in first then I will update to use his logic

src/app/ipv/middleware.js
```diff
@@ -64,8 +64,13 @@ async function fetchUserDetails(req) {
   return generateUserDetails(userDetailsResponse, req.i18n);
 }
 
-async function handleJourneyResponse(req, res, action) {
+async function handleJourneyResponse(req, res, action, currentPage = "") {
+  if (currentPage) {
+    action += `?currentPage=${currentPage}`;
```
Joe-Edwards-GDS: It'd be preferable to use axios to set the query params - that way it'll handle stuff like escaping and building the query string for you.

i.e. adding `{ params: { currentPage } }` to the axios config for the request.

src/app/ipv/middleware.js
```diff
@@ -64,8 +64,13 @@ async function fetchUserDetails(req) {
   return generateUserDetails(userDetailsResponse, req.i18n);
 }
 
-async function handleJourneyResponse(req, res, action) {
+async function handleJourneyResponse(req, res, action, currentPage = "") {
```
Joe-Edwards-GDS: Is this default value useful - could it just default to `undefined`?

I suppose it's useful to indicate that it's optional, in lieu of typescript :/

src/app/ipv/middleware.js
Joe-Edwards-GDS: Matter of opinion, but I think I'd slightly prefer `currentPageId` or just `pageId`? Feels more precise about what exactly the value is.

src/app/ipv/middleware.js
```diff
@@ -228,16 +232,12 @@ function handleAppStoreRedirect(req, res, next) {
...
-      case PHONE_TYPES.ANDROID:
-        res.redirect(APP_STORE_URL_ANDROID);
-        break;
-      default:
-        throw new Error("Unrecognised phone type: " + specifiedPhoneType);
+    if (specifiedPhoneType === PHONE_TYPES.IPHONE) {
+      res.redirect(APP_STORE_URL_APPLE);
+    } else if (specifiedPhoneType === PHONE_TYPES.ANDROID) {
+      res.redirect(APP_STORE_URL_ANDROID);
     }
```
Joe-Edwards-GDS: Careful - I think this reintroduces the same bug that we used to have. You need to return from each of these blocks or the control flow will continue through to the error.

I'd suggest that we do `return res.redirect(...)` so we don't fall through to any other branches

(I also think the switch is more readable, unless there was a particular reason to change it?)

Joe-Edwards-GDS (commented):
A handful of comments, but I think this is broadly looking good :)

It should hopefully also be clear for @sammyiyke how to commonise all the different routes.

src/app/ipv/middleware.js
RogVanGDS: Ah I think this was a rebasing conflict that went wrong, I'll revert

src/app/ipv/middleware.js
RogVanGDS: Yeah I thought it was worth showing that this is optional, it always feels a bit wrong to me to just ignore a parameter with no override

src/services/coreBackService.js
```diff
@@ -28,10 +28,10 @@ function generateAxiosConfig(req) {
   };
 }
 
-function postJourneyEvent(req, event) {
+function postJourneyEvent(req, event, queryParams = {}) {
   return axiosInstance.post(
     `${API_JOURNEY_EVENT}/${event}`,
-    {},
+    queryParams,
```
Joe-Edwards-GDS: Have you tested this? I think the second argument here is the request body, not query params.

You might need to do something like:
```
function postJourneyEvent(req, event, currentPage) {
  return axiosInstance.post(
    `${API_JOURNEY_EVENT}/${event}`,
    {},
    {
      ...generateAxiosConfig(req),
      params: currentPage ? { currentPage } : {},
  );
```

src/services/coreBackService.js
```diff
@@ -28,10 +28,10 @@ function generateAxiosConfig(req) {
   };
 }
 
-function postJourneyEvent(req, event) {
+function postJourneyEvent(req, event, queryParams = {}) {
```
Joe-Edwards-GDS: It feels a bit odd to pass arbitrary query params here - I think this method should capture the details of the API interface, which I think includes the specific parameters we expect (i.e. `currentPage`).

src/app/ipv/middleware.js
Joe-Edwards-GDS: (I actually meant for the actual query param, rather than just this variable, but I don't feel strongly if you want to avoid making the change all the way through the stack)

src/services/coreBackService.js
RogVanGDS: Ahh the build failed 🤦  thanks I'll change now

src/services/coreBackService.js
```diff
@@ -28,12 +28,14 @@ function generateAxiosConfig(req) {
...
 }
 
-function postJourneyEvent(req, event, queryParams = {}) {
-  return axiosInstance.post(
-    `${API_JOURNEY_EVENT}/${event}`,
-    queryParams,
-    generateAxiosConfig(req),
-  );
+function postJourneyEvent(req, event, currentPage) {
+  const queryParams = generateAxiosConfig(req);
```
Joe-Edwards-GDS: This should probably `requestConfig` or similar

Joe-Edwards-GDS (approved):

PR #1385, opened 2024-04-11
PYIC-5436 - Tidy up routing for journey action submissions

sammyiyke:
### What changed

[PYIC-5436](https://govukverify.atlassian.net/browse/PYIC-5436) - Tidy up routing for journey action submissions

Makes the generic journey action function `handleJourneyAction` robust to handle all `POST` journey avent

Major changes:

- Refactors `handleJourneyAction`
- Removes `handleEscapeAction` and merges it function into `handleJourneyAction`
- Removes `handleEscapeM2b` and merges it function into `handleJourneyAction`
- Removes `handleMultipleDocCheck` and merges it function into `handleJourneyAction`
- Removes `handleUpdateNameDobAction` and merges it function into `handleJourneyAction`

Major tests related changes:

Removes redundant tests for defaulting to `end` when no journey action is provided in form

Reasons:

- Backend validation added in PYIC-5537 removing need to default to `end`
- Previous related functions `handleEscapeAction`, `handleEscapeM2b `,  `handleMultipleDocCheck `, and `handleUpdateNameDobAction` merged into `handleJourneyAction`

### Why did it change

We want to make the generic journey action `handleJourneyAction` robust for all events, so that:

- The form submits the appropriate event - next, f2f etc.

- The handler maps that into a request to the backend POST /journey/next etc.

### Issue tracking

- [PYIC-5436](https://govukverify.atlassian.net/browse/PYIC-5436)

### Other considerations

[PYIC-5436]: https://govukverify.atlassian.net/browse/PYIC-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-5436]: https://govukverify.atlassian.net/browse/PYIC-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -345,13 +345,19 @@ module.exports = {
   // make considerations for handleEscapeAction pages with no req.body?.journey that defaults to `end`
   handleJourneyAction: async (req, res, next) => {
     const currentPage = req.params.pageId;
-    const pagesUsingSessionId = ["pyi-suggest-other-options", "pyi-cri-escape", "pyi-kbv-escape-m2b", "pyi-escape-m2b", "page-multiple-doc-check"];
+    const pagesUsingSessionId = [
+      "pyi-suggest-other-options",
+      "pyi-cri-escape",
+      "pyi-kbv-escape-m2b",
+      "pyi-escape-m2b",
+      "page-multiple-doc-check",
+    ];
     const journeyActions = {
-      "end": "end",
+      end: "end",
       "address-current": "address-current",
       "attempt-recovery": "attempt-recovery",
       "build-client-oauth-response": "build-client-oauth-response",
-      "next": "next",
+      next: "next",
       "next/end": "end",
       "next/f2f": "f2f",
       "next/dcmaw": "dcmaw",
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -345,7 +345,7 @@ module.exports = {
   // make considerations for handleEscapeAction pages with no req.body?.journey that defaults to `end`
   handleJourneyAction: async (req, res, next) => {
     const currentPage = req.params.pageId;
-    const pagesUsingSessionId = ["pyi-suggest-other-options", "pyi-cri-escape", "pyi-kbv-escape-m2b", "pyi-escape-m2b"];
+    const pagesUsingSessionId = ["pyi-suggest-other-options", "pyi-cri-escape", "pyi-kbv-escape-m2b", "pyi-escape-m2b", "page-multiple-doc-check"];
     const journeyActions = {
       "end": "end",
       "address-current": "address-current",
@@ -354,6 +354,8 @@ module.exports = {
       "next/f2f": "f2f",
       "next/dcmaw": "dcmaw",
       "next/bank-account": "bankAccount",
+      "next/passport": "ukPassport",
+      "next/driving-licence": "drivingLicence",
     };
     try {
       const action = journeyActions[req.body?.journey] || "next";
@@ -370,23 +372,6 @@ module.exports = {
       }
 
       await handleJourneyResponse(req, res, action, currentPage);
-    } catch (error) {
-      transformError(error, `error handling POST request on ${currentPageId}`);
-      next(error);
-    }
-  },
-
-  handleMultipleDocCheck: async (req, res, next, currentPageId) => {
-    try {
-      checkForSessionId(req, res);
-
-      if (req.body?.journey === "next/passport") {
-        await handleJourneyResponse(req, res, "ukPassport", currentPageId);
-      } else if (req.body?.journey === "next/driving-licence") {
-        await handleJourneyResponse(req, res, "drivingLicence", currentPageId);
-      } else {
-        await handleJourneyResponse(req, res, "end", currentPageId);
-      }
     } catch (error) {
       transformError(error, `error handling POST request on ${currentPageId}`);
       next(error);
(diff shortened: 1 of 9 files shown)


src/app/ipv/middleware.js
```diff
@@ -358,75 +341,54 @@ module.exports = {
...
+      "attempt-recovery": "attempt-recovery",
+      "build-client-oauth-response": "build-client-oauth-response",
+      next: "next",
+      "next/end": "end",
+      "next/f2f": "f2f",
+      "next/dcmaw": "dcmaw",
+      "next/bank-account": "bankAccount",
+      "next/passport": "ukPassport",
+      "next/driving-licence": "drivingLicence",
+    };
```
sammyiyke: Probably best taking this out of here into a separate file probably in the shared folder. Leaving it here in the meantime until I get feedback on the new implementation

src/app/ipv/middleware.js
```diff
@@ -358,87 +341,54 @@ module.exports = {
...
+      "address-current": "address-current",
+      "attempt-recovery": "attempt-recovery",
+      "build-client-oauth-response": "build-client-oauth-response",
+      next: "next",
+      "next/end": "end",
+      "next/f2f": "f2f",
+      "next/dcmaw": "dcmaw",
+      "next/bank-account": "bankAccount",
+      "next/passport": "ukPassport",
+      "next/driving-licence": "drivingLicence",
```
RogVanGDS: I was hoping that we might be able to do away with all this mapping, given we are now validating journey events on the backend, and posting some invalid journey events isn't as risky as it was before

I expect we will still need a list of pages that need to default to back instead of next, but we should be able to either change the radio button forms so that the journey event is already in the right format, or update the journey map to match if we need to

src/app/ipv/middleware.js
```diff
@@ -358,87 +341,54 @@ module.exports = {
...
-      if (req.body?.journey === "next/passport") {
-        await handleJourneyResponse(req, res, "ukPassport", currentPageId);
-      } else if (req.body?.journey === "next/driving-licence") {
-        await handleJourneyResponse(req, res, "drivingLicence", currentPageId);
+      const action = journeyActions[req.body?.journey] || "next";
+      if (pagesUsingSessionId.includes(currentPageId)) {
+        checkForSessionId(req, res);
       } else {
-        await handleJourneyResponse(req, res, "end", currentPageId);
+        checkForIpvAndOauthSessionId(req, res);
```
Joe-Edwards-GDS, line 360: Do you know what this distinction is based on? I can't see any obvious basis for the difference.

I'm wondering if we can just use `checkForIpvAndOauthSessionId` for everything, and then you don't need the list of pages to distinguish :)

src/app/ipv/middleware.js
```diff
@@ -358,87 +341,54 @@ module.exports = {
...
-      if (req.body?.journey === "next") {
-        await handleJourneyResponse(req, res, "next", currentPageId);
-      } else if (req.body?.journey === "next/bank-account") {
-        await handleJourneyResponse(req, res, "bankAccount", currentPageId);
-      } else {
-        await handleJourneyResponse(req, res, "end", currentPageId);
+      if (action === "build-client-oauth-response") {
+        req.session.ipAddress = req?.session?.ipAddress
+          ? req.session.ipAddress
+          : getIpAddress(req);
```
Joe-Edwards-GDS, line 366: This seems suspicious, but I'm scared of changing anything in analytics without fully understanding it! (I'm a bit surprised we don't just set the IP address on every request independently...)

Leave this in for now :)

src/app/ipv/middleware.js
```diff
@@ -358,87 +341,54 @@ module.exports = {
       delete req.session.currentPageStatusCode;
     }
   },
+
+  // make considerations for handleEscapeAction pages with no req.body?.journey that defaults to `end`
```
Joe-Edwards-GDS: Can we delete this comment?

src/app/ipv/middleware.js
Joe-Edwards-GDS: Do we need this mapping? Can we instead update the page templates to pass the correct event (e.g. updating `pyi-escape-m2b-form.njk` to just have `bankAccount`)

Joe-Edwards-GDS (commented):
A couple of comments but I think this is looking much neater!

src/app/ipv/middleware.js
sammyiyke: Good shout. I do agree, but was optimistic about changing it earlier.

src/app/ipv/middleware.js
sammyiyke: I didn't look that deep into it. But on the surface I figure that some pages **require** only the ipvSession ID while some **require** both the ipvSession and the OauthSession ID. And they fail if any of the required session ID is not present. 

I am not exactly sure why, as I think by default journeys should have both IDs present at each stage. 

I would look deeper and see what the reasoning behind the differentiation is and if changing it would break any test

Joe-Edwards-GDS: I think this method looks for the `ipvSessionId` OR `clientOauthSessionId` (not AND), so I was hoping we could just use that for everything.

It's fine to leave, since you've not changed the logic, but I suspect it isn't quite right.

Joe-Edwards-GDS (approved):

PR #1386, opened 2024-04-11
PYIC-5849: Tweaked RFC content following UCD review

RogVanGDS:
### What changed

Tweaked RFC content following UCD review

### Why did it change

A couple of changes requested after UCD reviewed the journey

### Issue tracking

- [PYIC-5849](https://govukverify.atlassian.net/browse/PYIC-5849)

[PYIC-5849]: https://govukverify.atlassian.net/browse/PYIC-5849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -255,7 +255,7 @@ module.exports = {
   // https://govukverify.atlassian.net/browse/PYIC-4859.
   updateJourneyState: async (req, res, next) => {
     try {
-      const allowedActions = ["/journey/end"];
+      const allowedActions = ["/journey/end", "/journey/back"];
 
       const action = allowedActions.find((x) => x === req.url);
 
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -255,7 +255,7 @@ module.exports = {
   // https://govukverify.atlassian.net/browse/PYIC-4859.
   updateJourneyState: async (req, res, next) => {
     try {
-      const allowedActions = ["/journey/end", "/journey/back"];
+      const allowedActions = ["/journey/end"];
 
       const action = allowedActions.find((x) => x === req.url);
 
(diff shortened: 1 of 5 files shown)


src/app/ipv/middleware.js
```diff
@@ -255,7 +255,7 @@ module.exports = {
   // https://govukverify.atlassian.net/browse/PYIC-4859.
   updateJourneyState: async (req, res, next) => {
     try {
-      const allowedActions = ["/journey/end"];
+      const allowedActions = ["/journey/end", "/journey/back"];
```
RogVanGDS: Will probably need to refactor this into the back button having it's own route when the validation work gets merged in, but I think it's an ok solution for now

Joe-Edwards-GDS: Hmm yes - we might need to think about this one. Do we need to worry about the `/journey/end` case for the validation work as well - have you tested that scenario?

Perhaps ideally the back link would include the `currentPage` as well, so it's accessible here.

It's probably ok to merge before that, if this is higher priority. It means you can submit a `back` event for any page, but we don't handle that event for most states anyway.

RogVanGDS: Yep I have tested the /journey/end for that scenario, it makes this logic safer because we can then only allow that event on the specific page that this route is built for.

I will have the routing to pass through the current page, but this is higher priority for the time being.

Hopefully once the validation goes in, then we won't be able to just submit a back event for any page 😄

Joe-Edwards-GDS: Ok - I think the comment above this method is no longer true with this change, so we probably need to update that, and update the validation PR accordingly (so it doesn't pass a hardcoded `pyi-f2f-delete-details`).

Joe-Edwards-GDS (dismissed):
I'll approve so you can merge if this is urgent, but that probably invalidates the other PR

Joe-Edwards-GDS (approved):

PR #1392, opened 2024-04-12
PYIC-5624: Add strategic app triage journeys and device sniffing

MikeCollingwood:
### What changed

- Enable journey routing for dcmaw triage behind a feature flag/ set: strategicApp 
- Added device sniffing
- Integrated fixed redirect urls

One journey example:

Screenshot 2024-04-23 at 11 22 04

Screenshot 2024-04-23 at 11 22 17

Screenshot 2024-04-23 at 11 22 28
Screenshot 2024-04-23 at 11 22 38

### Issue tracking

- [PYIC-4816](https://govukverify.atlassian.net/browse/PYIC-4816)
- [PYIC-4819](https://govukverify.atlassian.net/browse/PYIC-4819)
- [PYIC-5624](https://govukverify.atlassian.net/browse/PYIC-5624)

### Testing

> Tested with the changes in core-back

It passes E2E tests, and manual testing behind the feature flag, though will need new tests for these journeys

Screenshot 2024-04-23 at 11 17 29

[PYIC-5624]: https://govukverify.atlassian.net/browse/PYIC-5624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PYIC-4816]: https://govukverify.atlassian.net/browse/PYIC-4816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-4819]: https://govukverify.atlassian.net/browse/PYIC-4819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/ipv/middleware.test.js
```diff
@@ -953,9 +948,9 @@ describe("journey middleware", () => {
       };
     });
 
-    // PYIC-4816 Update tests to get iphone/android from session.
     it("sets an iPhone qrCode value for the page", async function () {
```
DanCorderIPV, line 1079: Probably worth adding a similar test for android here

src/app/ipv/middleware.test.js
```diff
@@ -983,9 +978,9 @@ describe("journey middleware", () => {
       };
     });
 
-    // PYIC-4816 Update tests to get iphone/android from session.
     it("sets an Android appDownloadUrl value for the page", async function () {
```
DanCorderIPV, line 1125: Probably worth adding a similar test for iphone here

src/app/ipv/middleware.js
```diff
@@ -224,7 +193,6 @@ function isValidPage(pageId) {
 }
 
 function handleAppStoreRedirect(req, res, next) {
-  // PYIC-4816 - consider whether we should override this with the current request's sniffed device type.
```
DanCorderIPV, line 216: Was this discussed with anyone?

src/app/ipv/middleware.js
```diff
@@ -371,63 +326,16 @@ module.exports = {
...
-  handleMultipleDocCheck: async (req, res, next, currentPageId) => {
-    try {
-      checkForSessionId(req, res);
-
-      if (req.body?.journey === "next/passport") {
-        await handleJourneyResponse(req, res, "ukPassport", currentPageId);
-      } else if (req.body?.journey === "next/driving-licence") {
-        await handleJourneyResponse(req, res, "drivingLicence", currentPageId);
-      } else {
-        await handleJourneyResponse(req, res, "end", currentPageId);
```
DanCorderIPV: The default case here is `end` but as far as I can see with the new code the default will be `next` is that deliberate?
Same in `handleEscapeM2b` and `handleEscapeAction`

src/app/ipv/middleware.js
```diff
@@ -371,63 +326,16 @@ module.exports = {
...
-      next(error);
-    }
-  },
+      req.session.ipAddress = req?.session?.ipAddress
+        ? req.session.ipAddress
+        : getIpAddress(req);
 
-  handleMultipleDocCheck: async (req, res, next, currentPageId) => {
-    try {
-      checkForSessionId(req, res);
```
DanCorderIPV: Do we not need to do this anymore?

src/app/ipv/middleware.test.js
```diff
@@ -562,17 +562,17 @@ describe("journey middleware", () => {
...
 
   context(
-    "handleMultipleDocCheck: handling journey action with ukPassport, drivingLicence, end",
+    "handleJourneyAction: handling journey action with ukPassport, drivingLicence, end",
     () => {
       it("should postJourneyEvent with ukPassport", async function () {
         req = {
           id: "1",
-          body: { journey: "next/passport" },
+          body: { journey: "ukPassport" },
```
DanCorderIPV: Do we need to sync this change with core-back?

src/app/ipv/router.js
```diff
@@ -41,43 +38,6 @@ router.post(
   (req, res, next) =>
     handleUpdateNameDobAction(req, res, next, "update-name-date-birth"),
 );
```
DanCorderIPV: This seems like quite a big change that isn't really related to this ticket and doesn't appear to keep exactly the same behaviour.
That worries me that either the change in behaviour is accidental, or this should be it's own ticket with related testing requirements.

src/views/ipv/page/confirm-address.njk
```diff
@@ -48,7 +48,7 @@
             },
             items: [
                     {
-                    value: "addressCurrent",
+                    value: "address-current",
```
DanCorderIPV: In terms of consistency why are we adding a hyphen here but removing one from `driving-licence` below?

DanCorderIPV (changes requested):

src/app/ipv/router.js
MikeCollingwood: This was intentional refactoring whilst I was dealing with the dilemma of creating a new handler (for each page that can lead to dcmaw) - turns out a lot of the handlers aren't needed and so I thought I could include a simplifying refactor into this PR

src/views/ipv/page/confirm-address.njk
MikeCollingwood: This is because these reflect the format of the corresponding journey event

src/app/ipv/middleware.test.js
MikeCollingwood: Ah, no since the ukPassport is the event core-back wants - it's just the `handleMultipleDocCheck` would take `next/passport` and call core-back with `ukPassport`

src/app/ipv/middleware.js
MikeCollingwood: In the use cases I've seen, the "next" or "end" event is actually provided in the `req.body?.journey` - so I should probably remove the default in this line of thinking

src/app/ipv/middleware.js
MikeCollingwood: Would like to go through this with you because I didn't see the benefit - but would be good to clarify what you mean

src/app/ipv/middleware.js
```diff
@@ -371,63 +326,16 @@ module.exports = {
...
-        await handleJourneyResponse(req, res, "next", currentPageId);
-      } else if (req.body?.journey === "next/bank-account") {
-        await handleJourneyResponse(req, res, "bankAccount", currentPageId);
-      } else {
-        await handleJourneyResponse(req, res, "end", currentPageId);
-      }
+      await handleJourneyResponse(
+        req,
+        res,
+        req.body?.journey || "next",
```
MikeCollingwood: Do we want the default next at all?

src/app/ipv/router.js
```diff
@@ -38,6 +41,43 @@ router.post(
...
+  formRadioButtonChecked,
+  (req, res, next) => handleEscapeAction(req, res, next, "pyi-cri-escape"),
+);
+router.post(
+  "/page/pyi-suggest-other-options",
+  parseForm,
+  csrfProtection,
+  formRadioButtonChecked,
+  (req, res, next) =>
+    handleEscapeAction(req, res, next, "pyi-suggest-other-options"),
```
MikeCollingwood: The page `pyi-suggest-other-options` existed on states which did't use the "end" journey event

src/app/shared/appDownloadHelper.js
```diff
@@ -1,22 +1,14 @@
...
-  if (phoneType === PHONE_TYPES.IPHONE) {
-    return browserUrl;
-  } else if (phoneType === PHONE_TYPES.ANDROID) {
-    return `intent://${SERVICE_DOMAIN}/ipv/app-redirect/#Intent;scheme=https;package=${ANDROID_APP_ID};S.browser_fallback_url=${encodeURIComponent(browserUrl)};end`;
+  if (phoneType !== PHONE_TYPES.IPHONE && phoneType !== PHONE_TYPES.ANDROID) {
+    throw new Error(`Unrecognised phone type: ${phoneType}`);
   }
 
-  throw new Error(`Unrecognised phone type: ${phoneType}`);
+  return SERVICE_URL + "/ipv/app-redirect/" + phoneType;
```
MikeCollingwood, line 9: Use path.join

src/app/shared/appDownloadHelper.js
```diff
@@ -1,22 +1,15 @@
...
-  if (phoneType === PHONE_TYPES.IPHONE) {
-    return browserUrl;
-  } else if (phoneType === PHONE_TYPES.ANDROID) {
-    return `intent://${SERVICE_DOMAIN}/ipv/app-redirect/#Intent;scheme=https;package=${ANDROID_APP_ID};S.browser_fallback_url=${encodeURIComponent(browserUrl)};end`;
+  if (phoneType !== PHONE_TYPES.IPHONE && phoneType !== PHONE_TYPES.ANDROID) {
+    throw new Error(`Unrecognised phone type: ${phoneType}`);
   }
 
-  throw new Error(`Unrecognised phone type: ${phoneType}`);
+  return path.join(SERVICE_URL || "/", "ipv", "app-redirect", phoneType);
```
Joe-Edwards-GDS: Careful - `path` is for file paths, not URLs.

On Windows this will produce `\ipv\app-redirect\android` (with backslashes, not forward!)

src/app/ipv/middleware.js
```diff
@@ -213,7 +222,6 @@ function isValidPage(pageId) {
 }
 
 function handleAppStoreRedirect(req, res, next) {
-  // PYIC-4816 - consider whether we should override this with the current request's sniffed device type.
```
DanCorderIPV, line 216: For the record, this now has it's own ticket PYIC-5560 so this comment isn't needed

src/app/ipv/middleware.js
```diff
@@ -322,31 +321,27 @@ module.exports = {
       const renderOptions = {
         pageId,
         csrfToken: req.csrfToken(),
+        dcmawEvent: req.isSmartphoneUser ? "smartphone" : "dcmaw",
         context,
       };
+      const phoneType = {
+        iphone: PHONE_TYPES.IPHONE,
+        android: PHONE_TYPES.ANDROID,
+      }[context];
```
DanCorderIPV: This works but every time I read it I do a double take. I think it would seem less odd if the object here was declared separately and called `contextToPhoneTypeMapping` or something similar.

src/app/development/middleware.js
```diff
@@ -69,11 +69,13 @@ async function templatesDisplayGet(req, res) {
   }
   if (templateId === PAGES.PYI_TRIAGE_DESKTOP_DOWNLOAD_APP) {
     renderOptions.qrCode = await qrCodeHelper.generateQrCodeImageData(
-      appDownloadHelper.getAppStoreRedirectUrl(PHONE_TYPES.IPHONE),
+      appDownloadHelper.getAppStoreRedirectUrl(
+        context !== "android" ? PHONE_TYPES.IPHONE : PHONE_TYPES.ANDROID,
```
DanCorderIPV: I think it would be good to have constants for the possible context values rather than using strings directly.

src/app/ipv/middleware.js
```diff
@@ -364,25 +359,25 @@ module.exports = {
...
-      "page-multiple-doc-check",
+      PAGES.PYI_SUGGEST_OTHER_OPTIONS,
+      PAGES.PYI_CRI_ESCAPE,
+      PAGES.PYI_KBV_ESCAPE_M2B,
+      PAGES.PYI_ESCAPE_M2B,
+      PAGES.PAGE_MULTIPLE_DOC_CHECK,
     ];
 
     try {
-      const action = req.body?.journey || "next";
```
DanCorderIPV, line 375: As discussed, you've checked that the default shouldn't ever be used so this should be safe.

src/app.js
```diff
@@ -130,6 +131,12 @@ app.use((req, res, next) => {
   }
 });
 
+app.use((req, res, next) => {
+  const parser = new UAParser(req.headers["user-agent"]);
+  req.isSmartphoneUser = parser.getDevice()["type"] === "mobile";
+  next();
+});
```
blakeyp: Do we need this to be middleware that applies to every request?

src/app/development/middleware.js
blakeyp: Not a huge deal but this would be easier to read if it was a positive condition `context === "android" ? ANDROID : IPHONE` 😃

src/app/ipv/middleware.js
blakeyp, line 375: Are there any implications to this no longer defaulting to `next` when undefined 🤔

src/app/ipv/middleware.js
```diff
@@ -273,15 +281,6 @@ module.exports = {
       next(error);
     }
   },
-  handlePageBackButton: async (req, res, next) => {
```
blakeyp, line 276: Is this not used?

src/app/shared/appDownloadHelper.js
```diff
@@ -1,22 +1,14 @@
...
 
 // Returns the url of the app redirect route for the requested phone type. For Android phones the url is wrapped in an
 // intent so the app is launched if it is already installed on the device.
 function getAppStoreRedirectUrl(phoneType) {
-  const browserUrl = SERVICE_URL + "/ipv/app-redirect/" + phoneType;
-
-  if (phoneType === PHONE_TYPES.IPHONE) {
-    return browserUrl;
-  } else if (phoneType === PHONE_TYPES.ANDROID) {
-    return `intent://${SERVICE_DOMAIN}/ipv/app-redirect/#Intent;scheme=https;package=${ANDROID_APP_ID};S.browser_fallback_url=${encodeURIComponent(browserUrl)};end`;
```
blakeyp, line 16: This previously set the intent url for Android, are we handling this in a separate ticket?

DanCorderIPV (dismissed):

src/app/shared/appDownloadHelper.js
DanCorderIPV: I think this is covered by https://govukverify.atlassian.net/browse/PYIC-5668
I think the intent should in the redirect URL anyway and not here

src/app/shared/appDownloadHelper.js
```diff
@@ -1,22 +1,14 @@
-const {
-  SERVICE_URL,
-  SERVICE_DOMAIN,
-  ANDROID_APP_ID,
-} = require("../../lib/config");
+const { SERVICE_URL } = require("../../lib/config");
 const PHONE_TYPES = require("../../constants/phone-types");
 
 // Returns the url of the app redirect route for the requested phone type. For Android phones the url is wrapped in an
 // intent so the app is launched if it is already installed on the device.
```
DanCorderIPV: This comment is out of date - please delete

src/app/ipv/middleware.js
MikeCollingwood: Thankfully there are no cases this applies to and tests pass with throwing an error whenever there's an undefined action, confirming it's ok

blakeyp (approved):

PR #1393, opened 2024-04-15
PYIC-5636

sam803:
## Proposed changes
 - Adds an error prefix to the html title if there is an error in a page
Screenshot 2024-04-15 at 14 32 16

### What changed
- adds a `govuk.errorTitlePrefix` value to the to translation files
- use `pageErrorState` instead of `error/errors` to determine if an error has occurred

### Why did it change
For accessibilty issues, if an error occurs we should prefix the page title with 'Error: '
- `govuk.errorTitlePrefix` was being referenced but is not defined
-  `error/errors` we being referenced but were not defined - use `pageErrorState` instead

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-5636

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -11,6 +11,7 @@
       "backLink": "Yn ôl",
       "errorSummaryTitle": "Mae problem",
       "error": "Gwall",
+      "errorTitlePrefix": "Gwall: ",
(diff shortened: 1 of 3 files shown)


Joe-Edwards-GDS (approved):
Looks sensible to me

PR #1394, opened 2024-04-15
PYIC-5637: set attributes so that focus is on an error if it occurs

sam803:
## Proposed changes
 Adds attributes to `govukErrorSummary` in `base.njk`
### What changed

Added `attributes: { autoFocus: true, tabindex: 0 }` so that when an error occurs the error summary is in focus

Screenshot 2024-04-15 at 17 11 06

### Why did it change
Currently the focus is on `skip to main content `link at the top of the page, but should be on the error as per guidance in https://design-system.service.gov.uk/components/error-summary/#options-error-summary-second-example--error-list

### Issue tracking
- https://govukverify.atlassian.net/browse/PYIC-5637

### DIFF
--- a/src/views/shared/base.njk
+++ b/src/views/shared/base.njk
@@ -58,7 +58,7 @@
             activeLanguage: htmlLang,
             class: "",
             languages: [
-            { 
+            {
               code: 'en',
               text: 'English',
               visuallyHidden: 'Change to English'
@@ -83,6 +83,7 @@
                    {% if errorState %}
                         {{ govukErrorSummary({
                             titleText: errorTitle | default('Error Summary'),
+                            attributes: { autoFocus: true, tabindex: 0 },
                             errorList: [
                                 {
                                     text: errorText | translate,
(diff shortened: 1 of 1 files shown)


src/views/shared/base.njk
```diff
@@ -83,6 +83,7 @@
                    {% if errorState %}
                         {{ govukErrorSummary({
                             titleText: errorTitle | default('Error Summary'),
+                            attributes: { autoFocus: true, tabindex: 0 },
```
Joe-Edwards-GDS, line 86: It looks as though the `govukErrorSummary` should automatically focus the element for us - do you know why this isn't working? Code is https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/error-summary/error-summary.mjs but I would sort of expect this to 'just work'...

I wonder if we're not properly loading the GOV.UK javascript :/

sam803: see better fix for this https://github.com/govuk-one-login/ipv-core-front/pull/1398

PR #1395, opened 2024-04-16
PYIC-5669: New Screen: What does the user want to change

sammyiyke:
## Proposed changes
[PYIC-5669](https://govukverify.atlassian.net/browse/PYIC-5669): New Screen: What does the user want to change

Creation of a new page titled "Update your details" as part of the "Continuation of Identity" initiative for IPV core 

### What changed

- Adds new "Update your details" page/screen for confirming and selecting user detail changes as part of the "Continuation of Identity" initiative
- Adds support for new data format (`formattedName`) returned to user details in tests

### Why did it change

As part of the COI user journey we need to allow users choose what details they would like to change.

This feature is critical for maintaining accurate and up-to-date user information in accordance with GPG45 specifications.

### Issue tracking

- [PYIC-5669](https://govukverify.atlassian.net/browse/PYIC-5669)

Screenshot 2024-04-16 at 11 11 38

Screenshot 2024-04-16 at 11 12 00

[PYIC-5669]: https://govukverify.atlassian.net/browse/PYIC-5669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-5669]: https://govukverify.atlassian.net/browse/PYIC-5669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/shared/reuseHelper.js
```diff
@@ -52,6 +56,7 @@ module.exports = {
   generateUserDetails: (userDetailsResponse, i18n) => {
     return {
       name: userDetailsResponse.data?.name,
+      formattedName: userDetailsResponse.data?.formattedName,
```
RogVanGDS: Wish we had typescript to make this a bit more obvious what was going on 😢 

Could we rename "name" to "fullName" at least to help distinguish between the two?

src/locales/en/translation.json
```diff
@@ -407,6 +407,33 @@
...
+        "summaryListHeading1": "Given names",
+        "summaryListHeading2": "Last name",
+        "summaryListHeading3": "Current home address",
+        "summaryListHeading4": "Date of birth",
+        "checkboxes": {
+          "heading": "Which details do you need to update?",
+          "hint": "Select all that apply.",
+          "errorMessage": "Select at least one option",
+          "checkbox1": "Given names",
+          "checkbox1Hint": "This is your first name and any middle names",
```
RogVanGDS: Not sure we need the `"any"` looking at the designs

src/views/ipv/page/pyi-update-details.njk
```diff
@@ -0,0 +1,109 @@
...
+    }) }}
+
+    <form id="confirmAddressActionForm" action="/ipv/page/{{ pageId }}" method="POST">
+            <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+            {% set checkboxConfig = {
+              name: "nationality",
+              fieldset: {
+                legend: {
+                  text: 'pages.pyiUpdateDetails.content.checkboxes.heading' | translate,
+                  isPageHeading: true,
```
RogVanGDS: I think we don't want `isPageHeading: true`, the design docs suggest that it should be set to true if the legend also acts as the heading for the page.

I would assume that this isn't the case given we have an h1 header above it in the page?

RogVanGDS: (We should only be using this if we want to avoid the screenreader from repeating the same heading twice, if the heading of the page is the same as the radio button header)

src/views/ipv/page/pyi-update-details.njk
```diff
@@ -0,0 +1,109 @@
...
+                },
+                {
+                  value: "dateOfBirth",
+                  text: 'pages.pyiUpdateDetails.content.checkboxes.checkbox4' | translate
+                },
+                {
+                  divider: 'pages.pyiUpdateDetails.content.checkboxes.checkbox5Divisor' | translate
+                },
+                {
+                  value: "none",
```
RogVanGDS: Slightly nit-picky I know, but could we change the value to be something like `back` / `cancel` instead of `none`? 

Personally `none` suggests that we are intentionally leaving it empty, instead of it being a valid journey event

src/app/shared/reuseHelper.js
sammyiyke: My reasoning here is to leave as much as possible unchanged since a new field is being added. I did rename it to fullName earlier but change it cause I felt the `name` keeps some consistency with what is sent from core-back too.

Just my thoughts.. But I am happy to be wrong..

src/views/ipv/page/pyi-update-details.njk
sammyiyke: Oh, didn't know that. Good to know 👍

src/views/ipv/page/pyi-update-details.njk
```diff
@@ -0,0 +1,122 @@
...
+        {% set givenNames = (givenNames.push(namePart.value), givenNames) %}
+      {% endif %}
+    {% endfor %}
+
+    {% set familyNames = [] %}
+    {% for namePart in userDetails.nameParts %}
+      {% if namePart.type == "FamilyName" %}
+        {% set familyNames = (familyNames.push(namePart.value), familyNames) %}
+      {% endif %}
+    {% endfor %}
```
Joe-Edwards-GDS: I'd probably slightly prefer this logic to be done in `reuseHelper.js` - it'll be easier than here, and it'll be reusable if we need to display the given/family name on another page.

Not critical - we could wait until we know we need to reuse it.

src/views/ipv/page/pyi-update-details.njk
Joe-Edwards-GDS, line 1: Can we update this path to be simply `update-details` - I think going forward we're not using the `pyi` prefix for URLs.

Joe-Edwards-GDS (commented):
Thanks @sammyiyke , I think this is looking good. Also needs a merge/rebase with main

src/views/ipv/page/pyi-update-details.njk
sammyiyke: Makes sense. Didn't think of it.

src/app/ipv/middleware.test.js
```diff
@@ -515,6 +519,10 @@ describe("journey middleware", () => {
 
       const expectedUserDetail = {
         name: "firstName LastName",
+        nameParts: {
+          givenName: "firstName",
```
Joe-Edwards-GDS, line 523: Do you think this should test the case where you have multiple given names?

src/app/shared/reuseHelper.js
```diff
@@ -50,8 +54,23 @@ module.exports = {
...
+    let givenName = "";
+    let familyName = "";
+    if (userDetailsResponse.data?.nameParts) {
+      userDetailsResponse.data.nameParts.forEach((namePart) => {
+        if (namePart.type === "GivenName") {
+          givenName = namePart.value;
+        } else if (namePart.type === "FamilyName") {
+          familyName = namePart.value;
+        }
+      });
```
Joe-Edwards-GDS: I don't think this is right - what if you have multiple given names?

What about something like:
```
const givenName = userDetailsResponse.data?.nameParts
  .filter((namePart) => namePart.type === "GivenName")
  .join(" ");
```

src/views/ipv/page/update-details.njk
```diff
@@ -14,35 +14,21 @@
...
-    {% endfor %}
-
     {% set rows = [
       {
         key: {
           text: 'pages.pyiUpdateDetails.content.summaryListHeading1' | translate
         },
         value: {
-          text: givenNames.join(' ')
+          text: userDetails.nameParts.givenName | safe
```
Joe-Edwards-GDS: Why are we using `safe` here? I think we do want to escape this

src/app/ipv/middleware.test.js
sammyiyke: I think it might be better to add that test case as a new test

src/app/shared/reuseHelper.js
sammyiyke: Went with a slightly different approach, idea is to try reducing the operations. Instead of using two separate `filter`/`find` operations

Joe-Edwards-GDS (approved):

PR #1396, opened 2024-04-16
PYIC-2681 Use Express built-in IP logic

Joe-Edwards-GDS:
### What changed

IP Address passed to core-back comes from the `req.ip` property from express: https://expressjs.com/en/api.html#req.ip, This uses the left-most value of the X-Forwarded-For header automatically.

We also use a fresh value for the IP on every request, rather than caching in the session.

### Why did it change

We previously used our own custom IP logic, and set the ip address on the session only when the user first visits the site (i.e. on the `/authorization` call).

This uses built-in functionality and ensures we have an up-to-date IP Address for each request.

N.B. we previously handled the `Forwarded` header as well as `X-Forwarded-For`, but as we currently use REST API Gateway, just `X-Forwarded-For` should be sufficient.

We will need to test this when introducing CloudFront, to make sure the headers chain together correctly.

### Issue tracking

- [PYIC-2681](https://govukverify.atlassian.net/browse/PYIC-2681)

[PYIC-2681]: https://govukverify.atlassian.net/browse/PYIC-2681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TobySaundersGDS (dismissed):
This should be good - the only 'issue' is that x-forwarded-for can be spoofed, but you are already in that situation so as a simplification of the code this looks good to me. This should also be fine once cloudfront is in place as the x-forwarded-for header is appended to by cloudfront, meaning you get the correct entry for the client with this approach.

Wynndow (dismissed):

Wynndow (approved):

PR #1397, opened 2024-04-16
PYIC-5741: Fixed back button wrong error screen

RogVanGDS:
### What changed

Fixed back button wrong error screen

### Why did it change

We weren't checking whether the user had gone back (we were trying to render an unexpected page) as part of the formRadioCheck validation for checking empty radio buttons.

This meant that if the user triggered the error screen on the previous page and then tried to go back, we would render a different page (even one that didn't exist such as `/ipv/dcmaw`, as the formRadioCheck would just try to render the current page if the journey body was empty, bypassing any unexpected page behaviour)

Now we are validating the expected page BEFORE running the formRadioCheck validation, to catch this error case and correctly route the user to the recoverable page.

### Issue tracking

- [PYIC-5741](https://govukverify.atlassian.net/browse/PYIC-5741)

[PYIC-5741]: https://govukverify.atlassian.net/browse/PYIC-5741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -232,15 +232,31 @@ function handleAppStoreRedirect(req, res, next) {
   }
 }
 
+async function handleUnexpectedPage(req, res, pageId) {
+  logError(
+    req,
+    {
+      pageId: pageId,
+      expectedPage: req.session.currentPage,
+    },
+    "page :pageId doesn't match expected session page :expectedPage",
+  );
+
+  req.session.currentPage = PAGES.PYI_ATTEMPT_RECOVERY;
+
+  return await saveSessionAndRedirect(
+    req,
+    res,
+    getIpvPagePath(PAGES.PYI_ATTEMPT_RECOVERY),
+  );
+}
+
 module.exports = {
   renderAttemptRecoveryPage: async (req, res) => {
     res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
       csrfToken: req.csrfToken(),
     });
   },
-  // This method is currently only used by a link on the pyi-f2f-delete-details page.
-  // It shouldn't be used for anything else, see
-  // https://govukverify.atlassian.net/browse/PYIC-4859.
   updateJourneyState: async (req, res, next) => {
     try {
       const currentPageId = req.params.pageId;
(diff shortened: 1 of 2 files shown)


src/app/ipv/middleware.js
```diff
@@ -227,15 +227,30 @@ function handleAppStoreRedirect(req, res, next) {
...
 module.exports = {
   renderAttemptRecoveryPage: async (req, res) => {
     res.render("ipv/page/pyi-attempt-recovery.njk", {
       csrfToken: req.csrfToken(),
     });
   },
-  // This method is currently only used by a link on the pyi-f2f-delete-details page.
-  // It shouldn't be used for anything else, see
-  // https://govukverify.atlassian.net/browse/PYIC-4859.
   updateJourneyState: async (req, res, next) => {
```
RogVanGDS, line 260: I removed this comment as it is no longer true, we are now using this as a handler for anchor tag driven journey events, and relying on backend validation to ensure that it isn't abused.

src/app/ipv/middleware.js
```diff
@@ -394,18 +395,26 @@ module.exports = {
...
+          pageErrorState: true,
+          context,
+        };
+
+        if (pageRequiresUserDetails(pageId)) {
+          renderOptions.userDetails = await fetchUserDetails(req);
+        }
+
+        req.renderOptions = renderOptions;
+
```
RogVanGDS: Not sure it made sense to make a call to the backend to fetch user details and input the correct renderOptions before we have decided whether the radio button is empty, but please shout if I've missed something

Joe-Edwards-GDS: Yeah, I think this makes sense.

Do we also need to create our own render method which fetches user details when necessary? I think it should basically happen any time we render a page...

RogVanGDS: Hmm we could definitely add our own render function which deals with fetching user details if we need to, but there aren't that many pages where we require user details as it currently stands

Joe-Edwards-GDS: Sure - I don't feel super strongly, I'd just like to avoid more cases like this :)

Joe-Edwards-GDS (dismissed):
Happy to approve if you don't think we need a common render function yet.

Joe-Edwards-GDS (approved):

PR #1398, opened 2024-04-16
PYIC-5637: call initAll for govuk javascript initialization

sam803:
## Proposed changes

Call initAll() function to initialise govuk javascript

### What changed

Adds a call to `window.GOVUKFrontend.initAll()` function, so that the default focus behaviour of the ErrorSummary gets set correctly. Followed guidance listed [here](https://frontend.design-system.service.gov.uk/v4/importing-css-assets-and-javascript/#add-the-javascript-file-to-your-html)  

Screenshot 2024-04-15 at 17 11 06

### Why did it change
Currently the focus is on `skip to main content `link at the top of the page, but should be on the error as per guidance id [documentation](https://design-system.service.gov.uk/components/error-summary/#options-error-summary-second-example--error-list)

### Issue tracking
- https://govukverify.atlassian.net/browse/PYIC-5637

### DIFF
--- a/src/views/shared/base.njk
+++ b/src/views/shared/base.njk
@@ -58,7 +58,7 @@
             activeLanguage: htmlLang,
             class: "",
             languages: [
-            { 
+            {
               code: 'en',
               text: 'English',
               visuallyHidden: 'Change to English'
@@ -138,6 +138,7 @@
     {% block scripts %}{% endblock %}
     <script nonce='{{ cspNonce }}' src="{{ assetsCdnPath }}/public/javascripts/application.js"></script>
     <script nonce='{{ cspNonce }}'>
+        window.GOVUKFrontend.initAll()
         window.DI = window.DI || {};
         window.DI.httpStatusCode = {{ statusCode | d(200) }};
         window.DI.journeyState = "{{ googleTagManagerPageId }}";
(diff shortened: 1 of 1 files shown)


Joe-Edwards-GDS (approved):
This looks like the right approach, thanks!

Tricky question, but do you know if there are any other components that this is likely to affect? i.e. that we might need to test before merging.

sam803:
> This looks like the right approach, thanks!
> 
> Tricky question, but do you know if there are any other components that this is likely to affect? i.e. that we might need to test before merging.

@Joe-Edwards-GDS - Just looking over the initAll function it initializes the following:
 - Accordion
 - Button
 - CharacterCount
 - Checkboxes
 - Details
 - ErrorSummary
 - ExitThisPage
 - Header

I can take a look and see which we are using and check their behaviour, or we can just init the ErrorSummary if worried e.g. https://frontend.design-system.service.gov.uk/v4/importing-css-assets-and-javascript/#select-and-initialise-an-individual-component ?

sam803:
> * ExitThisPage

Further to this we are only using ErrorSummary, Button and Details - I can't see any difference in behaviour in my local setup

Joe-Edwards-GDS:
Thanks for checking - I think we should definitely `initAll`, to avoid similar issues in future, just want to make sure we aren't going to be caught out by new behaviour.

PR #1407, opened 2024-04-25
PYIC-5971: fix design issue m2b journey page

shivanshuit914:
## Proposed changes

- Add `GOV.UK One Login` in all page title.
- Remove unwanted margin from h2, UL tag, m2b no photo id journey.
- Add missing `.` in welsh translation m2b no photo id journey.

### What changed

- Added default title suffix `GOV.UK One Login`
- Removed unwanted html tags, css classes from m2b no photo id journey.

### Why did it change

- Recommended changes by UCD after review.

### Issue tracking

- [PYIC-5971](https://govukverify.atlassian.net/browse/PYIC-5971)

Screenshot 2024-04-25 at 15 51 05

[PYIC-5971]: https://govukverify.atlassian.net/browse/PYIC-5971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/ipv/page/page-ipv-bank-account-start.njk
+++ b/src/views/ipv/page/page-ipv-bank-account-start.njk
@@ -13,19 +13,15 @@
   <h1 class="govuk-heading-l" id="header" data-page="{{ googleTagManagerPageId }}">{{ 'pages.pageIpvBankAccountStart.header' | translate }}</h1>
   <p class="govuk-body">{{ 'pages.pageIpvBankAccountStart.content.paragraph1' | translate }}</p>
 
-  <ul class="govuk-list" role="list">
-    <li>
-      <h2 class="govuk-heading-m govuk-!-margin-top-7">{{ 'pages.pageIpvBankAccountStart.content.subHeading' | translate }}</h2>
-      <p class="govuk-body govuk-!-margin-bottom-2">{{ 'pages.pageIpvBankAccountStart.content.paragraph2' | translate }}</p>
-      <ul class="govuk-list govuk-list--bullet ">
-        {% for listItem in 'pages.pageIpvBankAccountStart.content.personalDetails' | translate({ returnObjects: true }) %}
-          <li>{{ listItem }}</li>
-        {% endfor %}
-      </ul>
-    </li>
+  <h2 class="govuk-heading-m">{{ 'pages.pageIpvBankAccountStart.content.subHeading' | translate }}</h2>
+  <p class="govuk-body">{{ 'pages.pageIpvBankAccountStart.content.paragraph2' | translate }}</p>
+  <ul class="govuk-list govuk-list--bullet ">
+    {% for listItem in 'pages.pageIpvBankAccountStart.content.personalDetails' | translate({ returnObjects: true }) %}
+      <li>{{ listItem }}</li>
+    {% endfor %}
   </ul>
 
-  <h2 class="govuk-heading-m govuk-!-margin-top-7">{{ 'pages.pageIpvBankAccountStart.content.subHeading2' | translate }}</h2>
+  <h2 class="govuk-heading-m">{{ 'pages.pageIpvBankAccountStart.content.subHeading2' | translate }}</h2>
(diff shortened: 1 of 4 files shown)


src/views/ipv/page/page-ipv-bank-account-start.njk
```diff
@@ -13,19 +13,15 @@
...
-      <h2 class="govuk-heading-m govuk-!-margin-top-7">{{ 'pages.pageIpvBankAccountStart.content.subHeading' | translate }}</h2>
-      <p class="govuk-body govuk-!-margin-bottom-2">{{ 'pages.pageIpvBankAccountStart.content.paragraph2' | translate }}</p>
-      <ul class="govuk-list govuk-list--bullet ">
-        {% for listItem in 'pages.pageIpvBankAccountStart.content.personalDetails' | translate({ returnObjects: true }) %}
-          <li>{{ listItem }}</li>
-        {% endfor %}
-      </ul>
-    </li>
+  <h2 class="govuk-heading-m">{{ 'pages.pageIpvBankAccountStart.content.subHeading' | translate }}</h2>
+  <p class="govuk-body govuk-!-margin-bottom-2">{{ 'pages.pageIpvBankAccountStart.content.paragraph2' | translate }}</p>
```
blakeyp: Is this margin override necessary either? 
Also would you mind adding a screenshot of the page to the PR description so we can see how it looks now 🙂

src/views/shared/base.njk
```diff
@@ -31,7 +31,7 @@
     {%- if pageErrorState %}
         {{ 'general.govuk.errorTitlePrefix' | translate }}
     {% endif %}
-    {%- if pageTitleKey %}{{ pageTitleKey | translateWithContextOrFallback(context) }} – GOV.UK{% endif %}
+    {%- if pageTitleKey %}{{ pageTitleKey | translateWithContextOrFallback(context) }} – GOV.UK One Login{% endif %}
```
DanCorderIPV, line 34: This ensures that all page titles end with `GOV.UK One Login` but it doesn't ensure that the title and the H1 are always the same. Do we need to do that too?

DanCorderIPV (dismissed):
The changes here look good, not sure if they fully address the ticket though.

src/views/shared/base.njk
shivanshuit914: I think we need only for title @blakeyp  Can you please confirm :)

blakeyp: Yep it's only needed for page titles not the headings

src/views/ipv/page/page-ipv-bank-account-start.njk
shivanshuit914: No, looks good without margin override. I will remove it. 

Yes Uploaded it :)

src/views/shared/base.njk
DanCorderIPV: Sorry, that's not what I meant - the ticket says that page titles should be in the form `H1 - GOV.UK One Login` this change only addresses the second half of that - we have separate translation file entries for title and heading on most (all?) pages.
The text maybe the same, but if we're deciding that they should always be the same, should we delete the title entries from the translation files?

blakeyp: Ah I see, I think they are almost always the same but I've just checked for the M2B screens and there is a discrepancy for `pyi-no-match` - the title is "Sorry, we could not confirm your details" while the header is "Sorry, we cannot confirm your identity". We should check this against the original designs/content doc

(There is probably a separate issue across the board here about whether we really need separate entries or could refactor to always use the heading in the title)

blakeyp: It looks the title was unintentionally changed from "Sorry, we cannot confirm your identity" as part of the address-fraud check work - so we should correct this @shivanshuit914

blakeyp: (the Welsh is correct as is)

blakeyp (approved):

PR #1413, opened 2024-04-26
PYIC-5560: Add iphone/ android device sniffing

MikeCollingwood:
### What changed

I spent an hr looking into this, to abstract logic and add new sniffing logic

### Issue tracking

- [PYIC-5560](https://govukverify.atlassian.net/browse/PYIC-5560)

### Testing

- Manual testing:
  - iPhone journey works for MAM and DAD!
- E2E testing for core-back and front changes
  - This passed: 
Screenshot 2024-04-29 at 14 11 07

- This PR should be merged after the core-back PR to add more events into the journey map

[PYIC-5560]: https://govukverify.atlassian.net/browse/PYIC-5560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/shared/deviceSniffingHelper.js
+++ b/src/app/shared/deviceSniffingHelper.js
@@ -1,0 +1,37 @@
+const UAParser = require("ua-parser-js");
+const PHONE_TYPES = require("../../constants/phone-types");
+const OS_TYPES = require("../../constants/os-types");
+
+function modifyJourneyOnSniffing(req) {
+  const parser = new UAParser(req.headers["user-agent"]);
+
+  if (parser.getDevice()["type"] === "mobile") {
+    req.body.journey += "Smartphone";
+    switch (parser.getOS()["name"]) {
+      case OS_TYPES.IOS:
+        req.body.journey += "Iphone";
+        break;
+      case OS_TYPES.ANDROID:
+        req.body.journey += "Android";
+        break;
+    }
+  }
+}
+
+function preferenceSniffedPhoneType(req) {
+  const parser = new UAParser(req.headers["user-agent"]);
+
+  switch (parser.getOS()["name"]) {
+    case OS_TYPES.IOS:
+      return PHONE_TYPES.IPHONE;
+    case OS_TYPES.ANDROID:
+      return PHONE_TYPES.ANDROID;
+    default:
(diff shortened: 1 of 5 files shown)


src/app/shared/deviceSniffingHelper.js
```diff
@@ -0,0 +1,37 @@
...
+        req.body.journey += "Iphone";
+        break;
+      case OS_TYPES.ANDROID:
+        req.body.journey += "Android";
+        break;
+    }
+  }
+}
+
+function preferenceSniffedPhoneType(req) {
```
DanCorderIPV: Happy to discuss this, but I think this method would be easier to understand if it was called something like `sniffPhoneType` and took `req.params.specifiedPhoneType` as a parameter called `fallback` or similar.

src/app/shared/deviceSniffingHelper.js
```diff
@@ -0,0 +1,37 @@
...
+  const parser = new UAParser(req.headers["user-agent"]);
+
+  if (parser.getDevice()["type"] === "mobile") {
+    req.body.journey += "Smartphone";
+    switch (parser.getOS()["name"]) {
+      case OS_TYPES.IOS:
+        req.body.journey += "Iphone";
+        break;
+      case OS_TYPES.ANDROID:
+        req.body.journey += "Android";
```
DanCorderIPV: I'm a bit uncomfortable with this method modifying `req.body.journey` rather than returning a value. It feels like it's knowing too much about the request object for a device sniffing helper.

DanCorderIPV (changes requested):
Happy to discuss these comments as they are a bit picky :)

src/app/shared/deviceSniffingHelper.js
MikeCollingwood: I like your suggestion! Will go with it

src/app/shared/deviceSniffingHelper.js
MikeCollingwood: That's fair, I'll update it to `getJourneyOnSniffing` where it returns a string

DanCorderIPV (approved):

PR #1414, opened 2024-04-26
PYIC-6125: Added new COI success and no match pages

RogVanGDS:
### What changed

Added new COI success and no match pages

### Why did it change

To show new content changes required for the Continuity of Identity journey

### Issue tracking

- [PYIC-6125](https://govukverify.atlassian.net/browse/PYIC-6125)

Screenshot 2024-04-26 at 11 39 12
Screenshot 2024-04-26 at 11 39 28
Screenshot 2024-04-26 at 11 39 51
Screenshot 2024-04-26 at 11 40 22

[PYIC-6125]: https://govukverify.atlassian.net/browse/PYIC-6125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sam803 (dismissed):
LGTM

src/views/ipv/page/coi-no-match.njk
Joe-Edwards-GDS, line 1: I think UCD are paying more attention to the URLs of these pages now...

When we discussed with UCD, we expected these to be variants of `pyi-no-match` and `page-dcmaw-success`, but if it's easier to create entirely new pages, I expect they won't mind too much, but then we should confirm what the URLs should be.

RogVanGDS: Yeah that's a good point actually, I'll ask UCD if we can have alternatives. We can re-use the existing, but it's probably worth having a separate component now we have so many versions and just conditionally rendering the component in a base-screen

sam803 (approved):

PR #1415, opened 2024-04-26
PYIC-6156: Fix session validation

MikeCollingwood:


src/app/ipv/middleware.js
```diff
@@ -358,27 +375,23 @@ module.exports = {
...
-      PAGES.PYI_SUGGEST_OTHER_OPTIONS,
-      PAGES.PYI_CRI_ESCAPE,
-      PAGES.PYI_KBV_ESCAPE_M2B,
-      PAGES.PYI_ESCAPE_M2B,
-      PAGES.PAGE_MULTIPLE_DOC_CHECK,
-    ];
 
     try {
-      if (pagesUsingSessionId.includes(currentPageId)) {
-        checkForSessionId(req, res);
```
MikeCollingwood, line 371: These wouldn't block further handling logic (which re-writes res.render)

src/exceptions/MissingIpvAndOauthSessionIdsError.js
```diff
@@ -0,0 +1,10 @@
+function MissingIpvAndOauthSessionIdsError() {
+  return true;
```
MikeCollingwood, line 2: Sonarcloud would complain if it were empty

src/exceptions/MissingIpvAndOauthSessionIdsError.js
Joe-Edwards-GDS, line 1: I would definitely try and avoid messing about with prototypes in JS unless you absolutely have to! Creating custom errors in JS is also tricksy, to get the behaviour you expect.

If you want your own error type then declare it as a class similar to this:
```
class MyError extends Error {
  constructor() {
    super("Some message"); // Sets the message
    this.name = this.constructor.name; // Sets the name
    Error.captureStackTrace(this, this.constructor); // Generates the stack trace
  }
}
```

PR #1418, opened 2024-04-30
PYIC-5809: added generation of update details journey

sam803:
## Proposed changes

Add middleware to generate the journey based on the checkboxes selected when choosing to update details

### What changed

Adds specific middleware for the update-details page

### Why did it change

As part of continuity of identity, a user is able to select what they want to update (name, dob, address). The combination items that they select will determine the journey they need to take

### Issue tracking

- [PYIC-5809](https://govukverify.atlassian.net/browse/PYIC-5809)

[PYIC-5809]: https://govukverify.atlassian.net/browse/PYIC-5809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -31,6 +31,7 @@ const { saveSessionAndRedirect } = require("../shared/redirectHelper");
 const coreBackService = require("../../services/coreBackService");
 const qrCodeHelper = require("../shared/qrCodeHelper");
 const PHONE_TYPES = require("../../constants/phone-types");
+const UPDATE_DETAILS_JOURNEY_TYPES = require("../../constants/update-details-journeys");
 const appDownloadHelper = require("../shared/appDownloadHelper");
 const {
   getIpvPageTemplatePath,
@@ -207,6 +208,47 @@ function checkJourneyAction(req) {
     logError(req, err);
 
     throw new Error("req.body?.journey is missing");
+  }
+}
+
+// getCoiUpdateDetailsJourney determines the next journey based on the detailsToUpdate
+// field of the update-details page
+function getCoiUpdateDetailsJourney(detailsToUpdate) {
+  if (!detailsToUpdate) {
+    return;
+  }
+  // convert to array if its a string
+  if (typeof detailsToUpdate === "string") {
+    detailsToUpdate = [detailsToUpdate];
+  }
+
+  if (detailsToUpdate.includes("cancel")) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_CANCEL;
+  }
+  // send to update-names-dob journey if dateOfBirth selected
+  if (detailsToUpdate.includes("dateOfBirth")) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAMES_DOB;
+  }
+
+  const hasAddress = detailsToUpdate.includes("address");
+  const hasGivenNames = detailsToUpdate.includes("givenNames");
+  const hasLastName = detailsToUpdate.includes("lastName");
+
+  // send to update-names-dob journey if both names selected
+  if (hasGivenNames && hasLastName) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAMES_DOB;
+  }
+  if (hasGivenNames || hasLastName) {
+    // send to update-name-address journey if only one name and address selected
+    if (hasAddress) {
+      return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAME_ADDRESS;
+    }
+    // send to update-name journey if only one name selected and no address selected
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAME;
+  }
+  // send to address journey if just the address
+  if (hasAddress) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_ADDRESS;
   }
 }
 
@@ -424,6 +466,14 @@ module.exports = {
         }
         next();
       }
+    } catch (error) {
+      next(error);
+    }
+  },
+  formHandleUpdateDetailsCheckBox: async (req, res, next) => {
+    try {
+      req.body.journey = getCoiUpdateDetailsJourney(req.body.detailsToUpdate);
+      next();
     } catch (error) {
       next(error);
     }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -76,6 +76,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
@@ -160,9 +164,9 @@
         "filename": "src/app/ipv/middleware.test.js",
         "hashed_secret": "ec420572bb867a4f38076e71a4ac3b712326e496",
         "is_verified": false,
-        "line_number": 278
+        "line_number": 279
       }
     ]
   },
-  "generated_at": "2024-04-24T08:44:04Z"
+  "generated_at": "2024-04-30T13:58:12Z"
 }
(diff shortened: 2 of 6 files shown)


src/app/ipv/middleware.js
```diff
@@ -210,6 +211,40 @@ function checkJourneyAction(req) {
...
+      return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_CANCEL;
+    }
+    // send to update-names-dob journey if dateOfBirth or both names selected
+    if (
+      detailsToUpdate.includes("dateOfBirth") ||
+      (detailsToUpdate.includes("givenNames") &&
+        detailsToUpdate.includes("lastName"))
+    ) {
+      return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAMES_DOB;
+    }
```
RogVanGDS: What happens if the user selects given name and date of birth for example? Is that a valid selection?

src/app/ipv/middleware.js
```diff
@@ -210,6 +211,40 @@ function checkJourneyAction(req) {
...
+    // send to update-name-address journey if only one name selected
+    if (
+      detailsToUpdate.includes("givenNames") ||
+      detailsToUpdate.includes("lastName")
+    ) {
+      return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAME_ADDRESS;
+    }
```
RogVanGDS: Possibly I've gotten confused, but do we not need to know whether the user has also selected Address here? The Name Change sub journey has two separate journeys within it, one for Name + Address and the other for just name

src/app/ipv/middleware.js
sam803: @RogVanGDS According to the diagram in https://govukverify.atlassian.net/browse/PYIC-5809 if DOB is selected with anything, it should  go to UPDATE_NAMES_DOB

RogVanGDS: Ah ok, makes sense thanks 👍

src/app/ipv/middleware.js
```diff
@@ -428,6 +463,38 @@ module.exports = {
...
+        }
+        req.renderOptions = renderOptions;
+        res.render(getIpvPageTemplatePath(sanitize(pageId)), renderOptions);
+      } else {
+        next();
+      }
+    } catch (error) {
+      next(error);
+    }
+  },
```
RogVanGDS: Can we avoid duplicating the unexpected page logic by using the existing `formRadioButtonChecked` middleware function, and then having a separate middleware method that runs afterwards, which can then safely call `getCoiUpdateDetailsJourney()`?

src/app/ipv/middleware.js
```diff
@@ -210,6 +211,40 @@ function checkJourneyAction(req) {
...
 
+// getCoiUpdateDetailsJourney determines the next journey based on the detailsToUpdate
+// field of the update-details page
+function getCoiUpdateDetailsJourney(detailsToUpdate) {
+  if (detailsToUpdate) {
+    // convert to array if its a string
+    if (typeof detailsToUpdate === "string") {
+      detailsToUpdate = [detailsToUpdate];
+    }
+    // cancel event
```
RogVanGDS: Sorry to nit pick, but I think the if cases for some of these comments might be clear enough already (and I think we frequently fail to keep them up to date anyway) 🙈

src/app/ipv/middleware.js
sam803: @RogVanGDS 
So according to the ticket:
- DOB with anything OR both names selected (plus anything else I guess) goes to UPDATE_NAME_DOB
- Either name on its own OR with an address goes to UPDATE_NAME_ADDRESS
- Address on its own goes to UPDATE_ADDRESS

Theres not much point checking for the address in this case  as its going to the same journey if only one of the names is selected

src/app/ipv/middleware.js
sam803: @RogVanGDS sure yeh I can take a look at refactoring :)

src/app/ipv/middleware.js
RogVanGDS: Hm ok I think the ticket might actually be wrong, or something has fallen in the cracks between your ticket and [mine](https://govukverify.atlassian.net/browse/PYIC-5829).

If the user needs to change Address + Name, we delete all their VCs take them down to DCMAW CRI ->  ADDRESS CRI -> FRAUD CRI

If the user needs to just change Name (without address), we delete all their VCs **except** the Address VC, and we take them down DCMAW CRI -> FRAUD CRI.

So we will need a separate route from the front-end so the backend can differentiate between the two.

@Joe-Edwards-GDS Not sure if there is a separate ticket for this or if it's just been missed?

sam803: @RogVanGDS so my understanding is that once the user selects to change a name and/or address in this screen, then they get dropped into the sub-journey referenced in your ticket which takes them through either the name only or name+address journeys. Maybe I'm missing something though. Should we add a separate events for name only and name+address which route to the same sub journey?

Joe-Edwards-GDS: I think @RogVanGDS is right - we'll want separate events for name vs name + address. I expect that they'll both go to the same sub-journey, but will go to different entry-points.

Sorry, I hadn't originally realised that we needed to skip address if the user doesn't elect to change it. If it causes a lot of difficulty we could try and push back on the UCD, but it feels doable.

(More broadly, this is starting to stretch the capability of what our journey map can do - there's a piece of work in the roadmap to evaluate what the evolution of IPV Core journey routing looks like, but I don't want to half-bake it)

sam803: Thanks for the comments @Joe-Edwards-GDS @RogVanGDS I've added a separate event for name only, please see latest commit :)

src/app/ipv/middleware.js
```diff
@@ -210,6 +211,47 @@ function checkJourneyAction(req) {
...
+  if (typeof detailsToUpdate === "string") {
+    detailsToUpdate = [detailsToUpdate];
+  }
+
+  if (detailsToUpdate.includes("cancel")) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_CANCEL;
+  }
+  // send to update-names-dob journey if dateOfBirth selected
+  if (detailsToUpdate.includes("dateOfBirth")) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAMES_DOB;
```
RogVanGDS, line 230: I think this actually wants to be a `return saveSessionAndRedirect()` call, we want to send the user to the contact us webpage, which we usually handle on the front-end and not bother sending it to the backend.

Think it's fine to handle this in the `handleJourneyEvent()` method and not here directly, but we definitely want to avoid sending this to the backend I would think

src/app/ipv/middleware.js
```diff
@@ -210,6 +211,47 @@ function checkJourneyAction(req) {
...
+  }
+
+  const hasAddress = detailsToUpdate.includes("address");
+  const hasGivenNames = detailsToUpdate.includes("givenNames");
+  const hasLastName = detailsToUpdate.includes("lastName");
+
+  // send to update-names-dob journey if both names selected
+  if (hasGivenNames && hasLastName) {
+    return UPDATE_DETAILS_JOURNEY_TYPES.UPDATE_NAMES_DOB;
```
RogVanGDS, line 239: Same as above here

src/app/ipv/middleware.js
sam803: ah ok, great I was trying to figure out how to do that

sam803: @RogVanGDS according to https://govukverify.atlassian.net/browse/PYIC-5809 it goes to another form (update-name-date-birth) where the user is presented with another option to either go back or continue to the contact us page, so maybe better to let the backend handle this so it knows where to send the user if they select 'go back and check your details again'

See backend branch I'm running with if it makes more sense https://github.com/govuk-one-login/ipv-core-back/compare/PYIC-5809-update-journey-config?expand=1

Joe-Edwards-GDS: Yep, I think I agree @sam803

Joe-Edwards-GDS (approved):

PR #1420, opened 2024-05-01
BAU: Add govuk assets to copy-images command

MikeCollingwood:
### What changed

- the dist/public/images directory now looks like this:
Screenshot 2024-05-01 at 10 42 24
- which is due to now including govuk-frontend assets

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build-sass": "rm -rf dist/public/style.css && sass --no-source-map src/assets/scss/application.scss dist/public/stylesheets/application.css --style compressed",
     "watch-sass": "sass --no-source-map src/assets/scss/application.scss dist/public/stylesheets/application.css --style compressed --watch",
     "copy-locales": "mkdir -p dist && copyfiles -u 1 src/**/*.njk dist/ src/locales/**/** dist/ ",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "build-sass": "rm -rf dist/public/style.css && sass --no-source-map src/assets/scss/application.scss dist/public/stylesheets/application.css --style compressed",
     "watch-sass": "sass --no-source-map src/assets/scss/application.scss dist/public/stylesheets/application.css --style compressed --watch",
     "copy-locales": "mkdir -p dist && copyfiles -u 1 src/**/*.njk dist/ src/locales/**/** dist/ ",
-    "copy-images": "copyfiles -u 3 src/assets/images/** dist/public/images",
+    "copy-images": "copyfiles -u 3 src/assets/images/** dist/public/images && copyfiles -u 5 node_modules/govuk-frontend/govuk/assets/images/** dist/public/images",
```
MikeCollingwood, line 19: Should copy the other way round so our choice overrides govuk-frontend assets if there's an overlap

MikeCollingwood:
Closed because we are no longer using the CDN

PR #1423, opened 2024-05-02
PYIC-5246: Update content of multiple-doc-check page and pyi-post-office page and Welsh translations

thebauSoftwire:
### What changed
Updates for the `page-multiple-doc-check` page:
- update h2
- update radio button contents (inc adding a hint to 'other' option
Updates for `pyi-post-office`:
- updates to paragraphs 1 and 3
Welsh translations also updated accordingly.

See the following for [figma](https://www.figma.com/file/nglBHtbzJYEeST43Iu1jW3/Identity-Pod---UCD-Hive?type=design&node-id=18437-14581&mode=design&t=rHnlqPBDxpcUpaGS-4) designs and [Welsh translations](https://docs.google.com/document/d/1k6W66veKekAzl0cAQd6RdqDIjkvSocneM5_tPybb3aE/edit?pli=1)

**NOTE:** The designs include some text (`It takes about 10 minutes to prove your identity this way.`) which hasn't been marked as a change in the designs but doesn't exist on the existing `multiple-doc-check` page. The change for this is not included in this PR and will be done as a separate PR when the change has been approved.

### Issue tracking

- [PYIC-5246](https://govukverify.atlassian.net/browse/PYIC-5246)

[PYIC-5246]: https://govukverify.atlassian.net/browse/PYIC-5246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/ipv/page/page-multiple-doc-check.njk
+++ b/src/views/ipv/page/page-multiple-doc-check.njk
@@ -19,6 +19,7 @@
     {% endfor %}
   </ul>
   <p class="govuk-body">{{ 'pages.pageMultipleDocCheck.content.paragraph2' | translate | safe }}</p>
+  <p class="govuk-body">{{ 'pages.pageMultipleDocCheck.content.paragraph3' | translate | safe }}</p>
 
   <form id="multipleDocCheckingForm" action="/ipv/page/{{ pageId }}" method="POST">
     <input type="hidden" name="_csrf" value="{{ csrfToken }}">
--- a/src/views/ipv/page/page-multiple-doc-check.njk
+++ b/src/views/ipv/page/page-multiple-doc-check.njk
@@ -45,7 +45,8 @@
         },
         {
           value: "end",
-          text: 'pages.pageMultipleDocCheck.content.formRadioButtons.otherWayButtonText' | translateWithContext(context)
+          text: 'pages.pageMultipleDocCheck.content.formRadioButtons.otherWayButtonText' | translateWithContext(context),
+          hint: { text: 'pages.pageMultipleDocCheck.content.formRadioButtons.otherWayButtonTextHint' | translate }
         }
       ]
     } %}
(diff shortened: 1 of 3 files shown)


sam803 (dismissed):
lgtm :)

src/views/ipv/page/page-multiple-doc-check.njk
```diff
@@ -19,6 +19,7 @@
     {% endfor %}
   </ul>
   <p class="govuk-body">{{ 'pages.pageMultipleDocCheck.content.paragraph2' | translate | safe }}</p>
+  <p class="govuk-body">{{ 'pages.pageMultipleDocCheck.content.paragraph3' | translate | safe }}</p>
```
Wynndow: Do we need the `safe` filter here? Feels like we've not got any content that would get escaped?

sam803: In the welsh translation there is a ' char which might need it?

Wynndow: I think that's just an apostrophe? And actually looking at it closer I think it should be a `’` rather than a `'`

RogVanGDS: Yeah we don't need it for Welsh characters, it's only necessary if we are inputting html via the translation string

sam803: Good eyes, I'll remove the safe, the `'` is in a number of places in welsh translations (I copied it from another place) so might be worth another ticket to correct them?

Wynndow: Yep, happy to punt the apostrophe to another ticket

Wynndow (approved):

PR #1425, opened 2024-05-03
BAU: Revise inappropriate usages of path.join

Joe-Edwards-GDS:
`path.join` is used to construct file paths, not URL paths.

If used inappropriately, it may generate invalid paths on different operating systems (i.e. Windows), because it will use the wrong path separator (i.e. a backslash).

At the same time, address PYIC-4278

### DIFF
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -15,42 +15,33 @@ const {
   handleAppStoreRedirect,
 } = require("./middleware");
 
-// Remove this as part of PYIC-4278
-const { allTemplatesMoved } = require("../development/middleware");
-const { getRoutePath } = require("../../lib/paths");
-const path = require("path");
-const { UPDATE_DETAILS } = require("../../constants/ipv-pages");
+const {
+  PYI_ATTEMPT_RECOVERY,
+  UPDATE_DETAILS,
+} = require("../../constants/ipv-pages");
 
 const csrfProtection = csrf({});
 const parseForm = bodyParser.urlencoded({ extended: false });
 
-router.get(
-  path.join("/", "usefeatureset"),
-  validateFeatureSet,
-  renderFeatureSetPage,
-);
+function getPagePath(pageId) {
+  return `/page/${pageId}`;
+}
+
+router.get("/usefeatureset", validateFeatureSet, renderFeatureSetPage);
 
 router.get(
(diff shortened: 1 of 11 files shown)


src/app/ipv/router.js
```diff
@@ -15,42 +15,33 @@ const {
...
-);
+function getPagePath(pageId) {
+  return `/page/${pageId}`;
+}
+
+router.get("/usefeatureset", validateFeatureSet, renderFeatureSetPage);
 
 router.get(
-  path.join("/", "page", "attempt-recovery"),
+  getPagePath(PYI_ATTEMPT_RECOVERY),
```
Joe-Edwards-GDS, line 33: This is a change from `attempt-recovery` to `pyi-attempt-recovery`.

I can't see any evidence that we use `attempt-recovery` as a page URL (without the PYI prefix).

MikeCollingwood: The page url was `attempt-recovery` before-hand, so I wonder whether this routing is required?
```
router.get("/page/attempt-recovery", csrfProtection, renderAttemptRecoveryPage);
```

Joe-Edwards-GDS: I suspect you're right, although this might save an extra (pointless) redirect, if it's ever possible to be directed to `pyi-attempt-recovery` directly...

I'm tempted to leave it?

MikeCollingwood (approved):
This is a good fix to the misuse of path.join. The use of encodeURIComponent is also an improvement

src/lib/paths.js
```diff
@@ -1,24 +1,20 @@
...
 }
 
-function getIpvPageTemplatePath(page) {
-  return path.join("ipv", "page", addNunjucksExt(page));
-}
-
-function addNunjucksExt(path) {
-  return `${path}.njk`;
+function getTemplatePath(...pathComponents) {
+  const pageId = pathComponents.splice(-1, 1);
```
DanCorderIPV, line 12: I think pageId will be an array, does that work OK with the string interpolation below?

MikeCollingwood: It does work ok since an array `.toString()` returns the array values separated by commas - or just the value in the case of a single element in the array

Joe-Edwards-GDS: Good spot! Yeah it'll work although it's a bit more by luck than judgement 😬

PR #1427, opened 2024-05-07
SPT-712: Add header params to `RestDidCocumentMethod`

sam-leen-gds:
### What changed

- Set header.Content-Type: application/did+json
- Set header.Cache-Control: max-age=3600, private

### Why did it change

The [DID spec](https://www.w3.org/TR/did-core/#production) states that the media type should be set to application/did+json when publishing DID documents.

We, also, wish to use the cache control headers as a hint to RPs as to how often to retrieve the DID document.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -236,7 +236,7 @@ Resources:
   DevApiMapping:
     Type: AWS::ApiGatewayV2::ApiMapping
     Condition: IsDevelopment
-    DependsOn: RestApiGwDeployment202404121033
+    DependsOn: RestApiGwDeployment202405141531
     Properties:
       DomainName: !Ref DevApiDomain
       ApiId: !Ref RestApiGateway
@@ -997,7 +997,7 @@ Resources:
         Types:
           - REGIONAL
 
-  RestApiGwDeployment202404121033:
+  RestApiGwDeployment202405141531:
     DependsOn:
       - RestApiGatewayMethod
       - RestDidDocumentMethod
@@ -1074,8 +1074,10 @@ Resources:
       AuthorizationType: NONE
       MethodResponses:
         - StatusCode: "200"
+          ResponseParameters:
+              method.response.header.Cache-Control: true
           ResponseModels:
-            application/json: Empty
+            application/did+json: Empty
         - StatusCode: "404"
           ResponseModels: {}
       Integration:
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -1095,6 +1095,9 @@ Resources:
                 {
                     "message": "Not Found"
                 }
+        RequestParameters:
```
chris-dco: Should this not go in IntegrationResponses[0].ResponseParameters see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-method-integrationresponse.html#cfn-apigateway-method-integrationresponse-responseparameters. As RequestParameters looks like it goes to S3 https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-method-integration.html#cfn-apigateway-method-integration-requestparameters.

sam-leen-gds: Right you are, these have been moved.

chris-dco (approved):

deploy/template.yaml
```diff
@@ -1087,6 +1087,9 @@ Resources:
         IntegrationResponses:
           - StatusCode: "200"
             SelectionPattern: "200"
+            ResponseParameters:
+              integration.response.header.Content-Type: "application/did+json"
```
chris-dco: Think this should be `method.` rather than `integration.`
```suggestion
              method.response.header.Content-Type: "application/did+json"
```

deploy/template.yaml
```diff
@@ -1087,6 +1087,9 @@ Resources:
...
+              integration.response.header.Cache-Control: "max-age=3600, private"
```
chris-dco: Think this should be `method.` rather than `integration.`
```suggestion
              method.response.header.Cache-Control: "max-age=3600, private"
```

chris-dco (changes requested):

chris-dco:
Can you test this on dev please? It doesn't look tested at the moment.

deploy/template.yaml
chris-dco: If you find that this doesn't work give this a go:
```yaml
ResponseModels:
  application/did+json: !Ref DidJsonModel

# In the resources section
DidJsonModel:
    Type: AWS::ApiGateway::Model
    Properties:
      ContentType: 'application/did+json'
      RestApiId: !Ref RestApiGateway
      Schema: {}
```

chris-dco (approved):

blakeyp (approved):

PR #1438, opened 2024-05-13
PYIC-5809: rename action for go back in page-update-name to be more generic

sam803:
### What changed

Update the goback action in update name page to be more generic since this [PR](https://github.com/govuk-one-login/ipv-core-back/pull/1910) makes this action name misleading

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-5809

### DIFF
--- a/src/views/ipv/page/page-update-name.njk
+++ b/src/views/ipv/page/page-update-name.njk
@@ -68,7 +68,7 @@
           divider: 'pages.pageUpdateName.content.formRadioButtons.separateOptionsInFormText' | translate
         },
         {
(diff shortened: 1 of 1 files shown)


Wynndow (approved):

Wynndow:
Part of me is wondering if we should update the `update-name-date-birth` page to use `back` instead of `end` to be consistant.

sam803:
> Part of me is wondering if we should update the `update-name-date-birth` page to use `back` instead of `end` to be consistant.

yes I did consider it and was in two minds as there is only two options on `update-name-date-birth`

PR #1440, opened 2024-05-14
PYIC-5960: Add new confirm details page

sam803:
## Proposed changes

- Added new page to give user option to check and update their details

### What changed

 - new page added (rather than editing existing confirm-name-date-birth as the pages vary a lot)
 - added middleware to determine the error type from this page as there can be two different errors depending on which radio/check boxes are selected

### Why did it change

To allow user to update their name/address/dob from the 6MFC start page

### Issue tracking
- https://govukverify.atlassian.net/browse/PYIC-5960


src/app/ipv/router.js
```diff
@@ -49,6 +51,17 @@ router.post(
   handleJourneyAction,
 );
 
+// Special case to handle determination of COI journey type based on the checkboxes selected and determine the error type
+router.post(
+  getPagePath(CONFIRM_DETAILS),
+  parseForm,
+  csrfProtection,
+  formHandleUpdateDetailsCheckBox,
```
MikeCollingwood: Nice re-use of the existing handler, would it make sense to put the formHandleCoiDetailsCorrect before forming the req.body.journey?

src/app/ipv/middleware.js
```diff
@@ -480,6 +480,31 @@ module.exports = {
       next(error);
     }
   },
+  formHandleCoiDetailsCorrect: async (req, res, next) => {
```
MikeCollingwood: Could there be a different name for this handler? I know the DetailsCorrect comes from the attribute but I think it makes it seem as though the handler is for the case that they are correct

MikeCollingwood (commented):
Looks good, just had a couple of thoughts

src/app/ipv/router.js
sam803: thanks, yes I thought about that but would have to add a check in `formHandleUpdateDetailsCheckBox` to check if the journey != undefined (in the case of the user selects that their details are already correct) so it doesn't get overwritten, so wasn't sure what the best approach was

src/app/ipv/middleware.js
sam803: Yes sure, any suggestions? How about `formHandleConfirmDetailsErrors` ?

src/app/ipv/middleware.js
```diff
@@ -480,6 +480,31 @@ module.exports = {
...
+    try {
+      const { context, currentPage } = req?.session || "";
```
MikeCollingwood: Ah one thought here is I believe we can't destructure from the string, so maybe the default should be `{}`. I know it's used elsewhere in core-front but there's a ticket to clear it up, so you could let it be cleared up by [PYIC-5606](https://govukverify.atlassian.net/browse/PYIC-5606)

src/app/ipv/middleware.js
MikeCollingwood: That looks good though could maybe be coi-specific with `formHandleCoiDetailsCheck`?

src/views/ipv/page/confirm-details.njk
```diff
@@ -0,0 +1,141 @@
+{% extends "shared/base.njk" %}
```
Wynndow: Looking at the figma, I think this needs to be `confirm-your-details.njk`

src/app/ipv/router.js
MikeCollingwood: I see the dilemma. Could it make sense to use `formHandleUpdateDetailsCheckBox` inside `formHandleCoiDetailsCheck`? Since we only need to use the checkbox handler in the case where `(req.body.detailsCorrect === "no" && req.body.detailsToUpdate)`

It could be an extra else if, e.g.
```javascript
  formHandleCoiDetailsCheck: async (req, res, next) => {
    try {
      const { context, currentPage } = req?.session || {};
      if (req.body.detailsCorrect === "yes") {
        ...
      } else if (
        req.body.detailsCorrect === "no" && req.body.detailsToUpdate
      ) {
        formHandleUpdateDetailsCheckBox(req, res, next);
        // Or this maybe: req.body.journey = getCoiUpdateDetailsJourney(req.body.detailsToUpdate);
      } else if (
        !req.body.detailsCorrect ||
        (req.body.detailsCorrect === "no" && !req.body.detailsToUpdate)
      ) {
        ...
      }
      next();
    } catch (error) {
      next(error);
    }
  },
```

sam803: good shout, will add

MikeCollingwood (approved):
Looks good to me!

PR #1466, opened 2024-06-12
LS-2142 Review Autoscaling Policy

andyloughran:
### What changed

See ticket LS-2142; but I reviewed the scaling policy against the perf test on the 7th June to see how the application and the scaling policies behaved.

### Why did it change

The application did not scale fast enough under load.

### Issue tracking

- [LS-2142](https://govukverify.atlassian.net/browse/LS-2142)

## Checklists

[ ] - Review of configuration to make sure this is appropriate for implementation within the IPV-Core FE.
[ ] - Need to understand the impact the stubs performance will have had on the core application.

[LS-2142]: https://govukverify.atlassian.net/browse/LS-2142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -594,8 +594,7 @@ Resources:
         Cooldown: 30
         MinAdjustmentMagnitude: 6
         StepAdjustments:
-          - MetricIntervalLowerBound: 0
-            MetricIntervalUpperBound: 20
+          - MetricIntervalUpperBound: 20
             ScalingAdjustment: 100
           - MetricIntervalLowerBound: 20
             MetricIntervalUpperBound: 30
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -79,7 +79,6 @@ Mappings:
       nodeOldSpaceLimit: "384"
       nodeEnv: "development"
       languageToggle: true
-      desiredTaskCount: 1
       ga4Disabled: false
       uaDisabled: false
       contactUrl: "https://home.build.account.gov.uk/contact-gov-uk-one-login"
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -520,10 +520,6 @@ Resources:
           - UseCanaryDeployment
           - CODE_DEPLOY
           - ECS
-      DesiredCount: !FindInMap
-        - EnvironmentConfiguration
-        - !Ref AWS::AccountId
-        - desiredTaskCount
```
Joe-Edwards-GDS, line 526: If we remove this, we can remove it from the `EnvironmentConfiguration` mapping as well.

Joe-Edwards-GDS (dismissed):
I think we also discussed removing the `CoreFrontStepScaleInPolicy` - but happy to do that as a separate change.

deploy/template.yaml
andyloughran: Yes, given that this value is now delegated to the AutoScaling policy

andyloughran: Done,

Joe-Edwards-GDS (dismissed):

deploy/template.yaml
```diff
@@ -595,9 +584,11 @@ Resources:
       ServiceNamespace: ecs
       StepScalingPolicyConfiguration:
         AdjustmentType: PercentChangeInCapacity
-        Cooldown: 120
-        MinAdjustmentMagnitude: 5
+        Cooldown: 30
+        MinAdjustmentMagnitude: 6
```
andyloughran:
```suggestion
        MinAdjustmentMagnitude: 3
```

Joe-Edwards-GDS (approved):

PR #1470, opened 2024-06-17
PYIC-6774: modifications to update name dob page

sam803:
## Proposed changes

Modifications to updateNameDateBirth page following UCD review

### What changed

 - Only show warning when repeatFraudCheck is the context
 - Use `full name `instead of `name` in radio button description
 - adds missing full stops
### Why did it change

following UCD review these slight inaccuracies were noted.

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6774

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -786,16 +786,19 @@
       "header": "Update your name or date of birth",
       "titleCoi": "Update your full name or date of birth",
       "headerCoi": "Update your full name or date of birth",
+      "titleRepeatFraudCheck": "Update your full name or date of birth",
+      "headerRepeatFraudCheck": "Update your full name or date of birth",
       "content": {
         "paragraph1": "To update your full name or date of birth, you need to contact the GOV.UK One Login team.",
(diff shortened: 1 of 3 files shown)


src/locales/en/translation.json
```diff
@@ -786,16 +786,19 @@
       "header": "Update your name or date of birth",
       "titleCoi": "Update your full name or date of birth",
       "headerCoi": "Update your full name or date of birth",
+      "titleRepeatFraudCheck": "Update your full name or date of birth",
+      "headerRepeatFraudCheck": "Update your full name or date of birth",
```
Wynndow, line 790: ~~As these are the same, can we just drop the context for these fields?~~

Ignore me, I just clocked the versions above

Wynndow: Although do we ever use this page _without_ context? If not we can drop the context versions of the content and just use `title` and `header` (after updating them).

Joe-Edwards-GDS: It's used for the current 6MFC journey (which doesn't allow you to update any name or DOB).

But once COI is live, we could make that simplification :)

Wynndow (approved):

PR #1487, opened 2024-07-02
LS-2225 - Updated IPV scaling policy

andyloughran:


Dockerfile
```diff
@@ -14,7 +14,7 @@ FROM node:20.12.0-alpine3.19@sha256:ef3f47741e161900ddd07addcaca7e76534a9205e4cd
 RUN addgroup -S appgroup && adduser -S appuser -G appgroup
 
 RUN ["apk", "--no-cache", "upgrade"]
-RUN ["apk", "add", "--no-cache", "tini"]
+RUN ["apk", "add", "--no-cache", "tini", "curl"]
```
andyloughran, line 17: Add curl to enable the healthcheck

deploy/template.yaml
```diff
@@ -639,7 +639,7 @@ Resources:
       Namespace: "AWS/ECS"
       Statistic: "Average"
       Period: "60"
-      Threshold: "60"
+      Threshold: "50"
```
andyloughran, line 643: Reducing to get the alarm to fire sooner in the ramp up.

deploy/template.yaml
```diff
@@ -662,7 +662,7 @@ Resources:
       Namespace: "AWS/ECS"
       Statistic: "Average"
       Period: "60"
-      Threshold: "60"
+      Threshold: "40"
```
andyloughran, line 666: Reducing as we don't need to scale down aggressively unless we're very quiet.

deploy/template.yaml
```diff
@@ -460,9 +460,10 @@ Resources:
       HealthCheckEnabled: TRUE
       HealthCheckProtocol: HTTP
       HealthCheckPath: /healthcheck
-      HealthCheckTimeoutSeconds: 14
-      HealthCheckIntervalSeconds: 15
+      HealthCheckTimeoutSeconds: 30
```
andyloughran, line 463: Might be worth reviewing these values - my changes here need discussion, they're not recommendations.

deploy/template.yaml
```diff
@@ -799,6 +800,13 @@ Resources:
           PortMappings:
             - ContainerPort: 8080
               Protocol: tcp
+          HealthCheck:
```
andyloughran, line 803: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-healthcheck.html

Joe-Edwards-GDS:
We've made other scaling changes that partially supersede this, so assuming this is stale now.

PR #1504, opened 2024-07-29
Pyic:5606 Tidy-up core-front

shivanshuit914:
## Proposed changes

- Remove button_text in pages since it is not set anywhere
- refactor `const { context } = req?.session || "";`
- Merge identical methods, e.g. tryValidateClientResponse & tryValidateCriResponse

### What changed

- Refactored middleware and tests
- refactored view template file to remove default button texts

### Why did it change

- To Tidy-up core-front

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -90,10 +90,7 @@ async function handleBackendResponse(req, res, backendResponse) {
     return await handleJourneyResponse(req, res, backendResponse.journey);
   }
 
-  if (
-    backendResponse?.cri &&
-    validateRedirectUrlInResponse(backendResponse.cri)
-  ) {
+  if (backendResponse?.cri && tryValidateCriResponse(backendResponse.cri)) {
     logCoreBackCall(req, {
       logCommunicationType: LOG_COMMUNICATION_TYPE_RESPONSE,
       type: LOG_TYPE_CRI,
@@ -107,7 +104,7 @@ async function handleBackendResponse(req, res, backendResponse) {
 
   if (
     backendResponse?.client &&
-    validateRedirectUrlInResponse(backendResponse.client, true)
+    tryValidateClientResponse(backendResponse.client)
   ) {
     logCoreBackCall(req, {
       logCommunicationType: LOG_COMMUNICATION_TYPE_RESPONSE,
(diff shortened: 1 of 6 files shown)


src/app/ipv/middleware.js
```diff
@@ -104,7 +107,7 @@ async function handleBackendResponse(req, res, backendResponse) {
 
   if (
     backendResponse?.client &&
-    tryValidateClientResponse(backendResponse.client)
+    validateRedirectUrlInResponse(backendResponse.client, true)
```
blakeyp: Hmm I'm not sure about this - not a fan of boolean flag args but maybe the function could just take a `response` (object of any shape) and check for the presence of `redirectUrl` in that object. But then we lose the specific error message (client vs CRI) so I'm not convinced this refactor is really worth it 😕

shivanshuit914: Yes, we have a same object for both responses and we can't differentiate.

src/app/ipv/middleware.js
```diff
@@ -156,19 +159,13 @@ async function handleBackendResponse(req, res, backendResponse) {
...
-
-function tryValidateClientResponse(client) {
-  const { redirectUrl } = client;
+function validateRedirectUrlInResponse(response, isClient = false) {
+  const { redirectUrl } = response;
 
   if (!redirectUrl) {
-    throw new Error("Client Response redirect url is missing");
+    throw new Error(
+      `${isClient ? "Client" : "CRI"} Response RedirectUrl is missing`,
```
MikeCollingwood: Does it want to be specific about where the response comes from? I imagine it would be obvious from context but maybe I'm wrong?

shivanshuit914: We receive response from core-back, CRI is from CRI provider and cleint not 100% sure

MikeCollingwood: The refactoring might not be worth it if we want to keep the log message specific to the source then... I imagined that "Response RedirectUrl is missing" would be fine since given context you'd know where it'd come from?

blakeyp: Client is orchestration 
It should be clear from a sequence of logs in a journey yes

src/app/ipv/middleware.js
shivanshuit914: Removed it, thanks

blakeyp (approved):

PR #1508, opened 2024-07-30
PYIC-7105 frontend scaling comments and healthcheck logging

Joe-Edwards-GDS:
### What changed

- Added comments to the step scaling policy
- Added logging to the healthcheck endpoint

### Why did it change

- To help explain how the step scaling policy works
- To help analyse how and when healthcheck requests are made


deploy/template.yaml
Joe-Edwards-GDS, line 1: I have not changed the policy, just added explanatory commentary (so we can test changes in a controlled manner)

sam803 (approved):

PR #1509, opened 2024-08-05
PYIC-5598: Standardise page creation

MikeCollingwood:
### What changed

- Add views README
- Create common contact-us-link.njk
- Standardise pages
- Remove renderAttemptRecoveryPage

### Why did it change

- To make pages easier to understand and more consistent

### Issue tracking

- [PYIC-5598](https://govukverify.atlassian.net/browse/PYIC-5598)

[PYIC-5598]: https://govukverify.atlassian.net/browse/PYIC-5598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -319,11 +319,6 @@ async function handleUnexpectedPage(req, res, pageId) {
 }
 
 module.exports = {
-  renderAttemptRecoveryPage: async (req, res) => {
-    res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
-      csrfToken: req.csrfToken(),
-    });
-  },
   updateJourneyState: async (req, res, next) => {
     try {
       const currentPageId = req.params.pageId;
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -319,6 +319,11 @@ async function handleUnexpectedPage(req, res, pageId) {
 }
 
 module.exports = {
+  renderAttemptRecoveryPage: async (req, res) => {
+    res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
+      csrfToken: req.csrfToken(),
+    });
+  },
   updateJourneyState: async (req, res, next) => {
     try {
       const currentPageId = req.params.pageId;
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -4,6 +4,7 @@ const bodyParser = require("body-parser");
 const router = express.Router();
 
 const {
+  renderAttemptRecoveryPage,
   updateJourneyState,
   handleJourneyPage,
   handleJourneyAction,
@@ -16,6 +17,7 @@ const {
 } = require("./middleware");
 
 const {
+  PYI_ATTEMPT_RECOVERY,
   UPDATE_DETAILS,
   CONFIRM_DETAILS,
 } = require("../../constants/ipv-pages");
@@ -28,6 +30,12 @@ function getPagePath(pageId) {
 }
 
 router.get("/usefeatureset", validateFeatureSet, renderFeatureSetPage);
+
+router.get(
+  getPagePath(PYI_ATTEMPT_RECOVERY),
+  csrfProtection,
+  renderAttemptRecoveryPage,
+);
 
 router.get(getPagePath(":pageId"), csrfProtection, handleJourneyPage);
 
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -4,7 +4,6 @@ const bodyParser = require("body-parser");
 const router = express.Router();
 
 const {
-  renderAttemptRecoveryPage,
   updateJourneyState,
   handleJourneyPage,
   handleJourneyAction,
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -17,7 +17,6 @@ const {
 } = require("./middleware");
 
 const {
-  PYI_ATTEMPT_RECOVERY,
   UPDATE_DETAILS,
   CONFIRM_DETAILS,
 } = require("../../constants/ipv-pages");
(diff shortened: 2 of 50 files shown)


src/app/ipv/middleware.js
```diff
@@ -319,11 +319,6 @@ async function handleUnexpectedPage(req, res, pageId) {
 }
 
 module.exports = {
-  renderAttemptRecoveryPage: async (req, res) => {
-    res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
-      csrfToken: req.csrfToken(),
```
MikeCollingwood, line 324: This is the same as handleJourneyPage for PYI_ATTEMPT_RECOVERY

Screenshot 2024-08-05 at 09 33 06

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+  - And `csrfProtection()` checks it, in the POST route handler.
+  - Arguably not a required security feature since our forms only lead to change in position on a journey map.
+
+- Variable with the config for the govukRadios component. Represents key value pair where key is "journey".
+    ```html
+    {% set radiosConfig = {...} %}
+    ```
+  - The values given to the radio map directly with the event the state associated will produce, so keep to camelCase.
+
+- This is the conditionally-visible error message inside radio.
```
MikeCollingwood, line 117: I'm going to elaborate here

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+     - `qrCode`
+       - QR code to download the mobile doc checking app, dynamically input because can be IOS/ Android.
+     - `appDownloadUrl`
+       - Link to download the mobile doc checking app.
+     - `pageErrorState`
+       - Boolean indicating a form validating error on that page.
+     - `errorState`
+       - String indicating type of form error (`checkbox`/ `radiobox`).
+2. English/ Welsh content in `src/locales`.
+3. The `res.locals` variable set in `locales.js`.
```
blakeyp: *`locals.js` ?

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+   - Variables here can be referenced in Nunjucks like `"{{contactUsUrl}}"`.
+2. English/ Welsh content:
+   - These values are referenced in the template in single quotation marks, e.g.:
+    ```
+    {% set errorTitle = 'pages.pageMultipleDocCheck.content.formErrorMessage.errorSummaryTitleText' | translate %}
+    ```
+   - `translate` & variants are defined in `src/config/nunjucks.js`.
+     - It takes the reference as a key to find the content saved in `src/locales` files for english/ welsh.
+     - It can take secondary params, e.g.: `context`, which adds a suffix to the reference it's going to find.
+3. `locales.js`:
```
blakeyp: `local.js`

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+    {% extends "shared/base.njk" %}
+    ```
+  - Relevant details:
+    - `pageTitleKey` is used for finding title from content, and `pageErrorState` may trigger an "Error: " prefix.
+    `showLanguageToggle` toggles the language toggle.
+    - `showBack` toggles the back link.
+    - `errorState` toggles error summary, which includes: `errorTitle`, `errorText` and `errorHref`.
+    - `googleTagManagerPageId` sets state.
+
+- Import common, standard GovUK components.
```
blakeyp: add something like "only if they are used" ?

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+    ```
+
+- Hidden input for carrying the token from the GET request to the POST request.
+    ```html
+    <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+    ```
+  - `req.csrfToken()` creates the token in the GET route handler
+  - The result is saved in here
+  - And `csrfProtection()` checks it, in the POST route handler.
+  - Arguably not a required security feature since our forms only lead to change in position on a journey map.
```
blakeyp: Could we remove this - it's a little subjective and we do have a requirement to protect all our forms to prevent user actions being taken by someone/something malicious even if it is only navigating them to the next page

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+        id: "submitButton",
+        text: 'general.buttons.next' | translate
+    }) }}
+    ```
+
+- This is a (relatively) common link we have at the end of the page.
+    ```html
+    {% include "components/contact-us-link.njk" %}
+    ```
+    - The `contactUsUrl` is created in `locales.js` to include the `fromUrl` query.
```
blakeyp: `locals.js`

src/views/README.md
```diff
@@ -0,0 +1,159 @@
...
+        csrfProtection,
+        formRadioButtonChecked,
+        handleJourneyAction,
+    );
+    ```
+
+### Notes
+
+- When adding in a feature, check if a specific govUk component exists before styling a generic tag.
+- When updating a journey map to use the page it must produce consistent events, for a certain `context`, across the journey maps. This is enforced with unit tests.
```
blakeyp, line 158: Is it also worth referring to the Design System docs and Nunjucks docs?

src/views/README.md
```diff
@@ -0,0 +1,159 @@
+# Pages
```
blakeyp, line 1: To make this visible is it worth referencing to it from the root README?

src/app/ipv/router.js
```diff
@@ -31,12 +29,6 @@ function getPagePath(pageId) {
 
 router.get("/usefeatureset", validateFeatureSet, renderFeatureSetPage);
 
-router.get(
```
blakeyp: Is there a reason to remove this?

src/views/README.md
MikeCollingwood: Ah thank you for the clarification. Will remove appreciate it's a bit subjective

src/app/ipv/router.js
MikeCollingwood: `handleJourneyPage` seems like it would perform the same as `renderAttemptRecoveryPage` for this page. It also gives the opportunity to get the `attempt-recovery` page closer to others as well by providing it with the `pageId` and using it to determine the post url in the form

blakeyp: I guess the difference is it just renders the page directly without checking for session id etc?

MikeCollingwood: I see - I know we point people to it if there's an oauth error so I imagine it could be the case they don't have these and get pointed to the technical error page, losing information in the process. I will revert this 👍

blakeyp (approved):

PR #1512, opened 2024-08-06
PYIC-7043 Add X-Forwarded-Host header

Joe-Edwards-GDS:
Adds an `X-Forwarded-Host` header so Dynatrace can detect the application

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1003,7 +1003,7 @@ Resources:
         Types:
           - REGIONAL
 
-  RestApiGwDeployment202405141531:
+  RestApiGwDeployment202408070900:
     DependsOn:
       - RestApiGatewayMethod
       - RestDidDocumentMethod
@@ -1026,6 +1026,7 @@ Resources:
       RequestParameters:
         method.request.path.proxy: false
         method.request.header.x-forwarded-for: false
+        method.request.header.host: false
       Integration:
         Type: HTTP_PROXY
         ConnectionType: VPC_LINK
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -1026,6 +1026,7 @@ Resources:
       RequestParameters:
         method.request.path.proxy: false
         method.request.header.x-forwarded-for: false
+        method.request.header.host: false
```
Joe-Edwards-GDS, line 1029: This defines `host` as a possible request header

deploy/template.yaml
```diff
@@ -1036,6 +1037,7 @@ Resources:
         RequestParameters:
           integration.request.path.proxy: "method.request.path.proxy"
           integration.request.header.x-forwarded-for: "method.request.header.x-forwarded-for"
+          integration.request.header.x-forwarded-host: "method.request.header.host"
```
Joe-Edwards-GDS, line 1040: This adds an `x-forwarded-host` header on the response, with the value of the `host` header

deploy/template.yaml
```diff
@@ -1003,7 +1003,7 @@ Resources:
         Types:
           - REGIONAL
 
-  RestApiGwDeployment202405141531:
+  RestApiGwDeployment202408070900:
```
Joe-Edwards-GDS, line 1006: Forces an API gateway deployment

vam-gds (approved):
Joe, Ive approved but make sure the headers required for txma are getting sent through still (the ip addresses / that they arent effected - this doesnt like it would do anything to them and should be fine)

PR #1514, opened 2024-08-07
PYIC-7043 fix template and add SAM validate

Joe-Edwards-GDS:
Cloudformation is broken due to a bad reference.

Also added a SAM validate step to the pre-merge checks to avoid it happening again!

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -320,6 +320,11 @@ Resources:
   AccessLogsBucket:
     Type: AWS::S3::Bucket
     #checkov:skip=CKV_AWS_18: This is the bucket where our access logs go and AWS advise not sending a bucket's access logs to itself.
+    Metadata:
+      cfn-lint:
+        config:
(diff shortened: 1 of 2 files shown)


deploy/template.yaml
```diff
@@ -320,6 +320,11 @@ Resources:
   AccessLogsBucket:
     Type: AWS::S3::Bucket
     #checkov:skip=CKV_AWS_18: This is the bucket where our access logs go and AWS advise not sending a bucket's access logs to itself.
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks:
+            - W3045 # Using a legacy access control setup for now
```
Joe-Edwards-GDS, line 327: cc @blakeyp might be nice to fix this, but felt out of scope for now. AWS recommend using a bucket policy on the access logs instead of `LogDeliveryWrite` (which uses ACLs).

sam803 (approved):

PR #1519, opened 2024-08-12
BAU: Manually upgrade fast-xml-parser

AmritSidhu:
### What changed

fast-xml-parser vulnerable to ReDOS at currency parsing

### Why did it change

fixing a dependabot security alert

### Issue tracking

- [PYIC-5673](https://govukverify.atlassian.net/browse/PYIC-5673)

[PYIC-5673]: https://govukverify.atlassian.net/browse/PYIC-5673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "express": "4.19.2",
     "express-async-errors": "3.1.1",
     "express-session": "1.18.0",
+    "fast-xml-parser": "^4.4.1",
     "govuk-frontend": "4.8.0",
     "i18next": "23.10.0",
     "i18next-fs-backend": "2.3.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -69,6 +69,7 @@
     "express": "4.19.2",
     "express-async-errors": "3.1.1",
     "express-session": "1.18.0",
+    "fast-xml-parser": "^4.4.1",
```
blakeyp, line 72: We don't use this package directly so we should probably find the package that uses it and upgrade that if we can - looking in package-lock it looks like `@aws-sdk/client-sts` uses this which itself is a dependency of `@aws-sdk/client-dynamodb` - can we try updating this to the latest?

PR #1523, opened 2024-08-13
PYIC-6914: Add cookies parser and language to call to backend

MikeCollingwood:
### What changed

- Add cookies parser and language to call to backend

### Why did it change

- To allow us to use the lng cookie in the backend

### Issue tracking

- [PYIC-6914](https://govukverify.atlassian.net/browse/PYIC-6914)

### Notes

Must merge this before https://github.com/govuk-one-login/ipv-core-back/pull/2288

[PYIC-6914]: https://govukverify.atlassian.net/browse/PYIC-6914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/router.js
+++ b/src/app/ipv/router.js
@@ -1,6 +1,7 @@
 const express = require("express");
 const csrf = require("csurf");
 const bodyParser = require("body-parser");
+const cookieParser = require("cookie-parser");
 const router = express.Router();
 
 const {
@@ -48,6 +49,7 @@ router.post(
   csrfProtection,
   formHandleUpdateDetailsCheckBox,
   formRadioButtonChecked,
+  cookieParser(),
   handleJourneyAction,
 );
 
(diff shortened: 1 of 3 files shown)


src/app/ipv/router.js
```diff
@@ -1,6 +1,7 @@
 const express = require("express");
 const csrf = require("csurf");
 const bodyParser = require("body-parser");
+const cookieParser = require("cookie-parser");
```
MikeCollingwood: Required to get cookies from frontend into req.cookies

DanCorderIPV: Do we need to do this here? `cookieParser` is already included as app middleware (app.js line 130) which I think means that it should be running on every request already.

DanCorderIPV (approved):

PR #1524, opened 2024-08-13
PYIC-6156: Updated handleJourneyPage/Action session validation

AmritSidhu:
### What changed

The session validation logic (checking for ipvSessionId or clientOauthSessionId) is now the same in both the GET (handleJourneyPage) and POST (handleJourneyAction) handlers. 
The GET handler is updated to accommodate users with either ipvSessionId or criSessionId, ensuring users returning from the mobile app are not excluded. 
With the new validation approach, both the GET and POST handlers now check for either an ipvSessionId or a clientOauthSessionId across all pages. 
This unified validation logic removes the need to differentiate between pages using pagesUsingSessionId.

### Why did it change

The GET validation is not the same as for POST. So people can/ can't get onto a page with different conditionality to whether they can post to it

### Issue tracking

- [PYIC-6156](https://govukverify.atlassian.net/browse/PYIC-6156)

[PYIC-6156]: https://govukverify.atlassian.net/browse/PYIC-6156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -174,20 +174,6 @@ function tryValidateClientResponse(client) {
   return true;
 }
 
-function checkForSessionId(req, res) {
-  if (!req.session?.ipvSessionId) {
-    const err = new Error("req.ipvSessionId is missing");
-    err.status = HTTP_STATUS_CODES.UNAUTHORIZED;
-    logError(req, err);
-
-    req.session.currentPage = PAGES.PYI_TECHNICAL;
-    res.status(HTTP_STATUS_CODES.UNAUTHORIZED);
-    return res.render(getIpvPageTemplatePath(PAGES.PYI_TECHNICAL), {
-      context: "unrecoverable",
-    });
-  }
-}
-
 function checkForIpvAndOauthSessionId(req, res) {
   if (!req.session?.ipvSessionId && !req.session?.clientOauthSessionId) {
     const err = new Error(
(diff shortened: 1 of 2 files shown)


src/app/ipv/middleware.js
```diff
@@ -345,27 +331,28 @@ module.exports = {
...
       const { context } = req?.session || "";
 
-      // handles page id validation first
       if (!isValidPage(pageId)) {
         res.status(HTTP_STATUS_CODES.NOT_FOUND);
         return res.render(getTemplatePath("errors", "page-not-found"));
       }
 
-      if (!req.session?.ipvSessionId) {
+      if (!req.session?.ipvSessionId && !req.session?.criSessionId) {
```
blakeyp: Hmm I don't think `criSessionId` is something we store in the frontend session? in the mobile cross-browser scenario we store the `clientOauthSessionId` so that it can be used to relay the user back to their RP and it's in this scenario where we might have `clientOauthSessionId` but no `ipvSessionId`

blakeyp (changes requested):

src/app/ipv/middleware.js
AmritSidhu: Yes you are right, I typed out the wrong one here

shivanshuit914 (dismissed):

PR #1527, opened 2024-08-14
PYIC-7161: added delete identity page

sam803:
### What changed

Added a new screen for deleting an GOV UK onelogin identity

### Why did it change
To reduce the likelihood that users are directed to contact the call centre for support

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7161


src/views/ipv/page/delete-identity.njk
Joe-Edwards-GDS, line 1: I don't think this was on the ticket (sorry), but the Figma specifies that the URL should be `/delete-handover`

Joe-Edwards-GDS (commented):
Looks good, but one comment about the URL

Joe-Edwards-GDS (approved):

PR #1528, opened 2024-08-15
Pyic 6025 commonise form page rendering

DanCorderIPV:
NOTE - It might be easier to look at each commit in this PR separately

### What changed

Commonise page rendering when re-displaying due to a form validation error

### Why did it change

To make the code more robust and consistent

### Issue tracking

- [PYIC-6025](https://govukverify.atlassian.net/browse/PYIC-6025)

[PYIC-6025]: https://govukverify.atlassian.net/browse/PYIC-6025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -390,8 +390,6 @@ module.exports = {
         renderOptions.appDownloadUrl = appDownloadHelper.getAppStoreRedirectUrl(
           parseContextAsPhoneType(context),
         );
-      } else if (req.query?.errorState !== undefined) {
-        renderOptions.pageErrorState = req.query.errorState;
       } else if (req.session.currentPageStatusCode !== undefined) {
         res.status(req.session.currentPageStatusCode);
       }
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -319,208 +319,226 @@ async function handleUnexpectedPage(req, res, pageId) {
   );
 }
 
-module.exports = {
-  renderAttemptRecoveryPage: async (req, res) => {
-    res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
-      csrfToken: req.csrfToken(),
-    });
-  },
-  updateJourneyState: async (req, res, next) => {
-    try {
-      const currentPageId = req.params.pageId;
-      const { action } = req.params;
+async function renderAttemptRecoveryPage(req, res) {
+  res.render(getIpvPageTemplatePath(PAGES.PYI_ATTEMPT_RECOVERY), {
+    csrfToken: req.csrfToken(),
+  });
+}
 
-      if (action && isValidPage(currentPageId)) {
-        await handleJourneyResponse(req, res, action, currentPageId);
-      } else {
-        res.status(HTTP_STATUS_CODES.NOT_FOUND);
-        return res.render(getTemplatePath("errors", "page-not-found"));
-      }
-    } catch (error) {
-      next(error);
+async function updateJourneyState(req, res, next) {
+  try {
+    const currentPageId = req.params.pageId;
+    const { action } = req.params;
+
+    if (action && isValidPage(currentPageId)) {
+      await handleJourneyResponse(req, res, action, currentPageId);
+    } else {
+      res.status(HTTP_STATUS_CODES.NOT_FOUND);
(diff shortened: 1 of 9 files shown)


src/app/ipv/middleware.js
```diff
@@ -319,213 +313,237 @@ async function handleUnexpectedPage(req, res, pageId) {
...
-        );
-      } else if (req.query?.errorState !== undefined) {
-        renderOptions.pageErrorState = req.query.errorState;
-      } else if (req.session.currentPageStatusCode !== undefined) {
-        res.status(req.session.currentPageStatusCode);
-      }
+    // handles page id validation first
+    if (!isValidIpvPage(pageId)) {
+      render404(res);
+      return;
```
blakeyp: Is there a difference between this and `return render404(res);`? because I can see in a few places we still do `return res.render` and I know @AmritSidhu has seen some issues around this too

src/app/ipv/middleware.js
```diff
@@ -319,213 +313,237 @@ async function handleUnexpectedPage(req, res, pageId) {
...
-      const renderOptions = {
-        pageId,
-        csrfToken: req.csrfToken(),
-        context,
-      };
+async function handleJourneyPageInternal(
+  req,
+  res,
+  next,
+  pageErrorState = undefined,
```
blakeyp: I'm guessing Express doesn't like it if you put this on `handleJourneyPage` itself? maybe this 'internal' function could be a bit better named something like 'renderJourneyPageWithOptionalError`

src/app/ipv/middleware.js
DanCorderIPV: I don't think there's a difference - `render404()` returns void. If I see `return render404(res)` then I assume that there is a useful return value being passed along which I find misleading.
However, I think we probably do need to do `return next()` when that comes up because we don't know if `next()` is an async function (that returns `Promise<void>` instead of just `void`), so it's a bit of a mess :(

src/app/ipv/middleware.js
DanCorderIPV: Oh, that's a good point, I didn't try that

src/app/ipv/middleware.js
Joe-Edwards-GDS: Functionally there's no difference, but some people do it by convention to avoid mistakes where you accidentally try and render twice.

e.g. we had a bug at one point along the lines of:
```
if (someCondition) {
  render(somePage);
}
render(someDefaultPage);
```
if you always return, then you can't (easily) double-render... but the flipside is that as Dan says, it's a bit misleading because it's not actually returning anything.

Joe-Edwards-GDS: (I don't have a strong preference, but we should probably be consistent)

DanCorderIPV: It's probably safer to do `return render404()` everywhere.
There are places where the IntelliJ linter complains if you do
```
render404();
return;
```
But if you delete the `return` and then someone else adds more code in the future they could introduce a bug.

DanCorderIPV: Hmm, actually, ignore what I said about `return next();` `next()` is a wrapper provided by express rather than directly being the next middleware function.
Although for a similar reason to `return res.render();` `return next()` might be a good idea anyway.

src/app/ipv/middleware.js
```diff
@@ -347,16 +347,7 @@ async function updateJourneyState(req, res, next) {
...
-  return handleJourneyPageInternal(req, res, next);
-}
-
-async function handleJourneyPageInternal(
-  req,
-  res,
-  next,
-  pageErrorState = undefined,
-) {
+async function handleJourneyPage(req, res, next, pageErrorState = undefined) {
```
blakeyp, line 348: given it's a boolean maybe `pageHasErrorState` and default to false?

DanCorderIPV: It's not always a boolean :(

DanCorderIPV: Although, I guess in this function it is.
I was trying to keep the existing behaviour where if there isn't an error then pageErrorState just doesn't exist in the render options

blakeyp (approved):

PR #1529, opened 2024-08-15
PYIC-7160: add routing to delete account page when pressing continue

thebauSoftwire:
### What changed
Adds redirect to delete account when pressing continue on `delete-handover` page

### Why did it change
User can delete their account themselves to reduce the number of users contacting the call centre for help.

### Issue tracking

- [PYIC-7160](https://govukverify.atlassian.net/browse/PYIC-7160)

[PYIC-7160]: https://govukverify.atlassian.net/browse/PYIC-7160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/locals.js
+++ b/src/lib/locals.js
@@ -8,18 +8,15 @@ const {
   GA4_DISABLED,
   DT_RUM_URL,
   LOGOUT_URL,
-  DELETE_ACCOUNT_URL
+  DELETE_ACCOUNT_URL,
 } = require("./config");
 const { generateNonce } = require("./strings");
 
-const generateUrlFromString = (url, req) => {
-  const url = new URL(CONTACT_URL);
-  contactUsUrl.searchParams.set(
-    "fromUrl",
-    `${SERVICE_URL}${req.originalUrl}`,
-  );
+const generateUrlFromString = (stringURl, req) => {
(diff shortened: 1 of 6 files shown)


src/lib/locals.js
```diff
@@ -8,9 +8,16 @@ const {
   GA4_DISABLED,
   DT_RUM_URL,
   LOGOUT_URL,
+  DELETE_ACCOUNT_URL,
 } = require("./config");
 const { generateNonce } = require("./strings");
 
+const generateUrlFromString = (stringURl, req) => {
```
Joe-Edwards-GDS: This method should be `addFromUrlParam` or similar, I think?

Do we know whether that's desirable for the delete account link - it feels like it might be specific to the contact form...

Joe-Edwards-GDS (commented):
Nice - just one question about the url transformation

src/lib/locals.js
thebauSoftwire: you're right! My mistake for not reading it properly

sam803 (approved):

PR #1532, opened 2024-08-15
PYIC-7163: add missing back link to delete handover page

sam803:
## Proposed changes

Adds a hrefBack value for the delete handover page which was missed in https://github.com/govuk-one-login/ipv-core-front/pull/1527

 
### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7161


Joe-Edwards-GDS (approved):
I wonder if we should combine `showBack` and `hrefBack` 🤔 (is it ever valid to have one without the other?)

This is fine though

PR #1538, opened 2024-08-16
PYIC-7208: update types of photoid for dropdown content

shivanshuit914:
## Proposed changes

- update types of photo id for dropdown content for drop out pages.

### What changed

-  updated types of photo id for dropdown content for drop out pages.

### Issue tracking

- [PYIC-7208](https://govukverify.atlassian.net/browse/PYIC-7208)

Screenshot 2024-08-16 at 10 04 34
Screenshot 2024-08-16 at 10 04 49
Screenshot 2024-08-16 at 10 10 30

[PYIC-7208]: https://govukverify.atlassian.net/browse/PYIC-7208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/locales/cy/translation.json
```diff
@@ -302,8 +302,8 @@
...
-          "textDropout": "<p class=\"govuk-body\">Gallwch ddefnyddio: </p><ul class=\"govuk-list--bullet\"><li>Trwydded yrru cerdyn-llun yn y DU</li><li>Trwydded yrru cerdyn-llun yr Undeb Ewropeaidd (UE)</li><li>Pasbort y DU</li><li>pasbort o'r tu allan i'r DU</li><li>Cerdyn adnabod cenedlaethol o wlad Ardal Economaidd Ewropeaidd (AEE)</li><li>Trwydded breswylio biometrig y DU (BRP)</li><li>Cerdyn preswylio biometrig y DU (BRC)</li><li>Trwydded Gweithwyr Ffiniol y DU (FWP)</li></ul>",
+          "text": "<p class=\"govuk-body\">For the GOV.UK ID Check app, you can use a:</p><ul class=\"govuk-list--bullet\"><li>UK photocard driving licence</li><li>UK passport</li><li>non-UK passport with a biometric chip</li><li>UK biometric residence permit (BRP)</li><li>UK biometric residence card (BRC)</li><li>UK Frontier Worker permit</li></ul><p class=\"govuk-body\">At a Post Office, you can use a:</p><ul class=\"govuk-list--bullet\"><li>UK photocard driving licence</li><li>European Union (EU) photocard driving licence</li><li>UK passport</li><li>non-UK passport</li><li>National identity card from a European Economic Area (EEA) country</li><li>UK biometric residence permit (BRP)</li></ul>",
```
blakeyp: I'm not a big fan of having HTML in content files and it especially won't lend itself well to the new translation process where we'll be passing this JSON to the translators - can we try and split out the lists and move the HTML into the template?

shivanshuit914: Yes makes sense, updated it.

Had to create variable for html, there is no other way of adding html.

blakeyp: Looks much better 👍

blakeyp (approved):
Rule of 3 - don't think we need to yet but if we end up with more instances of this details components between templates we should think about pulling it out into its own template/include and define common content

PR #1541, opened 2024-08-16
PYIC-7204: update content for the pyi start page

shivanshuit914:
## Proposed changes

- Update content for the `page-ipv-identity-document-start` page

### What changed

- Updated the content

### Issue tracking

- [PYIC-7204](https://govukverify.atlassian.net/browse/PYIC-7204)

Screenshot 2024-08-16 at 15 43 15

[PYIC-7204]: https://govukverify.atlassian.net/browse/PYIC-7204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blakeyp (commented):
Looks good but HOLD this PR until we add Welsh translations

src/app/ipv/middleware.js
```diff
@@ -320,6 +320,15 @@ async function renderAttemptRecoveryPage(req, res) {
   });
 }
 
+async function renderStaticPage(req, res) {
```
blakeyp: could this take the page id as a parameter? or if not directly could it return a middleware function so we can call it like `staticPageMiddleware(PAGES.PAGE_IPV_IDENTITY_DOCUMENT_TYPES)`

src/app/ipv/middleware.js
```diff
@@ -320,6 +320,15 @@ async function renderAttemptRecoveryPage(req, res) {
...
+  return res.render(
+    getIpvPageTemplatePath(PAGES.PAGE_IPV_IDENTITY_DOCUMENT_TYPES),
+    {
+      csrfToken: req.csrfToken(),
```
blakeyp: Assuming there are no forms on static pages we don't need a csrf token?

blakeyp:
Looks good but HOLD until we add Welsh translations

src/locales/cy/translation.json
```diff
@@ -1019,28 +1019,19 @@
...
-        ],
-        "subHeading2": "Pasbort y DU",
-        "paragraph3": "Gallwch ddefnyddio unrhyw pasbort y DU. Os yw’ch pasbort y DU wedi dod i ben, gallwch barhau i’w ddefnyddio i brofi eich hunaniaeth hyd at 18 mis ar ôl ei ddyddiad dod i ben.",
-        "subHeading3": "Pasbort o’r tu allan i’r DU gyda sglodion biometrig",
-        "paragraph4": "Gallwch ddefnyddio pasbort o’r tu allan i’r DU os:",
-        "nonUkPassportRequirements": [
-          "mae ganddo sglodyn biometrig",
-          "nid yw wedi dod i ben"
+        "paragraph1": "Gallwch ddefnyddio un o’r mathau canlynol o ID gyda llun:",
+        "documentsList": [
```
AmritSidhu: We have a reusable block for this "typesOfPhotoId.appDocumentList", these translations should be moved there

src/locales/cy/translation.json
```diff
@@ -1053,29 +1044,29 @@
...
-        "imgAltText": "The front cover of a passport with the biometric chip symbol. The symbol is rectangular with a circle in the middle.",
-        "subHeading4": "UK biometric residence permit (BRP)",
-        "paragraph6": "You can use a UK biometric resident permit (BRP) that has not expired. You might have a BRP if you have a visa for longer than 6 months or indefinite leave to remain.",
-        "subHeading5": "UK biometric residence card (BRC)",
-        "paragraph7": "You can use a BRC that has not expired. You might have a BRC if you’re in the UK as a family member of a citizen from a European Economic Area (EEA) country.",
-        "subHeading6": "UK Frontier Worker permit (FWP)",
-        "paragraph8": "You can use an FWP that has not expired. You might have an FWP if you work in the UK but live in another country."
+        "paragraph5": "I wirio os oes gan eich pasbort sglodyn biometrig, edrychwch am y symbol sglodyn biometrig petryal ar y clawr blaen.",
+        "imgAltText": "Clawr blaen pasbort gyda'r symbol sglodyn biometrig. Mae'r symbol yn betryal gyda chylch yn y canol.",
```
AmritSidhu: Was you able to confirm if we needed the special character, I see it used elsewhere now `’` for welsh translations. If so it should be added here? *update *<-- I see you have now created a ticket for this

src/locales/en/translation.json
```diff
@@ -948,28 +948,19 @@
...
-        ],
-        "subHeading2": "UK passport",
-        "paragraph3": "You can use any UK passport if it has not expired.",
-        "subHeading3": "Non-UK passport with a biometric chip",
-        "paragraph4": "You can use a non-UK passport if it:",
-        "nonUkPassportRequirements": [
-          "has a biometric chip",
-          "has not expired"
+        "paragraph1": "You can use one of the following types of photo ID:",
+        "documentsList": [
```
AmritSidhu: Same as above, this can be reused

src/locales/cy/translation.json
DanCorderIPV: I've replaced all the `'` with `’` where I've been able to find an existing example in this file. I'll leave the rest for that ticket

src/locales/cy/translation.json
DanCorderIPV: If we could use the component here I'd agree, but I think if we want to re-use the translations here we should pull them out into a component and use that rather than use the same translation text in two separate places.

src/locales/cy/translation.json
DanCorderIPV: Also this text isn't identical :( there's `(FWP)` at the end that doesn't appear in the existing component - probably an oversight by UCD

DanCorderIPV: I have raised this with UCD, they don't have an immediate answer though

DanCorderIPV: UCD have said that we can add `(FWP)` everywhere

src/locales/cy/translation.json
DanCorderIPV: I've changed my mind on this one and just commonised the text as you originally suggested

AmritSidhu (approved):

DanCorderIPV:
@blakeyp @shivanshuit914 is this OK to merge now?

blakeyp:
@DanCorderIPV are we missing some translations for the typesOfPhotoId page? we'll need those before we can release

DanCorderIPV:
> @DanCorderIPV are we missing some translations for the typesOfPhotoId page? we'll need those before we can release

`typesOfPhotoId` is a component used on the `no-photo-id-xxxx` pages so I don't think we need it for this ticket? It shares some text with the updated `page-ipv-identity-document-start` page so I put that text into a shared part of the translation file.

blakeyp:
@DanCorderIPV ah I see, thanks for clarifying. Good to go then 🚀

blakeyp (approved):

PR #1542, opened 2024-08-19
PYIC-6156: Updated handleJourneyPage/Action session validation

AmritSidhu:
### What changed

- Seperate the session validation and page logic into a new funcition. This is now used in both the GET (handleJourneyPage) and POST (handleJourneyAction) handlers.
- Remove duplicated test
- Use correct page template ID's in test cases now that the POST (handleJourneyAction) function also checks if page is valid
- Add session.currentPage to test cases as the POST (handleJourneyAction) function also checks if session.currentPage is valid

### Why did it change

The GET validation is not the same as for POST. So people can/ can't get onto a page with different conditionality to whether they can post to it

### Issue tracking

- [PYIC-6156](https://govukverify.atlassian.net/browse/PYIC-6156)

[PYIC-6156]: https://govukverify.atlassian.net/browse/PYIC-6156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -176,16 +176,6 @@ function tryValidateClientResponse(client) {
   return true;
 }
 
-function checkForIpvSessionId(req, res) {
-  if (!req.session?.ipvSessionId) {
-    const err = new Error("req.ipvSessionId is missing");
-    err.status = HTTP_STATUS_CODES.UNAUTHORIZED;
-    logError(req, err);
-
-    return renderTechnicalError(req, res);
-  }
-}
-
 function checkForIpvAndOauthSessionId(req, res) {
   if (!req.session?.ipvSessionId && !req.session?.clientOauthSessionId) {
     const err = new Error(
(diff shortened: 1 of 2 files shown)


src/app/ipv/middleware.js
```diff
@@ -330,6 +320,51 @@ async function renderAttemptRecoveryPage(req, res) {
   });
 }
 
+async function validateSessionAndPage(req, res, pageId) {
+  // Check if the page is valid
+  if (!isValidIpvPage(pageId)) {
+    res.status(HTTP_STATUS_CODES.NOT_FOUND);
+    return res.render(getTemplatePath("errors", "page-not-found"));
```
DanCorderIPV: We now have the `render404()` function

src/app/ipv/middleware.js
```diff
@@ -406,21 +423,13 @@ async function handleJourneyPage(req, res, next, pageErrorState = undefined) {
...
-    PAGES.PYI_CRI_ESCAPE,
-    PAGES.PYI_SUGGEST_OTHER_OPTIONS,
-  ];
+  const pageId = req.params.pageId;
 
   try {
-    if (pagesNotUsingOAuthSessionId.includes(currentPageId)) {
-      checkForIpvSessionId(req, res);
-    } else {
-      checkForIpvAndOauthSessionId(req, res);
```
DanCorderIPV, line 422: Is it safe to drop the check for the OAuth session ID?

AmritSidhu: I believe the only scenario we have a `clientOauthSessionId` is for the page `pyi-timeout-recoverable` (the mobile cross-browser scenario). I moved the check to `validateSessionAndPage` function which checks specifically for that page as well as the ID

src/app/ipv/middleware.js
```diff
@@ -330,6 +320,50 @@ async function renderAttemptRecoveryPage(req, res) {
...
+  }
+
+  // Handle the unrecoverable timeout page
+  if (pageId === PAGES.PYI_TIMEOUT_UNRECOVERABLE) {
+    req.session.currentPage = PAGES.PYI_TIMEOUT_UNRECOVERABLE;
+    return res.render(getIpvPageTemplatePath(req.session.currentPage));
+  }
+
+  if (req.session.currentPage !== pageId) {
+    await handleUnexpectedPage(req, res, pageId);
```
blakeyp, line 363: Hmm can we just `return await` this rather than this function returning true/false?

src/app/ipv/middleware.js
```diff
@@ -350,27 +384,8 @@ async function handleJourneyPage(req, res, next, pageErrorState = undefined) {
...
-      );
-
-      return renderTechnicalError(req, res);
-    } else if (pageId === PAGES.PYI_TIMEOUT_UNRECOVERABLE) {
-      req.session.currentPage = PAGES.PYI_TIMEOUT_UNRECOVERABLE;
-      return res.render(getIpvPageTemplatePath(req.session.currentPage));
-    } else if (req.session.currentPage !== pageId) {
-      return await handleUnexpectedPage(req, res, pageId);
+    if (!(await validateSessionAndPage(req, res, pageId))) {
+      return;
```
blakeyp, line 392: and just call `await validateSessionAndPage(req, res, pageId)` here?

DanCorderIPV: Won't we end up calling `res.render()` twice? Do we know if that's safe? A quick look at the docs didn't turn anything up.

src/app/ipv/middleware.test.js
```diff
@@ -876,6 +939,76 @@ describe("journey middleware", () => {
     },
   );
 
+  context(
+    "handleJourneyAction: should render timeout-recoverable page if present clientOauthSessionId",
+    () => {
+      it("should render the technical unrecoverable page", async function () {
```
blakeyp: *timeout-recoverable

blakeyp (approved):

PR #1547, opened 2024-08-20
IPS-954 async session

Joe-Edwards-GDS:
### What changed

Generate the session id asynchronously in a separate middleware

### Why did it change

Trying to eliminate blocking operations

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -4,10 +4,10 @@ const path = require("path");
 const session = require("express-session");
 const { DynamoDBClient } = require("@aws-sdk/client-dynamodb");
 const DynamoDBStore = require("connect-dynamodb")(session);
+const uid = require("uid-safe");
 const {
   frontendVitalSignsInit,
 } = require("@govuk-one-login/frontend-vital-signs");
-
 
 const {
   PORT,
@@ -15,6 +15,7 @@ const {
   SESSION_TABLE_NAME,
   ENABLE_PREVIEW,
   LANGUAGE_TOGGLE_ENABLED,
+  SESSION_COOKIE_NAME,
 } = require("./lib/config");
 
 const { setLocals } = require("./lib/locals");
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -93,21 +95,34 @@ i18next
 
 app.use(i18nextMiddleware.handle(i18next));
 
+// Generate a new session ID asynchronously if needed
+// `express-session` does not support sync session ID generation
+// https://github.com/expressjs/session/issues/107
+app.use(async (req, res, next) => {
+  const sessionId = getCookie(req, SESSION_COOKIE_NAME, SESSION_SECRET);
+  if (!sessionId) {
+    req.generatedSessionId = await uid(24);
```
Joe-Edwards-GDS, line 104: The default behaviour for `express-session` calls `uid.sync` which uses blocking crypto; this version is async.

src/app.js
```diff
@@ -93,21 +95,34 @@ i18next
...
     cookie: {
-      name: "ipv_core_service_session",
+      name: SESSION_COOKIE_NAME,
       expires: false,
       secret: SESSION_SECRET,
       signed: true,
       secure: "auto",
     },
+    // Use the generated session ID, or fall back to the default behaviour
+    genid: (req) => req.generatedSessionId || uid.sync(24),
```
Joe-Edwards-GDS, line 125: Still need the fallback if e.g. the existing session id is invalid, but this should be infrequent

vam-gds (approved):

src/app.js
```diff
@@ -93,21 +94,39 @@ i18next
 
 app.use(i18nextMiddleware.handle(i18next));
 
+app.use(cookieParser());
+
+// Generate a new session ID asynchronously if no session cookie
+// `express-session` does not support sync session ID generation
```
blakeyp, line 100: *async ?

blakeyp (approved):

PR #1556, opened 2024-08-22
BAU Avoid blocking crypto

Joe-Edwards-GDS:
### What changed

- Change nonce generation to use async randomBytes
- Change request id generation to use non-blocking Math.random
- Change session id generation to use async uid (where possible)

### Why did it change

Blocking operations cause a significant degradation in performance when under high load. Making these operations async means the work will be offloaded to the nodejs threadpool and allow other operations to continue.

In the perf environment, when under load:
- Without this change: healthchecks degrade as the task came under load and end up frequently timing out
- After this change: healthchecks consistently return quickly (mostly <10ms and all <50ms)

### Issue tracking

- [IPS-954](https://govukverify.atlassian.net/browse/IPS-954)

[IPS-954]: https://govukverify.atlassian.net/browse/IPS-954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -4,6 +4,7 @@ const path = require("path");
 const session = require("express-session");
 const { DynamoDBClient } = require("@aws-sdk/client-dynamodb");
 const DynamoDBStore = require("connect-dynamodb")(session);
+const uid = require("uid-safe");
 
 const {
   PORT,
@@ -11,6 +12,7 @@ const {
   SESSION_TABLE_NAME,
   ENABLE_PREVIEW,
   LANGUAGE_TOGGLE_ENABLED,
+  SESSION_COOKIE_NAME,
 } = require("./lib/config");
 
 const { setLocals } = require("./lib/locals");
@@ -89,20 +91,38 @@ i18next
 
 app.use(i18nextMiddleware.handle(i18next));
 
+app.use(cookieParser());
+
+// Generate a new session ID asynchronously if no session cookie
+// `express-session` does not support async session ID generation
+// https://github.com/expressjs/session/issues/107
+app.use(async (req, res, next) => {
+  if (!req.cookies[SESSION_COOKIE_NAME]) {
+    req.generatedSessionId = await uid(24);
+  }
+  next();
+});
+
 app.use(
   session({
-    name: "ipv_core_service_session",
+    name: SESSION_COOKIE_NAME,
     store: sessionStore,
     saveUninitialized: false,
     secret: SESSION_SECRET,
     unset: "destroy",
     resave: false,
     cookie: {
-      name: "ipv_core_service_session",
+      name: SESSION_COOKIE_NAME,
       expires: false,
       secret: SESSION_SECRET,
       signed: true,
       secure: "auto",
+    },
+    // Use the newly generated session ID, or fall back to the default behaviour
+    genid: (req) => {
+      const sessionId = req.generatedSessionId || uid.sync(24);
+      delete req.generatedSessionId;
+      return sessionId;
     },
   }),
 );
@@ -128,8 +148,6 @@ app.use((req, res, next) => {
   });
   next();
 });
-
-app.use(cookieParser());
 
 app.use((req, res, next) => {
   res.set(
--- a/src/app.js
+++ b/src/app.js
@@ -32,6 +32,7 @@ const {
 const { pageNotFoundHandler } = require("./handlers/page-not-found-handler");
 const {
   securityHeadersHandler,
+  cspHandler,
 } = require("./handlers/security-headers-handler");
 
 const APP_VIEWS = [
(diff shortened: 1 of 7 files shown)


src/lib/locals.js
```diff
@@ -13,10 +13,10 @@ const {
 const { generateNonce } = require("./strings");
 
 module.exports = {
-  setLocals: function (req, res, next) {
+  setLocals: async function (req, res, next) {
     res.locals.uaContainerId = GTM_ID;
     res.locals.ga4ContainerId = GTM_ID_GA4;
-    res.locals.cspNonce = generateNonce();
+    res.locals.cspNonce = await generateNonce();
```
Wynndow, line 19: Time to expose my poor async skills...

If we're awaiting this method, won't execution block until it returns?

Wynndow: OK, so reading up, await within an async method will block, but the execution from where the method is called will continue.

In that case, how can we be sure that the nonce will be generated by the time the response is generated?

Joe-Edwards-GDS: It's actually a bit cleverer than that - `async`/`await` is (approximately) just syntactic sugar for `Promise.then`.

When you hit an `await` it suspends execution of that event - this is not the same as blocking, because it allows the event loop to pick up other business (i.e. other requests). Once the promise completes, it will queue another task to resume execution of the event from the point that it was awaiting, which will be picked up when the event loop is free.

It might be a bit more obvious what's happening if we rewrote this to use 'plain' promises:
```
setLocals: function (req, res, next) {
    res.locals.uaContainerId = GTM_ID;
    res.locals.ga4ContainerId = GTM_ID_GA4;
    (generateNonce()).then((nonce) => {
        res.locals.cspNonce = nonce;
        res.locals.isGa4Disabled = GA4_DISABLED;
        res.locals.isUaDisabled = UA_DISABLED;
        res.locals.dynatraceRumUrl = DT_RUM_URL;
        ...
        next();
    });
```
This makes it clear that everything after the `await` is actually run in a separate context (including `next()` - which is what tells express to continue processing the request)

Wynndow: OK - so the flow of execution for that particular request wouldn't continue on to the next middleware until the nonce has been generated? The other business that the event loop would pick up would be whole new requests coming in?

Joe-Edwards-GDS: Contrast this to async code in Java:
```
CompleteableFuture<Object> future = dispatchSomeAsyncTask();

Object value = future.get();
```
The `get()` call here _is_ blocking - the current thread will block until the result of the future is available. The upside is that in Java you can have multiple threads, so blocking isn't quite as fatal as it is in JS

Joe-Edwards-GDS: > OK - so the flow of execution for that particular request wouldn't continue on to the next middleware until the nonce has been generated? The other business that the event loop would pick up would be whole new requests coming in?

Yeah exactly

Wynndow: I'm assuming operations like this are reasonably quick to actually complete given the need te happen during the request/response cycle - I'm now wondering why making them async has made such a large perf difference. Maybe here isn't the place to have that discussion though!

src/app.js
```diff
@@ -87,21 +91,39 @@ i18next
...
-      name: "ipv_core_service_session",
+      name: SESSION_COOKIE_NAME,
       expires: false,
       secret: SESSION_SECRET,
       signed: true,
       secure: "auto",
     },
+    // Use the newly generated session ID, or fall back to the default behaviour
+    genid: (req) => {
+      const sessionId = req.generatedSessionId || uid.sync(24);
```
Wynndow, line 123: When would we need to fallback?

src/handlers/security-headers-handler.js
```diff
@@ -2,16 +2,21 @@ module.exports = {
...
     res.set({
       "Strict-Transport-Security": "max-age=31536000",
+      "X-Frame-Options": "DENY",
+      "X-XSS-Protection": "0",
+      "X-Content-Type-Options": "nosniff",
+    });
+    res.removeHeader("X-Powered-By");
+    next();
+  },
+  cspHandler: (req, res, next) => {
```
Wynndow, line 12: Why split this out?

src/app.js
Joe-Edwards-GDS: If the user has a session cookie but it's invalid for some reason (e.g. it's an old session, or they've just messed with the cookie themselves), then we'll still need to generate a new session.

We could try and work that all out above, but it doesn't seem worth it.

src/handlers/security-headers-handler.js
Joe-Edwards-GDS: It's really a consequence of moving the `setLocals` call below static assets (so we don't generate a nonce for static assets).

We can still put the other security headers on static assets, but not the CSP.

src/lib/locals.js
Joe-Edwards-GDS: I'll message you

Wynndow (approved):

PR #1564, opened 2024-08-30
BAU Stop duplicate logging

blakeyp:
### What changed

Removes the standalone healthcheck router and registers the healthcheck endpoint on the main router so we get (non-duplicated) logging across all routes

### Why did it change

Earlier this month we registered the logging middleware on the healthcheck router to enable logging for the healthcheck, but as this router is ultimately registered on the `app` alongside the main router it means requests fall through 2 instances of logging middleware and we see double

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -178,13 +178,10 @@ if (ENABLE_PREVIEW) {
   router.use("/dev", require("./app/development/router"));
 }
 
-const healthcheckRouter = express.Router();
-healthcheckRouter.use(loggerMiddleware);
-healthcheckRouter.get("/healthcheck", (req, res) => {
+router.get("/healthcheck", (req, res) => {
   return res.status(200).send("OK");
 });
 
-app.use(healthcheckRouter);
 app.use(router);
 
 app.use(journeyEventErrorHandler);
--- a/src/app.js
+++ b/src/app.js
@@ -148,6 +148,7 @@ app.use((req, res, next) => {
   });
   next();
 });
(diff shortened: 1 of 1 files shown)


andyloughran:
If we do this, then healthcheck inherits all teh middlewares above it - making it too fat, so we need to move healtcheck up the app.js file.

src/app.js
Joe-Edwards-GDS, line 1: This feels a bit weird - do we misunderstand how middlewares bound to routers work?

As Andy says, the reason we have a separate router is so that we _don't_ get all the same middlewares as on the main app routes, and I'm pretty sure that must have been working for the main logger, since we didn't get healthcheck logs until we added it to the healthcheck router...

Worth noting that the logger middleware does produce two logs per request - one for req start, and one for req end, is that what we're seeing?

Joe-Edwards-GDS, line 1: @andyloughran regardless of this, it's worth noting that the healthcheck logs are now very noisy (especially with an increased rate of them) - more than our 'real' logs a lot of the time.

Easy enough to filter out in queries, but we may want to consider dropping once we understand how they work.

src/app.js
```diff
@@ -78,6 +78,7 @@ app.use(
 
 app.use(setLocals);
 app.use(cspHandler);
+app.use(loggerMiddleware);
```
Joe-Edwards-GDS: I think we include the session ID in the logs - do we need this to come after the session configuration?

Joe-Edwards-GDS (approved):

PR #1571, opened 2024-09-10
Pyic 5881 browser functional testing

DanCorderIPV:
### What changed

Add initial browser tests

### Why did it change

We want more reliable and faster tests

### Issue tracking

- [PYIC-5881](https://govukverify.atlassian.net/browse/PYIC-5881)

[PYIC-5881]: https://govukverify.atlassian.net/browse/PYIC-5881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


browser-tests/functional-tests/example.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  // Start a session
+  await page.goto(getAuthoriseUrlForJourney("exampleContext"));
+
+  // Go to the DCMAW CRI
+  await page.click("input[type='radio'][value='appTriage']");
+  await page.click("button[id='submitButton']");
+
+  // When we come back from DCMAW with access_denied, core back eventually sends us to page-multiple-doc-check
+  const url = page.url();
+  expect(url).toBe("http://localhost:4601/ipv/page/page-multiple-doc-check");
```
MikeCollingwood: Minor but may we in-line the check of the `url` variables, would prefer to avoid naming like `url2` seen later on

browser-tests/functional-tests/example.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    "Prove your identity at a Post Office with one of the following types of photo ID",
+  );
+});
+
+test("Example welsh language", async ({ page }) => {
+  // Start a session
+  await page.goto(getAuthoriseUrlForJourney("exampleWelshLanguage"));
+
+  // Click the language toggle
+  await page.click('[rel="alternate"]');
```
MikeCollingwood: Could we tack `[hreflang="cy"]` onto the end of this locator to make it clearer?

browser-tests/functional-tests/example.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  await page.click("input[type='radio'][value='ukPassport']");
+  await page.click("button[id='submitButton']");
+
+  // When we come back from web passport with access_denied, core back eventually sends us to prove-identity-another-type-photo-id with context 'passport'
+  const url2 = page.url();
+  expect(url2).toBe("http://localhost:4601/ipv/page/prove-identity-another-type-photo-id");
+
+  const contextSpecificTextLocator = await page.getByText("Use your UK photocard driving licence");
+  const textCount = await contextSpecificTextLocator.count();
+  expect(textCount).toBe(1);
```
MikeCollingwood: Minor but could replace the use of a count with
```javascript
expect(contextSpecificTextLocator).toBeVisible();
```

browser-tests/imposter/config/core-back-api.yaml
```diff
@@ -0,0 +1,156 @@
...
+            application/json:
+              schema:
+                type: "object"
+        500:
+          description: 500 response
+          content: {}
+
+  /healthcheck:
+    get:
+      description: "returns a 200 for Route53 health checks to use"
```
MikeCollingwood, line 66: Ah I see Route53 is an AWS-based service, can we remove this healthcheck?

browser-tests/compose.yaml
```diff
@@ -0,0 +1,56 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
+    depends_on:
+      core-front-browser-test:
+        condition: service_healthy
+    environment:
+      WEBSITE_HOST: http://localhost:4601
+    volumes:
+      - ./snapshot.spec.js-snapshots:/snapshot.spec.js-snapshots
```
MikeCollingwood: Are we taking snapshots? If not, we may also want to remove the `--update-snapshots` tag in the package.json script

browser-tests/compose.yaml
```diff
@@ -0,0 +1,56 @@
+services:
+  core-back-imposter:
+    container_name: core-back-imposter
+    build:
+      context: .
+      dockerfile: imposter/Dockerfile
+    ports:
+      - "4602:8080"
+
+  core-front-browser-test:
```
MikeCollingwood: Could this be renamed so it's clear it's just spinning up core-front, and not doing any testing on it's own. Maybe `core-front` to be simple?

browser-tests/imposter/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM outofcoffee/imposter:3.44.0 AS imposter
+
+FROM eclipse-temurin:11.0.22_7-jre-jammy
+COPY --from=imposter /opt/imposter/lib /opt/imposter/lib
+
+# your custom config
+COPY imposter/config /opt/imposter/config
+
+# optional docker healthcheck
```
MikeCollingwood: Is this optional, in the `core-front-browser-test` it depends on the container for this being healthy so I assume this is needed, unless I'm confusing this with the healthcheck from `core-back-api.yaml`
```
      core-front-browser-test:
        condition: service_healthy
```

I think the comment above this one could be removed

browser-tests/functional-tests/example.spec.js
```diff
@@ -0,0 +1,63 @@
...
+});
+
+// Put the journey into the 'state' parameter of the request so that imposter sends it back as the IpvSessionId
```
MikeCollingwood: Maybe link to the readme here as it's an important thing to understand. At least, I spent a bit of time not knowing the full idea behind this method

browser-tests/functional-tests/example.spec.js
```diff
@@ -0,0 +1,63 @@
+const { test, expect } = require("@playwright/test");
+
+test("Example page navigation", async ({ page }) => {
```
blakeyp: can these be 'real' rather than example tests? the intention is just to have one set of page rendering scenarios like these, rather than building out scenarios for all the pages/journeys

browser-tests/compose.yaml
DanCorderIPV: Good point, this is a hangover from the spike code

browser-tests/imposter/config/core-back-api.yaml
DanCorderIPV: https://aws.amazon.com/route53/

This file is a combination of the core back open API files that define the core back endpoints.
It's safest/most realistic if we just copy all the endpoints over even if we don't use them all

browser-tests/functional-tests/example.spec.js
DanCorderIPV: Fixed by test refactor

browser-tests/functional-tests/example.spec.js
DanCorderIPV: I tried that multiple times before and it wouldn't work. Just tried it again and it does... Maybe something else was going wrong at the same time.

browser-tests/functional-tests/example.spec.js
DanCorderIPV: Nice :)

browser-tests/compose.yaml
DanCorderIPV: I didn't want to just call it core-front so it wouldn't get confused with the local running core front. Will try to think of something else.

browser-tests/imposter/Dockerfile
DanCorderIPV: True, this file was copied wholesale from an imposter example file

blakeyp (approved):

PR #1574, opened 2024-09-12
Snapshot tests

DanCorderIPV:
### What changed

Add snapshot tests

### Why did it change

So that we can easily see if we change the core-front UI


browser-tests/snapshot-tests/snapshot.spec.js
Joe-Edwards-GDS, line 1: One thing worth bearing in mind - this snapshots directory is 20MB, and if we end up having a lot of churn on the screenshots, that could quickly balloon the size of the repository. (E.g. any change to the layout will add an extra 20MB)

If the changes stay isolated to particular pages then it should be ok. If it proves to be a problem then we could investigate other solutions - e.g. Git LFS - but any existing screenshots will be stuck in the history.

browser-tests/snapshot-tests/snapshot.spec.js
```diff
@@ -0,0 +1,104 @@
+const { test, expect } = require("@playwright/test");
+const path = require("path");
+const fs = require("fs");
+
+test("Snapshots", async ({ page }) => {
+
+  test.setTimeout(120000)
+  for (const pageName of Object.keys(pagesToTest)) {
+    const contexts = pagesToTest[pageName];
+    const contextsToTest = contexts.length > 0 ? contexts : [ undefined ];
```
MikeCollingwood, line 10: Could this always append `undefined` so we don't need to add it to the pagesToTest contexts?

```
const contextsToTest = [...contexts, undefined];
```

browser-tests/snapshot-tests/snapshot.spec.js
```diff
@@ -0,0 +1,104 @@
...
+      if (!testedTemplates.find(tt => tt === templateName)) {
+        missingTemplates.push(templateName);
+      }
+    }
+
+    expect(missingTemplates).toStrictEqual([]);
+  });
+});
+
+const pagesToTest = {
```
MikeCollingwood, line 51: Could we spread over all of the views defined, then you can explicitly redefine ones which want contexts? This would eliminate the need for the above test

```javascript
const pagesToTest = {
  ...Object.fromEntries(
    fs.readdirSync("/../../src/views/ipv/page")
      .map(file => [parse(file).name, []])
  ),
  "no-photo-id-security-questions-find-another-way": ["dropout", undefined],
  // ...
  "pyi-triage-desktop-download-app": ["android", "iphone"],
  "update-details-failed": ["repeatFraudCheck",undefined],
  "update-name-date-birth": ["repeatFraudCheck","reuse","rfcAccountDeletion",undefined]
}
```

browser-tests/snapshot-tests/snapshot.spec.js
```diff
@@ -0,0 +1,104 @@
...
+test("Snapshots", async ({ page }) => {
+
+  test.setTimeout(120000)
+  for (const pageName of Object.keys(pagesToTest)) {
+    const contexts = pagesToTest[pageName];
+    const contextsToTest = contexts.length > 0 ? contexts : [ undefined ];
+
+    for (const context of contextsToTest) {
+      for (const language of [ "en", "cy" ]) {
+        var url = `http://localhost:4601/dev/template/${pageName}/${language}`;
```
MikeCollingwood, line 14: Is the base of this URL able to be taken from config?

browser-tests/snapshot-tests/snapshot.spec.js
MikeCollingwood: We could reduce the memory demand by hashing the image data? You wouldn't see the old screenshot but would be required to investigate changes - and we have UCD as a source of truth on what they should look like.

```javascript
        const imageData = await page.screenshot({ fullPage: true }).toString('base64');
        const hash = crypto.createHash('sha256').update(imageData).digest('hex');

        await expect(hash).toMatchSnapshot(`${pageName}-${language}-hash.txt`);
```

https://playwright.dev/docs/test-snapshots#non-image-snapshots

Joe-Edwards-GDS: That would mean the screenshot needs to be byte-for-byte identical - I think if you use images then playwright will allow a certain tolerance (configurably), so any small jitter in smoothing or compression doesn't cause a false positive.

(It'd also make it much harder to figure out whether a change is correct, if you can't see the previous version!)

browser-tests/snapshot-tests/snapshot.spec.js
DanCorderIPV: One of the points of the test is to make sure people think about adding pages to the snapshots - that way they won't forget to include contexts (which we have no way of automatically finding)

browser-tests/snapshot-tests/snapshot.spec.js
DanCorderIPV: There are pages that are only ever displayed with contexts so we don't want `undefined` on them

browser-tests/snapshot-tests/snapshot.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  "pyi-triage-mobile-confirm": [],
+  "pyi-triage-mobile-download-app": ["android", "iphone"],
+  "pyi-triage-select-device": [],
+  "pyi-triage-select-smartphone": [],
+  "reprove-identity-start": [],
+  "sorry-could-not-confirm-details": ["deleteDetailsReuse",undefined],
+  "update-details-failed": ["repeatFraudCheck",undefined],
+  "update-details": [],
+  "update-name-date-birth": ["repeatFraudCheck","reuse","rfcAccountDeletion",undefined]
+}
```
blakeyp, line 105: This isn't ideal but probably the only way we can test with contexts - I guess at least we can pull this out and reuse it for [PYIC-7279](https://govukverify.atlassian.net/browse/PYIC-7279) if there's no other way

MikeCollingwood: And the accessibility tests for pages I'm working on [PYIC-7309](https://govukverify.atlassian.net/browse/PYIC-7309)

browser-tests/snapshot-tests/snapshot.spec.js
DanCorderIPV: I don't think so easily, we'd presumably want to take it from the docker compose file

browser-tests/snapshot-tests/snapshot.spec.js
MikeCollingwood: Another option could be to use jpeg ([docs](https://playwright.dev/docs/api/class-page#page-screenshot-option-type)), which is an option with playwright. With it, you can also reduce the screenshot quality but using jpeg should reduce the image size significantly.

DanCorderIPV: So the jpeg option doesn't work with `toHaveScreenshot()` but we can make our own equivalent functionality fairly easily.
Turning jpeg quality down to 20 (out of 100) still looks fine for checking the text etc and drops the image size by just over 50%
I have talked to a Playwright person on their discord and they say that they normally recommend using Git LFS if file size becomes an issue.

Joe-Edwards-GDS: If you do make this change, make sure you rebase and squash the commits containing the screenshots before merging (to remove the pngs from the history) :)

MikeCollingwood:
Could this PR also update the `pre-merge-checks.yml` to include a step for uploading the artifacts on failure. It could help devs see the difference when it's run in GHA

```yaml
    - name: Upload test artifacts
      if: failure()
      uses: actions/upload-artifact@v3
      with:
        name: test-results
        path: ./browser-tests/test-results/
```

DanCorderIPV:
There appear to be some quite large differences between the jpeg images on my laptop and on github (over 8000 different pixels in one image) so I'm not comfortable with this approach and will revert to PNGs.
If we want to keep the repo size down we'll probably need to use Git LFS.

DanCorderIPV:
> Could this PR also update the `pre-merge-checks.yml` to include a step for uploading the artifacts on failure. It could help devs see the difference when it's run in GHA
> 
> ```yaml
>     - name: Upload test artifacts
>       if: failure()
>       uses: actions/upload-artifact@v3
>       with:
>         name: test-results
>         path: ./browser-tests/test-results/
> ```

Actually I won't do this right now as I don't want to spend too long on this and as we're running in a docker container we'd need to extract the test results from there (possibly with another mounted volume)

DanCorderIPV:
Going to retry the jpg changes as there was a real UI change on main that conflicted with the tests

DanCorderIPV:
Ok, jpegs are working and the files are about 50% of the size they were before.
@blakeyp if you are happy with this getting merged let me know and I'll squash this branch and force push it (or create a new branch) so we don't start with loads of image file churn in the repo.

browser-tests/snapshot-tests/snapshot.spec.js-snapshots/sorry-could-not-confirm-details-cy-deleteDetailsReuse-chromium-linux.png
MikeCollingwood, line 1: There are a few PNG files here. They might've slipped through the net so thought I'd flag

DanCorderIPV: I'll remove them in the squash

MikeCollingwood (approved):

DanCorderIPV:
Force pushed to squash all the image file churn

PR #1575, opened 2024-09-13
PYIC-7309: Add accessibility check for template pages

MikeCollingwood:
### What changed

- Add accessibility check for template pages
- Make browser tests run in parallel for performance

### Why did it change

- To build out our core-front tests to include accessibility testing

### Issue tracking

- [PYIC-7309](https://govukverify.atlassian.net/browse/PYIC-7309)

[PYIC-7309]: https://govukverify.atlassian.net/browse/PYIC-7309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


browser-tests/functional-tests/accessibility.spec.js
```diff
@@ -0,0 +1,32 @@
+const { test, expect } = require("@playwright/test");
+const { iteratePagesAndContexts } = require("../data/pagesAndContexts.js");
+const AxeBuilder = require("@axe-core/playwright");
+
+test.describe.parallel("Accessibility tests", () => {
+  test.setTimeout(120000);
+  iteratePagesAndContexts(
+    (pageName, context, language, url) => {
+      test(`Accessibility check for ${pageName}, context ${context} and language ${language}`, async ({ page }) => {
+        await page.goto(url);
```
blakeyp, line 12: Some of our pages have a details component that we should probably open before running the scan if we can. See https://playwright.dev/docs/accessibility-testing#configuring-axe-to-scan-a-specific-part-of-a-page
(is it possible to just try opening a details component if it's there?)

browser-tests/functional-tests/accessibility.spec.js
```diff
@@ -0,0 +1,32 @@
...
+        await assertNoAccessibilityViolations(page);
+      });
+    }
+  );
+
+  (["radio", "checkbox"]).forEach((errorState) => {
+    ["en", "cy"].forEach((language) => {
+      test(`Accessibility check for ${errorState} error state in ${language}`, async ({ page }) => {
+        await page.goto(`http://localhost:4601/dev/template/confirm-your-details/${language}?errorState=${errorState}`);
```
blakeyp: Probably best to get the domain for this from config or at least define it in one place somewhere?

blakeyp: Also what does setting the error state actually do on this specific template 🤔

browser-tests/functional-tests/accessibility.spec.js
MikeCollingwood: I will look at implementing this but maybe just in a test for a single page, like below for the error state? Testing for other pages is unlikely to increase coverage meaningfully and this way the tests run quicker

browser-tests/functional-tests/accessibility.spec.js
MikeCollingwood: I thought that myself and commented the same thing on the Snapshot tests PR. This was the response: [thread](https://github.com/govuk-one-login/ipv-core-front/pull/1574#discussion_r1758825390), so I've gone to keep it the same here. If you think it's still worth it I can see what I can do.

On your second point, I wrote a bit on it in the [README for pages on creating pages](https://github.com/govuk-one-login/ipv-core-front/blob/main/src/views/README.md#:~:text=pageErrorState,/%20radiobox). Our error handling has redundancy - we have _two_ objects which lead to error states, but this is the only one where can specify whether it is the radio or checkbox which is failing. It causes either the radio buttons or the checkboxes to show validation errors

browser-tests/functional-tests/accessibility.spec.js
blakeyp: Fine to leave it if it's a faff

browser-tests/functional-tests/accessibility.spec.js
blakeyp: Can't we pull it from the env vars defined in the docker compose file?

browser-tests/functional-tests/accessibility.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        await assertNoAccessibilityViolations(page);
+      });
+    })
+  });
+
+  ["en", "cy"].forEach((language) => {
+    test(`Accessibility check for open details in ${language}`, async ({ page }) => {
+      await page.goto(`http://localhost:4601/dev/template/pyi-triage-select-smartphone/${language}`);
+      await page.click(".govuk-details__summary");
```
blakeyp: The docs mention something about using `.waitFor()` - do we need to do that?
>   // It is important to waitFor() the page to be in the desired
  // state *before* running analyze(). Otherwise, axe might not
  // find all the elements your test expects it to scan.

MikeCollingwood: The tests passed without this but I will include just in case!

browser-tests/test-results/.last-run.json
```diff
@@ -0,0 +1,4 @@
+{
```
MikeCollingwood, line 1: Will remove.

browser-tests/functional-tests/accessibility.spec.js
```diff
@@ -0,0 +1,45 @@
...
+        await page.goto(url);
+
+        // Open all summaries on the page
+        const allSummaries = await page.locator("summary").all();
+        for (const summary of allSummaries) {
+          await summary.click();
+        }
+        const allSummaryDetails = await page.locator(".govuk-details__text").all();
+        for (const details of allSummaryDetails) {
+          await details.waitFor();
```
blakeyp, line 21: What's the difference between `click()` and `waitFor()` ?

MikeCollingwood: The click is the action but in the second case we're waiting for the summaries to have opened, just to be sure they have

blakeyp: Ah I see, thanks

blakeyp (approved):

PR #1584, opened 2024-10-01
Pyic 6736 strategic app journey map

DanCorderIPV:
### What changed

Adding ability to handle `identify-device` page response from core-back

### Why did it change

We will need this feature to deal with CoI journeys when using the strategic app.

Overall these changes will need to be released in 4 sections, starting with this one, then alternating core back and front.
I've run the end to end tests with this and the current head of core-back to check that it doesn't break anything.

### Issue tracking

- [PYIC-6736](https://govukverify.atlassian.net/browse/PYIC-6736)

[PYIC-6736]: https://govukverify.atlassian.net/browse/PYIC-6736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -42,7 +42,10 @@ const {
 } = require("../../lib/paths");
 const PAGES = require("../../constants/ipv-pages");
 const { parseContextAsPhoneType } = require("../shared/contextHelper");
-const { sniffPhoneType, detectAppTriageEvent} = require("../shared/deviceSniffingHelper");
+const {
+  sniffPhoneType,
+  detectAppTriageEvent,
+} = require("../shared/deviceSniffingHelper");
 const ERROR_PAGES = require("../../constants/error-pages");
 
 const directoryPath = path.join(__dirname, "/../../views/ipv/page");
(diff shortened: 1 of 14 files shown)


src/app/ipv/middleware.js
```diff
@@ -144,11 +147,21 @@ async function handleBackendResponse(req, res, backendResponse) {
...
+    await req.session.save(function (err) {
+      if (err) {
+        logError(req, err, "Error saving session");
+        throw err;
+      }
+    });
+
+    // Special case handling for "identify-device". This is used by core-back to signal that we need to
+    // check the user's device and send back the relevant "appTriage" event.
+    if (backendResponse.page === PAGES.IDENTIFY_DEVICE) {
```
blakeyp, line 159: It feels a bit weird this being a page response when it's not actually a page 🤔 
Why is it we need this for the CoI scenarios? I think originally we just always sniffed the device to get the right appTriage event without an additional call to core-back

DanCorderIPV: It is weird and I don't hugely like it, but it looks like our options are:
1. Do this
2. Introduce a new type of response from core-back for just this case
3. Have a huge number of journey map states for handling the name change routes as they go directly from a backend processing step to DCMAW at the moment, so we'd need turn each existing backend step into 3 to then go to the correct appTriage page afterwards.

blakeyp: We agreed to go with this for now, with a tech debt ticket raised to consider adding a new response type - [PYIC-7534](https://govukverify.atlassian.net/browse/PYIC-7534)

blakeyp (approved):

PR #1586, opened 2024-10-04
PYIC-7543: fix session saving before redirect

Joe-Edwards-GDS:
### What changed

The save method is not actually async - it takes a callback, so we can't await it.

Instead, we can just use the method we already have.

### DIFF
--- a/src/app/ipv/middleware.js
+++ b/src/app/ipv/middleware.js
@@ -131,7 +131,7 @@ async function handleBackendResponse(req, res, backendResponse) {
 
     req.session.ipvSessionId = null;
     const { redirectUrl } = backendResponse.client;
-    return res.redirect(redirectUrl);
+    return saveSessionAndRedirect(redirectUrl);
   }
 
   if (backendResponse?.page) {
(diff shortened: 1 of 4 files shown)


src/app/shared/redirectHelper.js
```diff
@@ -1,8 +1,8 @@
 const { logError } = require("./loggerHelper");
 
 module.exports = {
-  saveSessionAndRedirect: async (req, res, redirectUrl) => {
-    await req.session.save(function (err) {
```
Joe-Edwards-GDS, line 5: Typescript would have spotted this immediately!

I'm making good progress on running TS/JS interop, so watch this space...

DanCorderIPV (dismissed):

DanCorderIPV (dismissed):

src/app/ipv/middleware.js
```diff
@@ -131,7 +131,7 @@ async function handleBackendResponse(req, res, backendResponse) {
 
     req.session.ipvSessionId = null;
     const { redirectUrl } = backendResponse.client;
-    return res.redirect(redirectUrl);
+    return saveSessionAndRedirect(req, res, redirectUrl);
```
DanCorderIPV, line 134: Not at all needed right now, but do we need something similar on line 107?

DanCorderIPV (approved):

src/app/ipv/middleware.js
Joe-Edwards-GDS: Yeah potentially - it's even less likely to be an issue when redirecting to a CRI (since they shouldn't immediately bounce back), but would be good to be consistent.

PR #1588, opened 2024-10-09
PYIC-7555 TS supporting modules

Joe-Edwards-GDS:
### What changed

Typescript for the non-middleware modules

Also a couple of small tweaks to support internal dependencies properly.


src/app/development/middleware.ts
```diff
@@ -0,0 +1,82 @@
+import fs from "fs/promises";
```
Joe-Edwards-GDS, line 1: (small change to use async fs methods)

src/config/nunjucks.ts
```diff
@@ -0,0 +1,81 @@
+import addLanguageParam from "@govuk-one-login/frontend-language-toggle";
+import { Application } from "express";
+import i18next from "i18next";
+import nunjucks, { Environment } from "nunjucks";
+import { kebabCaseToPascalCase } from "../app/shared/stringHelper";
+import config from "../lib/config";
+
+interface FilterContext {
```
Joe-Edwards-GDS, line 8: I can't actually work out where this comes from, and whether there's a better type definition for it in one of the libraries... Open to suggestions!

src/lib/config.ts
```diff
@@ -1,16 +1,16 @@
...
-if (!appEnv.isLocal) {
-  serviceConfig.coreBackAPIUrl = appEnv.getServiceURL("core-back-api");
-}
+const appEnv = cfenv.getAppEnv();
+const coreBackApiUrl = appEnv.isLocal
+  ? undefined
+  : appEnv.getServiceURL("core-back-api");
 
-module.exports = {
+export default {
```
Joe-Edwards-GDS, line 11: This is a slight change (and has caused all the JS imports to need updating), but feels easier to manage than exporting each config value individually.

Also means the imports (once using TS) are cleaner - as you can just `import config from "./config"` and then access things on that object.

src/lib/logger.ts
```diff
@@ -41,9 +42,6 @@ const loggerMiddleware = pinoHttp({
   },
   // Set to `false` to prevent standard serializers from being wrapped.
   wrapSerializers: false,
-  autoLogging: {
-    ignorePaths: [],
```
Joe-Edwards-GDS, line 45: `ignorePaths` is deprecated (and no longer appears in the type definitions) - but we weren't using it anyway!

tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "compilerOptions": {
     "allowJs": true,
     "target": "es2022",
-    "module": "nodenext",
+    // For now we specifically compile to commonJS for interop with existing libraries
+    // Once fully transitioned to TS, we can switch to nodenext and output ES modules
+    "module": "commonjs",
```
Joe-Edwards-GDS, line 7: This caused me a bit of a headache...

Everything works fine with `nodenext` except for the govuk-frontend imports - because those define both CJS and ESM versions.

TSC (appropriately) determines that the core-front modules should output as CommonJS, but then tries to load the govuk-frontend ones as ESM. This fails, because the ESM imports can't be replicated in CJS. However, if you ask it to specifically use commonjs here, then it writes out a require anyway, and it all works as before.

(In future we should be able to change this back and turn the whole thing into ESM)

src/app/development/middleware.ts
```diff
@@ -0,0 +1,95 @@
...
+    const directoryPath = path.resolve("views/ipv/page");
+
+    // Load available templates and convert into radio option objects for the GOV.UK Design System nunjucks template
+    if (!config.TEMPLATE_CACHING || !templateRadioOptions) {
+      const templateFiles = await fs.readdir(directoryPath);
+      templateRadioOptions = templateFiles.map((file) => ({
+        text: path.parse(file).name,
+        value: path.parse(file).name,
+      }));
+    }
```
Joe-Edwards-GDS, line 35: Refactored this a little bit so it also respects the template caching config - GitHub scanning still picks up on it, but it should mitigate.

src/lib/config.ts
thebauSoftwire: agreed - nice!

src/handlers/error-handler.test.ts
```diff
@@ -1,20 +1,16 @@
...
-} = require("../../src/handlers/journey-event-error-handler");
-const proxyquire = require("proxyquire");
+import { AxiosError, AxiosInstance, AxiosResponse } from "axios";
+import { expect } from "chai";
+import { NextFunction, Request, Response } from "express";
+import proxyquire from "proxyquire";
+import sinon from "sinon";
+import { pageNotFoundHandler } from "../../src/handlers/page-not-found-handler";
+import { serverErrorHandler } from "../../src/handlers/internal-server-error-handler";
+import { journeyEventErrorHandler } from "../../src/handlers/journey-event-error-handler";
```
thebauSoftwire: instead of importing these one per line, I'm wondering if we could add a `index.ts` to the `src/handlers` folder to hold something like:
```
export {serverErrorHandler} from './internal-server-error-handler.ts
export {journeyEventErrorHandler} from './journey-event-error-handler.ts
export {pageNotFoundHandler} from './page-not-found-handler.ts
```
so that when we import them, we can just do something like
`import {serverErrorHandler, journeyEventErrorHandler, pageNotFoundHandler} from '../../src/handlers`

src/handlers/internal-server-error-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+import axios from "axios";
+import PAGES from "../constants/ipv-pages";
+import { getIpvPageTemplatePath, getErrorPageTemplatePath } from "../lib/paths";
+import ERROR_PAGES from "../constants/error-pages";
+import { ErrorRequestHandler } from "express";
+
+const hasStatus = (input: unknown): input is { status: number } =>
+  !!(input as { status: number }).status;
+
+export const serverErrorHandler: ErrorRequestHandler = (
```
thebauSoftwire: just one for my own learning, is there a reason why these are named exports instead of default exports?

src/handlers/error-handler.test.ts
thebauSoftwire: this is quite small but could be handy for bigger modules in future

src/services/coreBackService.ts
```diff
@@ -0,0 +1,115 @@
...
+export interface CriCallbackRequest {
+  authorizationCode?: string;
+  credentialIssuerId: string;
+  error?: string;
+  errorDescription?: string;
+  redirectUri: string;
+  state?: string;
+}
+
+function generateAxiosConfig(url: string, req: Request): AxiosRequestConfig {
```
thebauSoftwire: another q for me - there's a mix of regular `function` and arrow functions in this repo, is there a preference going forward?

thebauSoftwire (dismissed):
Nice! Mostly questions from me and one suggestion

src/handlers/error-handler.test.ts
Joe-Edwards-GDS: Yeah, if we end up with loads of these then it's a possibility, but I think I prefer the separate exports for now! They work better for tree-shaking/mocking/etc. and for the most part your IDE should handle the imports for you...

Definitely a matter of taste :)

src/handlers/internal-server-error-handler.ts
Joe-Edwards-GDS: Great question - I think they'd probably be better as default exports, so I'll change them!

src/services/coreBackService.ts
Joe-Edwards-GDS: This is another question of taste, and I don't think there's a perfect answer... The semantics are slightly different.

For expressions (e.g. `arr.map((a) => b)` vs. `arr.map(function(a) { return b; })`, I would always use an arrow function, unless I need `this` binding (which is fairly rare these days).

For top-level definitions, it's a bit more debatable - I prefer using const for a couple of reasons:
- Consistency - that way _all_ your functions are arrow functions (unless for a good reason, which signals that there's something funky about the function)
- It's more functional - that way your 'functions' are just variables like everything else

The last point is particularly relevant with Typescript definitions where you want to use a function type:
```
// function declaration - needs to use individual param/return types
function someRequestHandler(req: Request, res: Response, next: NextFunction): Response {
  ...
}

// arrow function variable - typed as a handler
const someRequestHandler: RequestHandler = (req, res, next) => {
  ...
}

// or the awkward middle-ground - which is kinda the worst of both worlds
const someRequestHandler: RequestHandler = function (req, res, next) {
  ...
}
```

I can change this for consistency, and we could consider adding an eslint rule to enforce it!

thebauSoftwire: I didn't know you couldn't use function types with `function`, thanks for the insight :)

thebauSoftwire (approved):

PR #1587, opened 2024-10-07
PYIC-3912 Typescript setup

Joe-Edwards-GDS:
### What changed

Add Typescript configuration so that we can support a mixture of TS and JS files
- Add Typescript and `tsconfig.json`
- Update dev and test commands to run using `ts-node`
- Update build process to build JS from typescript

### Why did it change

Typescript is safer to write, as it provides compile-time typechecking

### Issue tracking

- [PYIC-3912](https://govukverify.atlassian.net/browse/PYIC-3912)

[PYIC-3912]: https://govukverify.atlassian.net/browse/PYIC-3912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1591, opened 2024-10-10
PYIC-3967 restore ts

Joe-Edwards-GDS:
### What changed

Small update to the original PR: https://github.com/govuk-one-login/ipv-core-front/pull/1587

Moves the express-async-errors import so it can correctly patch express.

### Why did it change

The change in import order meant that some routes were not correctly patched, and some errors would cause the node process to crash.


browser-tests/functional-tests/functional.spec.js
```diff
@@ -107,6 +107,33 @@ test.describe("Android tests", () => {
...
+
+    // Go to the error
+    await page.goto("/ipv/journey/page-ipv-identity-document-start/error")
+
+    // When we come back with an error
+    const textLocator = await page.getByText("Sorry, there is a problem");
+    await expect(textLocator).toBeVisible();
+  });
+
+  test("Handles an unexpected error after a CRI callback", async ({ page }) => {
```
Joe-Edwards-GDS, line 123: This is the test that fails without the import change - curiously it only seems to happen when we get an error on the follow-up journey engine request (not the cri callback itself).

blakeyp (approved):

PR #1593, opened 2024-10-10
Pyic 7559 replace csrf handling

DanCorderIPV:
### What changed

Replace CSRF library and update other packages depending on the `cookies` package with a version below 7.0.0

### Why did it change

The old library was deprecated and had a security issue with one of its dependencies

### Issue tracking

- [PYIC-7559](https://govukverify.atlassian.net/browse/PYIC-7559)

[PYIC-7559]: https://govukverify.atlassian.net/browse/PYIC-7559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/credential-issuer/router.js
```diff
@@ -1,12 +1,10 @@
 const express = require("express");
-const csrf = require("csurf");
 
 const router = express.Router();
 
 const { sendParamsToAPI, sendParamsToAPIV2 } = require("./middleware");
-const csrfProtection = csrf({});
 
-router.get("/callback", csrfProtection, sendParamsToAPI);
-router.get("/callback/:criId", csrfProtection, sendParamsToAPIV2);
```
DanCorderIPV: Adding the csrf function to the app means that we don't need to add it to individual routes

src/app/credential-issuer/middleware.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
 
       return handleBackendResponse(req, res, apiResponse?.data);
     } catch (error) {
-      error.csrfToken = req.csrfToken();
+      error.csrfToken = req.csrfToken(true);
```
DanCorderIPV: The `true` here makes the library generate a new token for each request (as the previous library did) instead of one token per user session.

Joe-Edwards-GDS: This is the invocation that I'm suspicious of - why is it only this error handler that does this, and what is going to read this property of the error anyway?

(If it's not doing anything, can we remove it?)

DanCorderIPV: Good point, I'm not sure why we even have CSRF protection here as this is a callback from the CRI so presumably won't have a CSRF token anyway.
This line was added in https://github.com/govuk-one-login/ipv-core-front/pull/634/files by @AmritSidhu , Amrit do you remember why this was needed?
Worth noting that the new version of this function below does not have this CSRF code

DanCorderIPV: I also can't see anywhere that the CSRF property of the error object is used.

package.json
```diff
@@ -78,13 +78,13 @@
     "body-parser": "1.20.3",
     "cfenv": "1.2.4",
     "connect-dynamodb": "3.0.3",
-    "cookie-parser": "1.4.6",
+    "cookie-parser": "1.4.7",
     "copyfiles": "2.4.1",
-    "csurf": "1.11.0",
+    "csrf-sync": "^4.0.3",
```
blakeyp, line 86: Can we pin the version

src/app/credential-issuer/middleware.js
AmritSidhu: I want to say that it was added due to it complaining that it was missing at the time and it required it. But im afraid I really cant remember any specific reason

DanCorderIPV: Turns out that this has already been removed in the move to TypeScript so isn't relevant to this PR any more :)

blakeyp (approved):

PR #1594, opened 2024-10-11
PYIC-7552: enable outage banner for production

shivanshuit914:
## Proposed changes

- enable display outage banner for production

### What changed

- Enabled outageBanner flag for production

### Issue tracking

- [PYIC-7552](https://govukverify.atlassian.net/browse/PYIC-7552)

[PYIC-7552]: https://govukverify.atlassian.net/browse/PYIC-7552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1595, opened 2024-10-14
Revert "PYIC-7552: enable outage banner for production"

shivanshuit914:
Reverts govuk-one-login/ipv-core-front#1594


Joe-Edwards-GDS (approved):
Fine to merge this to get it out of prod asap, but I think we should get rid of this in all other environments too.

PR #1610, opened 2024-10-17
PYIC-7556: ts ipv middleware

thebauSoftwire:
### What changed
- Migrates the ipv middleware to TS as well as some helpers
- Github personal access token now required to run core-front
  - README updated (must add copy `.npmrc.template` file and add token here

### Issue tracking

- [PYIC-7556](https://govukverify.atlassian.net/browse/PYIC-7556)

[PYIC-7556]: https://govukverify.atlassian.net/browse/PYIC-7556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/pre-merge-checks.yml
```diff
@@ -29,7 +29,11 @@ jobs:
       - name: Use Node.js 20.x
         uses: actions/setup-node@v4
         with:
-          node-version: 20.11.1
+          node-version: 20.11.sef
+      - name: Setup .npmrc
+        run: |
+          cp .npmrc.template .npmrc && \
+          sed -i s/TOKEN_WITH_READ_PACKAGE_PERMISSION/${{ secrets.GITHUB_TOKEN }}/ .npmrc
```
thebauSoftwire, line 38: I don't have access to the repo secrets so could someone with access add an access token under `GITHUB_TOKEN` for me?

src/app/ipv/middleware.test.ts
```diff
@@ -1,39 +1,46 @@
...
-  APP_STORE_URL_ANDROID,
-  SERVICE_URL,
-} = require("../../lib/config");
-const qrCodeHelper = require("../shared/qrCodeHelper");
-const PHONE_TYPES = require("../../constants/phone-types");
-const {
-  SUPPORTED_COMBO_EVENTS,
-} = require("../../constants/update-details-journeys");
-const {
+import { Response, Request, NextFunction } from "express";
```
thebauSoftwire, line 1: I'm refactoring the tests in a separate PR so it doesn't hold this one back

.github/workflows/pre-merge-checks.yml
Joe-Edwards-GDS: I think this should be provided automatically by github... Looks like there's a different weird error going on before it gets here though 🤔

.npmrc.template
```diff
@@ -0,0 +1,3 @@
+engine-strict=true
+@govuk-one-login:registry=https://npm.pkg.github.com
```
Joe-Edwards-GDS, line 2: I assume this doesn't break the existing `@govuk-one-login` packages, which come from the main npm registry?

Makes me wonder if we could publish the vocab types to NPM as well... Something for later I think.

package.json
```diff
@@ -74,9 +74,11 @@
   "dependencies": {
     "@aws-sdk/client-dynamodb": "3.624.0",
     "@babel/traverse": "7.23.9",
+    "@govuk-one-login/data-vocab": "^1.7.2",
     "@govuk-one-login/frontend-analytics": "2.0.1",
     "@govuk-one-login/frontend-language-toggle": "1.1.0",
     "@govuk-one-login/frontend-passthrough-headers": "1.0.0",
+    "@types/ua-parser-js": "^0.7.39",
```
Joe-Edwards-GDS: Types should generally be installed as dev-dependencies (they're only used at build-time)

src/app.ts
```diff
@@ -35,19 +35,25 @@ declare global {
...
   }
 }
 
 // Extend session object with properties we expect
 declare module "express-session" {
   interface SessionData {
-    ipvSessionId: string;
-    clientOauthSessionId?: string;
+    ipvSessionId: string | null;
+    clientOauthSessionId?: string | null;
```
Joe-Edwards-GDS: When can these be `null` as opposed to `undefined`/not specified?

.github/workflows/pre-merge-checks.yml
TobySaundersGDS: you might need the packages permission? 
https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#about-scopes-and-permissions-for-package-registries
Also I'm not sure what .sef is - the action doesn't seem to like it though

src/app.ts
```diff
@@ -35,19 +35,25 @@ declare global {
   namespace Express {
     interface Request {
       generatedSessionId?: string;
+      cri?: {
+        id?: string;
+        redirectUrl?: string;
+      };
+      redirectURL?: URL;
```
Joe-Edwards-GDS: Where do we use these? (And do we really need to?)

Not expecting you to do any major refactorings, but this is highlighting suspicious behaviour!

src/app/ipv/middleware.ts
Joe-Edwards-GDS, line 1: I was wondering about setting up eslint rules to help sort these - it upsets me a little that the constants are all distributed. What do you think?

Probably as part of https://govukverify.atlassian.net/browse/PYIC-7557, since it'll touch a lot of files.

.github/workflows/pre-merge-checks.yml
thebauSoftwire: oops the `sef` node version was a typo 🤦‍♀️

src/app/ipv/middleware.ts
```diff
@@ -1,85 +1,129 @@
...
+
+const isClientResponse = (
+  res: PostJourneyEventResponse,
+): res is ClientResponse => {
+  return (res as ClientResponse).client !== undefined;
+};
+
+const isPageResponse = (res: PostJourneyEventResponse): res is PageResponse => {
+  return (res as PageResponse).page !== undefined;
+};
```
Joe-Edwards-GDS, line 81: I wonder if it's worth pulling these, the validators and the response types into a separate module? Not too fussed, just trying to avoid making this file too big!

.github/workflows/pre-merge-checks.yml
Joe-Edwards-GDS: Ah, yeah I think we might also need to grant permissions in the vocab repo. IIRC @Wynndow set this up for the api tests, so hopefully remembers where the setting is...

src/app/ipv/middleware.ts
```diff
@@ -207,48 +260,50 @@ function getCoiUpdateDetailsJourney(detailsToUpdate) {
   } else if (hasGivenNames) {
     return SUPPORTED_COMBO_EVENTS.UPDATE_GIVEN_NAMES;
   }
-}
+};
 
-function isInvalidDetailsToUpdate(detailsToUpdate) {
+function isInvalidDetailsToUpdate(detailsToUpdate: string[]): boolean {
```
Joe-Edwards-GDS: Change this to an arrow function?

src/app/ipv/middleware.ts
```diff
@@ -207,48 +260,50 @@ function getCoiUpdateDetailsJourney(detailsToUpdate) {
...
+export const pageRequiresUserDetails = (pageId: string): boolean => {
   return [
     PAGES.PAGE_IPV_REUSE,
-    PAGES.CONFIRM_NAME_DATE_BIRTH,
-    PAGES.CONFIRM_ADDRESS,
     PAGES.CONFIRM_DETAILS,
     PAGES.UPDATE_DETAILS,
-  ].includes(pageId);
-}
+  ].some((page) => page === pageId);
```
Joe-Edwards-GDS, line 245: This is a bit annoying! If you want to keep the existing method you'd have to cast the array to `string[]`, but this is ok too.

The real solution will be if/when TS supports proper bounded generics.

src/app/ipv/middleware.ts
```diff
@@ -439,20 +514,33 @@ async function checkFormRadioButtonSelected(req, res, next) {
...
   }
-}
+};
 
-async function formHandleUpdateDetailsCheckBox(req, res, next) {
+export const formHandleUpdateDetailsCheckBox: RequestHandler = async (
+  req,
+  res,
+  next,
+): Promise<void> => {
```
Joe-Edwards-GDS: I think you shouldn't need the return type if you put a type on the `const` itself

Joe-Edwards-GDS (commented):
reviewed up to the middleware

.github/workflows/pre-merge-checks.yml
Wynndow: Already spoke with @thebauSoftwire but for posterity core-front needs adding to the allow list on the vocab repo. Needs someone with the correct perms to do it though.

src/app/shared/deviceSniffingHelper.ts
```diff
@@ -1,12 +1,15 @@
...
+import OS_TYPES from "../../constants/os-types";
+import EVENTS from "../../constants/events";
 
 // The AppTriage event is special in that we want to send a more specialised version if we can detect the current
 // device type as being an Android phone or iPhone.
-function detectAppTriageEvent(req) {
-  const detectedPhone = sniffPhoneType(req);
+export const detectAppTriageEvent = (
+  req: Request,
+): (typeof EVENTS)[keyof typeof EVENTS] => {
```
Joe-Edwards-GDS: It's probably fine just to return `string` here?

If you want it to be specific, then maybe typescript the constants file and exporting this as a separate type.

src/app/shared/deviceSniffingHelper.ts
```diff
@@ -16,9 +19,12 @@ function detectAppTriageEvent(req) {
     default:
       return EVENTS.APP_TRIAGE;
   }
-}
+};
 
-function sniffPhoneType(req, fallback = null) {
+export const sniffPhoneType = (
+  req: Request,
+  fallback: string | null = null,
```
Joe-Edwards-GDS: should this just be an optional param?

src/app.ts
thebauSoftwire: In the ipv middleware (line 136), they are being set to null for some reason. I wasn't sure if setting it to null was intentional or not so I left as null to avoid potentially breaking something further down the line e.g. making requests to an endpoint

src/app.ts
thebauSoftwire: I don't think we need the redirectUrl in the cri obj but `criHelper.buildCredentialIssuerRedirectURL` attaches a redirectUrl to the request

src/app/shared/redirectHelper.ts
```diff
@@ -0,0 +1,16 @@
...
+  redirectUrl: string,
+): void => {
+  req.session.save((err) => {
+    if (err) {
+      logError(req, err, "Error saving session");
+      throw err;
+    }
+    return res.redirect(redirectUrl);
+  });
+};
```
Joe-Edwards-GDS, line 19: I would suggest making this return a promise, so it's more obvious that it's actually async.

You'll need to use the Promise constructor - i.e. `return new Promise((resolve, reject) => ... );`

src/app/shared/reuseHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+      {
+        buildingNumber: "7",
+        streetName: "Acorne Terrace",
+        addressLocality: "London",
+        postalCode: "N16 4QF",
+      },
+    ],
+  },
+} as AxiosResponse<ProvenUserIdentityDetails>;
```
Joe-Edwards-GDS: Can we make this a plain `ProvenUserIdentityDetails` - the response part isn't useful/relevant, and means we can avoid this cast.

src/app/shared/reuseHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export const generateUserDetails = (
+  userDetailsResponse: AxiosResponse<ProvenUserIdentityDetails>,
```
Joe-Edwards-GDS: Similarly, this could just accept a `ProvenUserIdentityDetails`

src/constants/update-details-journeys.ts
```diff
@@ -7,14 +7,17 @@ const SUPPORTED_COMBO_EVENTS = Object.freeze({
...
 });
 
-const UNSUPPORTED_COMBO_EVENTS = Object.freeze({
+export const UNSUPPORTED_COMBO_EVENTS = Object.freeze({
   address: "address",
   givenNames: "given",
   familyName: "family",
   dateOfBirth: "dob",
-});
+}) satisfies Record<UpdateDetailsOptions, string>;
```
Joe-Edwards-GDS: What's this for?

src/lib/csrf.ts
```diff
@@ -1,7 +1,8 @@
 import { csrfSync } from "csrf-sync";
+import { Request } from "express";
 
 const { csrfSynchronisedProtection } = csrfSync({
-  getTokenFromRequest: (req) => {
+  getTokenFromRequest: (req: Request) => {
```
Joe-Edwards-GDS, line 5: (this argument should be inferred from the csrf-sync types - but you can specify it if you like)

src/services/coreBackService.ts
```diff
@@ -104,8 +144,9 @@ export const postCriCallback = (
 
 export const getProvenIdentityUserDetails = (
   req: Request,
-): Promise<AxiosResponse> => {
-  return axiosInstance.get(
+): Promise<AxiosResponse<ProvenUserIdentityDetails>> => {
+  // prettier-ignore
+  return axiosInstance.get( // NOSONAR
```
Joe-Edwards-GDS: Can we accept it as a false positive in sonar instead - otherwise we'll suppress other (possibly genuine) issues on this line?

You'll need to log into Sonar with your github account, if you don't have permissions then let me know and I should be able to...

src/app.ts
Joe-Edwards-GDS: Hmm I see - that's pretty weird, it looks like we call `buildCredentialIssuerRedirectURL`, attach the redirect to the request, and then immediately call `redirectToAuthorize` which picks it off and redirects??

Can we just generate the redirect as a string, and then redirect using that string? I don't see any need to put it on the request object. (Same for the `cri` property, which also seems to be placed onto the request and immediately picked off again).

I know this is nothing to do with you, but this has helped highlight the weirdness!

src/app.ts
Joe-Edwards-GDS: Ah I see - I think that's just clumsiness distinguishing null/undefined in the existing code. I'd suggest making those lines either delete the property from the session, or set it to undefined.

(If it breaks everything then you have licence to blame me 😄 )

src/app/shared/reuseHelper.ts
thebauSoftwire: We still ended up needing a cast for the name type so have added those instead (for some reason TS couldn't figure out that it matched the type 🤔 )

src/constants/update-details-journeys.ts
thebauSoftwire: ah good spot, removing this - I think this was from trying to more strictly type `details` on line 248 in the ipv middleware but opted for just casting it as `UpdateDetailsOptions` instead since, at this point we can be more confident it's of type `UpdateDetailsOptions` given we're handling the "cancel" option already above it

src/app.ts
thebauSoftwire: I've ended up just removing this helper and placing the logic directly into ipv middleware because it was being used in just one place and could be simplified quite a bit

src/services/coreBackService.ts
thebauSoftwire: I don't see anywhere where I can mark this as a false positive so I think I may not have permissions. Are you able to do it instead?

src/app.ts
```diff
@@ -42,12 +42,13 @@ declare global {
 // Extend session object with properties we expect
 declare module "express-session" {
   interface SessionData {
-    ipvSessionId: string;
+    ipvSessionId: string | null;
```
Joe-Edwards-GDS: This is weird - you're making this nullable, but in the middleware you're setting it to undefined.  I thought we had strict null checks, which would spot this, but it's not applying for some reason. 🤔

In any case, I think it would be clearest if this was an optional property like all the others...

src/app/shared/redirectHelper.ts
Joe-Edwards-GDS: This is quite right, it'll probably look something like:
```
return new Promise((resolve, reject) => {
  req.session.save(function (err) {
    if (err) {
      reject(err);
    } else {
      resolve(res.redirect(redirectUrl));
    }
  });
});
```

(Then you'll want to make sure the result is always returned and/or awaited)

src/app/ipv/middleware.ts
```diff
@@ -116,57 +145,67 @@ async function handleBackendResponse(req, res, backendResponse) {
...
     data: backendResponse,
   };
   req.log.error({ message, level: "ERROR" });
   throw new Error(message.description);
-}
+};
 
-function tryValidateCriResponse(criResponse) {
-  if (!criResponse?.redirectUrl) {
+const isValidCriResponse = (criResponse: CriResponse): boolean => {
```
Joe-Edwards-GDS: (I wonder if these should go in `validators/postJourneyEventResponse` as well?)

Joe-Edwards-GDS (commented):
Looks almost there - really just the promise thing that definitely needs fixing 🙂

Joe-Edwards-GDS (approved):
Thanks!

PR #1612, opened 2024-10-21
PYIC-7574: add configurable notification banner

shivanshuit914:
## Proposed changes

- Add configurable notification banner 

### What changed

- Implemented middleware handler to read SSM params to display notification banner

### Issue tracking

- [PYIC-7574](https://govukverify.atlassian.net/browse/PYIC-7574)

## Checklists

[PYIC-7574]: https://govukverify.atlassian.net/browse/PYIC-7574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/notification-banner-handler.ts
```diff
@@ -0,0 +1,29 @@
+import { RequestHandler } from "express";
+import { getParameter } from "../services/parameterStoreService";
+
+const notificationBannerHandler: RequestHandler = async (req, res, next) => {
+  try {
+    const data = await getParameter("notification-banner");
```
shivanshuit914: Not sure if we need `notification-banner` dynamically to load from config/env.

browser-tests/compose.yaml
```diff
@@ -24,6 +24,7 @@ services:
       EXTERNAL_WEBSITE_HOST: http://localhost:4601
       CONTACT_URL: https://home.account.gov.uk/contact-gov-uk-one-login
       LANGUAGE_TOGGLE: true
+      notification-banner: '{"pageId":"/dev/template/confirm-your-details/en","bannerMessage":"<h3>Test banner</h3> <p>This is a test banner</p>","bannerMessageCy":"<h3>Welsh Test banner</h3> <p>This is a Welsh test banner</p>","startTime":1729604741710,"endTime":1918907190000}'
```
Joe-Edwards-GDS: Can we try and stick with standard format for environment variables - i.e. `NOTIFICATION_BANNER`?

deploy/template.yaml
```diff
@@ -918,6 +918,18 @@ Resources:
...
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - ssm:GetParameter
+                  - ssm:GetParameters
+                  - ssm:GetParametersByPath
+                Resource:
+                  - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:*"
```
Joe-Edwards-GDS: I think we can simplify this - have a look at the `SSMParameterReadPolicy` configuration in core-back.

We should also restrict the parameters that are accessible - we don't need access to every parameter in the account.

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,27 @@
...
+  bannerMessage: string;
+  bannerMessageCy: string;
+  startTime: number;
+  endTime: number;
+}
+
+export const getParameter = async (
+  name: string,
+): Promise<BannerConfig | undefined> => {
+  if (process.env.NODE_ENV === "local") {
```
Joe-Edwards-GDS: IMO this probably belongs in the `notification-banner-handler` instead (so we avoid calling the 'parameter store service' if it's local)

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,27 @@
...
+    return JSON.parse(process.env[name] ?? "{}");
+  }
+
+  const client = new SSMClient({ region: "eu-west-2" });
+  const data = await client.send(new GetParameterCommand({ Name: name }));
```
Joe-Edwards-GDS: We'll need to be a bit careful here, since this will run on _every_ request. It should be cached, and ideally the cache should be refreshed asynchronously!

You could either:
- Refresh the cache on-demand - i.e. only after a request needs it
- Refresh on a schedule in the background (perhaps using a node scheduler library)

src/handlers/notification-banner-handler.ts
Joe-Edwards-GDS: Can we catch errors thrown here - we should log them, but not break rendering the page.

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,27 @@
...
+  bannerType?: string;
+  bannerMessage: string;
+  bannerMessageCy: string;
+  startTime: number;
+  endTime: number;
+}
+
+export const getParameter = async (
+  name: string,
+): Promise<BannerConfig | undefined> => {
```
Joe-Edwards-GDS: This is a bit odd - the method appears to be a generic parameter store get, but it always returns a `BannerConfig`?

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,27 @@
+import { SSMClient, GetParameterCommand } from "@aws-sdk/client-ssm";
+
+interface BannerConfig {
+  pageId: string;
+  bannerType?: string;
+  bannerMessage: string;
+  bannerMessageCy: string;
+  startTime: number;
```
Joe-Edwards-GDS: Can I suggest that these are ISO8601 timestamps instead? That way they'll be much easier to configure

src/handlers/notification-banner-handler.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    const currentTime = Date.now();
+    if (
+      req.path === data.pageId &&
+      currentTime >= data.startTime &&
+      currentTime <= data.endTime
+    ) {
+      res.locals.displayBanner = true;
+      res.locals.bannerType = data.bannerType;
```
Joe-Edwards-GDS: What is `bannerType`?

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,27 @@
+import { SSMClient, GetParameterCommand } from "@aws-sdk/client-ssm";
+
+interface BannerConfig {
```
Joe-Edwards-GDS: Do you think we should set this up so that:
- You can configure multiple banners at the same time?
- You can configure each banner to appear on multiple pages?

Joe-Edwards-GDS (commented):
This looks along the right lines, but I've got a few suggestions

src/handlers/notification-banner-handler.ts
shivanshuit914: Its parameter in banner component, if we want to display green success banner we can pass it from here.

src/services/parameterStoreService.ts
shivanshuit914: Hm, need to extend this for multiple page/banner support. Thanks

src/handlers/notification-banner-handler.ts
```diff
@@ -0,0 +1,41 @@
...
+  BannerConfig,
+} from "../services/parameterStoreService";
+import { transformError } from "../app/shared/loggerHelper";
+
+const notificationBannerHandler: RequestHandler = async (req, res, next) => {
+  try {
+    res.locals.displayBanner = false;
+    const bannerConfigs =
+      process.env.NODE_ENV === "local"
+        ? JSON.parse(process.env["NOTIFICATION_BANNER"] ?? "{}")
```
Joe-Edwards-GDS: Should probably default to `[]` now

src/handlers/notification-banner-handler.ts
```diff
@@ -0,0 +1,41 @@
...
+        currentTime <= data.endTime
+      ) {
+        res.locals.displayBanner = true;
+        res.locals.bannerType = data.bannerType;
+        res.locals.bannerMessage =
+          req.i18n.language === "en"
+            ? data.bannerMessage
+            : data.bannerMessageCy;
+      }
+      next();
```
Joe-Edwards-GDS: This will call `next()` for every banner config - I think it should go outside the loop?

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+  const cacheDuration = 5 * 60 * 1000; // Cache duration in milliseconds (5 minutes)
+  const cachedParam = parameterCache.get(name);
+  if (cachedParam && cachedParam.expiration > Date.now()) {
+    return cachedParam.value;
+  }
+  const client = new SSMClient({ region: "eu-west-2" });
+  const data = await client.send(new GetParameterCommand({ Name: name }));
+
+  if (!data.Parameter?.Value) {
+    return;
```
Joe-Edwards-GDS: What does this case represent? This is happening before the cache, meaning that this type of response will not be cached.

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+const parameterCache: Map<string, { value: JSON; expiration: number }> =
+  new Map();
+
+export const getParameter = async (name: string): Promise<JSON | undefined> => {
+  const cacheDuration = 5 * 60 * 1000; // Cache duration in milliseconds (5 minutes)
+  const cachedParam = parameterCache.get(name);
+  if (cachedParam && cachedParam.expiration > Date.now()) {
+    return cachedParam.value;
+  }
+  const client = new SSMClient({ region: "eu-west-2" });
```
Joe-Edwards-GDS: We should build one SSMClient and then reuse it, rather than constructing a new one each time

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+  const data = await client.send(new GetParameterCommand({ Name: name }));
+
+  if (!data.Parameter?.Value) {
+    return;
+  }
+
+  const parameterValue = JSON.parse(data.Parameter?.Value);
```
Joe-Edwards-GDS: I don't think parameters need to be JSON-valued - would it make sense to return/cache the string value here, and parse the JSON in `getNotificationBanner`?

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+  parameterCache.set(name, {
+    value: parameterValue,
+    expiration: Date.now() + cacheDuration,
+  });
+
+  return parameterValue;
+};
+
+export const getNotificationBanner = async (): Promise<
```
Joe-Edwards-GDS: Should this live inside `notification-banner-handler.ts`?

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+  bannerMessageCy: string;
+  startTime: string;
+  endTime: string;
+}
+
+const parameterCache: Map<string, { value: JSON; expiration: number }> =
+  new Map();
+
+export const getParameter = async (name: string): Promise<JSON | undefined> => {
+  const cacheDuration = 5 * 60 * 1000; // Cache duration in milliseconds (5 minutes)
```
Joe-Edwards-GDS: Can you pull this out into a constant? It would be even better if it was config, but not totally necessary.

src/services/parameterStoreService.ts
```diff
@@ -0,0 +1,43 @@
...
+  const cachedParam = parameterCache.get(name);
+  if (cachedParam && cachedParam.expiration > Date.now()) {
+    return cachedParam.value;
+  }
```
Joe-Edwards-GDS, line 17: I think there's a race condition here when your cache expires - all requests that arrive while requesting the new value will fetch it as well.

That's probably ok under low load, but may be an issue under high load (e.g. at 2000r/s). You might need to store a promise in the cache instead.

src/services/parameterStoreService.ts
Joe-Edwards-GDS, line 1: If this doesn't perform well enough, we could also consider a stale-while-revalidate pattern - where we return a stale value and refresh the cache asynchronously. In any case we should make sure to get another load test.

deploy/template.yaml
Joe-Edwards-GDS: (I think this still applies)

Joe-Edwards-GDS (commented):
Thanks - it's looking better with the caching, but might need to tighten it up a little more.

src/handlers/notification-banner-handler.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  beforeEach(() => {
+    req = {
+      session: {},
+      path: "/some-page",
+    } as any;
+
+    res = {
+      locals: {},
+    } as any;
+    process.env.NODE_ENV = "production";
```
Joe-Edwards-GDS, line 22: We should be a little bit careful with this - it's not the end of the world, but this will leak out of the test, and could affect other cases.

If you want to test that case, we could reset the environment after each test. Alternatively we could just skip testing the local-only behaviour - since it's not production code.

Joe-Edwards-GDS (approved):
One thing we could tidy up in the test, but otherwise looks good.

We should set up the new config (with an empty array) before merging though :)

PR #1624, opened 2024-10-25
PYIC-7496: reuse and coi journey browser tests

thebauSoftwire:
…tity-details as well as the update details screens

### What changed
- Adds browser tests for reuse and coi journeys.
- Stubs call to `/user/proven-identity-details` endpoint

Note: snapshot tests already covered the display of the `confirm-details`, `update-details` and `page-ipv-reuse` screens so didn't add any. However, the error handling I've added as functional tests.

### Why did it change
Increase coverage of browser tests to cover the `/user/proven-identity-details` endpoint

### Issue tracking

- [PYIC-7496](https://govukverify.atlassian.net/browse/PYIC-7496)

[PYIC-7496]: https://govukverify.atlassian.net/browse/PYIC-7496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/browser-tests/functional-tests/core-back-errors.spec.js
+++ b/browser-tests/functional-tests/core-back-errors.spec.js
@@ -1,0 +1,29 @@
+const { test, expect } = require("@playwright/test");
+const { getAuthoriseUrlForJourney } = require("./helpers");
+
+test.describe("Error tests", () => {
+  test("Handles an unexpected error from core-back", async ({ page }) => {
+    // Start a session
+    await page.goto(getAuthoriseUrlForJourney("testError"));
+
+    // Go to the error
+    await page.goto("/ipv/journey/page-ipv-identity-document-start/error")
+
+    // When we come back with an error
+    const textLocator = await page.getByText("Sorry, there is a problem");
+    await expect(textLocator).toBeVisible();
+  });
+
+  test("Handles an unexpected error after a CRI callback", async ({ page }) => {
+    // Start a session
+    await page.goto(getAuthoriseUrlForJourney("testCriError"));
+
+    // Go to the DCMAW CRI
+    await page.click("input[type='radio'][value='appTriage']");
+    await page.click("button[id='submitButton']");
+
+    // When we come back from DCMAW with an error
+    const textLocator = await page.getByText("Sorry, there is a problem");
+    await expect(textLocator).toBeVisible();
+  });
+});
--- a/browser-tests/functional-tests/core-back-errors.spec.js
+++ b/browser-tests/functional-tests/core-back-errors.spec.js
@@ -4,10 +4,11 @@ const { getAuthoriseUrlForJourney } = require("./helpers");
 test.describe("Error tests", () => {
   test("Handles an unexpected error from core-back", async ({ page }) => {
     // Start a session
-    await page.goto(getAuthoriseUrlForJourney("testError"));
+    await page.goto(getAuthoriseUrlForJourney("testUnexpectedError"));
 
-    // Go to the error
-    await page.goto("/ipv/journey/page-ipv-identity-document-start/error")
+    // Go to the DCMAW CRI
+    await page.click("input[type='radio'][value='appTriage']");
+    await page.click("button[id='submitButton']");
 
     // When we come back with an error
     const textLocator = await page.getByText("Sorry, there is a problem");
--- a/browser-tests/functional-tests/functional.spec.js
+++ b/browser-tests/functional-tests/functional.spec.js
@@ -1,9 +1,5 @@
 const { test, expect } = require("@playwright/test");
-
-const HTTP_HEADER_USER_AGENT_ANDROID =
-  "Mozilla/5.0 (Linux; Android 8.0.0; Nexus 5X Build/OPR6.170623.013) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.98 Mobile Safari/537.36";
-const HTTP_HEADER_USER_AGENT_IPHONE =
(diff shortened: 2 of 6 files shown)


browser-tests/functional-tests/functional.spec.js
```diff
@@ -69,6 +69,71 @@ test.describe.parallel("Functional tests", () => {
     const url = page.url();
     expect(url).toBe(`${domainUrl}/ipv/page/page-multiple-doc-check`);
   });
+
+  test("Successfully gets proven user details from core-back for the page-ipv-reuse screen", async ({ page }) => {
```
thebauSoftwire, line 69: I wasn't sure what other scenarios to include in the functional tests since things like handling of multiple names and address, throwing of errors are covered by the snapshot tests and/or unit tests. So settled on these as they covered the routing, handling of the core-back calls and handling of errors displayed to the user

Joe-Edwards-GDS: Yep, I think what we're really looking for on reuse is to ensure that the user data is passed back correctly from the backend API request.

As you say, the snapshot tests cover the precise rendering behaviour, and the UTs cover the transformation logic.

browser-tests/functional-tests/functional.spec.js
Joe-Edwards-GDS, line 1: Do you think it's worth splitting this file up yet?

(cc @blakeyp / @DanCorderIPV - if you had any plans)

Joe-Edwards-GDS, line 1: The other thing that _might_ be useful to test is the mapping from the checkbox options (e.g. on update-details) to the backend event.

It's fairly well covered by UTs, but given that it's one of the more unique form inputs on the site. E.g. confirming that when you check given name + address it invokes a backend endpoint with `/journey/given-names-and-address`. (Definitely don't check all cases - one is probably enough!)

What do you think?

thebauSoftwire: ah good point, I'll add this now thanks :)

Joe-Edwards-GDS (dismissed):
LGTM - might be worth having Dan or Patrick cast an eye, if they've got time.

browser-tests/functional-tests/functional.spec.js
thebauSoftwire: I've done some splitting up for that file based on what was already there, let me know if you disagree :)

Joe-Edwards-GDS (dismissed):

browser-tests/functional-tests/core-back-errors.spec.js
```diff
@@ -0,0 +1,29 @@
+const { test, expect } = require("@playwright/test");
+const { getAuthoriseUrlForJourney } = require("./helpers");
+
+test.describe("Error tests", () => {
+  test("Handles an unexpected error from core-back", async ({ page }) => {
+    // Start a session
+    await page.goto(getAuthoriseUrlForJourney("testError"));
+
+    // Go to the error
```
DanCorderIPV: This does test the right thing, but it took me a little while to work out how.
It might fit with the other tests a bit more if this also just submitted the initial form and core back returned an error, rather than using the more unusual "submit and action via a page request" route.

browser-tests/imposter/config/api-config.yaml
```diff
@@ -213,3 +218,107 @@ resources:
         {
           "page": "prove-identity-another-type-photo-id"
         }
+
+  # Journeys where there is an existing identity
```
DanCorderIPV: I think these two `/journey/next` routes should go up around line 53 so they are together with the default version.

browser-tests/imposter/config/api-config.yaml
```diff
@@ -213,3 +218,107 @@ resources:
...
+      currentPage:
+        operator: NotExists
+    response:
+      statusCode: 200
+      content: |
+        {
+          "page": "confirm-your-details"
+        }
+
+  - method: GET
```
DanCorderIPV, line 277: This needs a comment explaining what it's for

browser-tests/imposter/config/api-config.yaml
```diff
@@ -213,3 +218,107 @@ resources:
...
+              "postalCode": "BA2 5AA",
+              "buildingNumber": "8",
+              "addressLocality": "BATH",
+              "validFrom": "1000-01-01",
+              "subBuildingName": ""
+            }
+          ]
+        }
+
+  - method: POST
```
DanCorderIPV, line 312: Comment please

browser-tests/imposter/config/api-config.yaml
```diff
@@ -213,3 +218,107 @@ resources:
...
+          "page": "confirm-your-details"
+        }
+
+  - method: GET
+    path: /user/proven-identity-details
+    requestHeaders:
+      content-type: "application/json"
+      ipv-session-id:
+        value: "KennethDecerqueira"
+        operator: Contains
```
DanCorderIPV, line 283: I'm a little nervous about using `Contains` in this configuration - it starts to make this config harder to reason about and we don't want a general simulation of core back, just the specific routes we expect to test.
That said, I can see that it saves a fair bit of duplication here, but I think we need to keep an eye on it and not let it get too clever.

browser-tests/functional-tests/functional.spec.js
DanCorderIPV: Looks good to me thanks

DanCorderIPV (approved):

PR #1625, opened 2024-10-28
PYIC-7573 Update redaction method to cope with relative URLs

Joe-Edwards-GDS:
### What changed

- Update URL redaction to work on relative URLs
- Update other error scenarios to skip pino-http error generation

### Why did it change

- The URL redaction logic didn't work on relative URLs
- pino-http will generate a spurious error (and stack trace) for any responses with status >=500

### Issue tracking

- [PYIC-7573](https://govukverify.atlassian.net/browse/PYIC-7573)

[PYIC-7573]: https://govukverify.atlassian.net/browse/PYIC-7573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -1,22 +1,26 @@
 import { Request, RequestHandler, Response } from "express";
 import pino, { Logger } from "pino";
 import pinoHttp from "pino-http";
-import { HANDLED_ERROR } from "../handlers/internal-server-error-handler";
 
-const sensitiveParams = ["request", "code"];
+export const HANDLED_ERROR = new Error(
+  "Placeholder errors that do not require additional logging",
+);
+
+const SENSITIVE_PARAMS = ["request", "code"];
+const BASE_PLACEHOLDER = "http://placeholder-for-redaction";
 
 export const redactQueryParams = (
   url: string | undefined,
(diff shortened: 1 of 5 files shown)


src/lib/logger.ts
```diff
@@ -1,22 +1,26 @@
...
-      const parsedUrl = new URL(url);
-      for (const param of sensitiveParams) {
+      const parsedUrl = new URL(url, BASE_PLACEHOLDER);
+      for (const param of SENSITIVE_PARAMS) {
         if (parsedUrl.searchParams.has(param)) {
           parsedUrl.searchParams.set(param, "hidden");
         }
       }
-      return parsedUrl.href;
+      return parsedUrl.href.replace(BASE_PLACEHOLDER, "");
```
Joe-Edwards-GDS, line 23: This is a bit of a fudge - URL does not support relative urls (because they are not technically URLs...), but we can work around by using a placeholder base and then stripping it afterwards.

(See https://github.com/nodejs/node/issues/12682 for a very long thread about this...)

blakeyp (approved):

PR #1629, opened 2024-10-30
PYIC-7603: upgrade express js to latest version 5.0.0

shivanshuit914:
## Proposed changes

- Update express js to 5.0.0

### What changed

- Updated  express js to 5.0.0
- Removed express-async-errors
- Updated body parser to latest version

### Issue tracking

- [PYIC-7603](https://govukverify.atlassian.net/browse/PYIC-7603)

[PYIC-7603]: https://govukverify.atlassian.net/browse/PYIC-7603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (commented):
Thanks Shiv - do you think this lets us simplify any of our error handling?

I.e. we should be able to just throw errors from our middleware (like you would in Java, say), rather than needing to catch them and call `next(err)`.

shivanshuit914:
> Thanks Shiv - do you think this lets us simplify any of our error handling?
> 
> I.e. we should be able to just throw errors from our middleware (like you would in Java, say), rather than needing to catch them and call `next(err)`.

Yes, I am creating separate PR for it. Thought to create separate PR in case if there is breaking changes in migration and need to rollback.

Joe-Edwards-GDS (approved):

PR #1630, opened 2024-10-31
PYIC-7573 Return correct status code for CSRF errors

Joe-Edwards-GDS:
### What changed

Identify `http-errors` errors and use their status code.

### Why did it change

CSRF errors use the `http-errors` library, but these weren't being recognised and resulted in a 500, rather than the desired 403.


src/handlers/internal-server-error-handler.ts
Joe-Edwards-GDS, line 1: Perhaps we could have used the `http-errors` library ourselves, but not sure there's much value in changing it again now :shrug:

Wynndow (approved):

PR #1631, opened 2024-11-01
PYIC-7638 Convert browser tests to TS

Joe-Edwards-GDS:
### What changed

Typescriptified the browser tests

### Why did it change

We love TS now

### Issue tracking

- [PYIC-7638](https://govukverify.atlassian.net/browse/PYIC-7638)

[PYIC-7638]: https://govukverify.atlassian.net/browse/PYIC-7638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


browser-tests/playwright.config.ts
```diff
@@ -1,5 +1,7 @@
-const { devices } = require("@playwright/test");
-require("dotenv").config();
+import { devices } from "@playwright/test";
+import dotenv from "dotenv";
+
+dotenv.config();
 
 module.exports = {
```
thebauSoftwire: Is there a reason why we still use `module.exports` here instead of an es6 export? Could something like this work?
```
export default defineConfig({
  workers: 6,
  projects: [
    {
      name: "chromium",
      use: {
        ...devices["Desktop Chrome"],
        headless: true,
      },
    },
  ],
  use: {
    baseURL: process.env?.WEBSITE_HOST || "http://localhost:4601",
    headless: true,
  },
})
```

thebauSoftwire: And just one tiny thing, I think line 21 in the browser tests README needs to be updated to `playwright.config.ts`

thebauSoftwire (commented):
Looks good, just one comment

thebauSoftwire (approved):

PR #1635, opened 2024-11-01
PYIC-7603: remove next error

shivanshuit914:
## Proposed changes

- Refactor middleware remove next(error) 

### What changed

- Removed next(error) that does try catch, and now uses express 5 error handling

### Issue tracking

- [PYIC-7603](https://govukverify.atlassian.net/browse/PYIC-7603)

[PYIC-7603]: https://govukverify.atlassian.net/browse/PYIC-7603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/tests/handleJourneyActionRequest.test.ts
+++ b/src/app/ipv/tests/handleJourneyActionRequest.test.ts
@@ -58,10 +58,10 @@ describe("handleJourneyActionRequest", () => {
     coreBackServiceStub.postJourneyEvent = callBack;
     callBack.onCall(0).returns(clientEventResponse);
 
-    await middleware.handleJourneyActionRequest(testReq, res, next);
-    expect(next).to.have.been.calledWith(
-      sinon.match.has("message", "Client Response redirect url is missing"),
-    );
+    await expect(
+      (async () =>
+        await middleware.handleJourneyActionRequest(testReq, res, next))(),
+    ).to.be.rejectedWith(Error, "Client Response redirect url is missing");
   });
 
   it("should call next with an error message given redirect url is missing from CRI event response", async () => {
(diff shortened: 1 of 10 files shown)


src/app/ipv/tests/handleJourneyActionRequest.test.ts
```diff
@@ -78,22 +78,26 @@ describe("handleJourneyActionRequest", () => {
...
-    await middleware.handleJourneyActionRequest(req, res, next);
-    expect(
-      coreBackServiceStub.postJourneyEvent.firstCall,
-    ).to.have.been.calledWith(
-      req,
-      "some-journey-event",
-      req.session.currentPage,
-    );
+    try {
+      await middleware.handleJourneyActionRequest(req, res, next);
```
Joe-Edwards-GDS, line 90: Note that this will implicitly pass if no error is thrown (although it'll probably log a warning, because there will be no assertions).

It might be better to assert the error and call separately - e.g.:
```
await expect(() => middleware.handleJourneyActionRequest(req, res, next)).to.be.rejectedWith(Error);

expect(coreBackServiceStub.postJourneyEvent.firstCall)...
```

Alternatively, you could put a fail afterwards, but it feels a bit clunkier:
```
try {
  await middleware.handleJourneyActionRequest(req, res, next);
  expect.fail("Expected exception");
} catch (err) {
  // assert on errors
}
```

Joe-Edwards-GDS: (a few other cases of this elsewhere!)

Joe-Edwards-GDS (approved):
This looks ok, but a note on the way you're asserting on exceptions...

Have you been able to check that it does handle exceptions properly when they're thrown in the running app? I don't see why it wouldn't, but nice to be confident.

shivanshuit914:
> This looks ok, but a note on the way you're asserting on exceptions...
> 
> Have you been able to check that it does handle exceptions properly when they're thrown in the running app? I don't see why it wouldn't, but nice to be confident.

Yes, I did run few journey on my dev env, it worked fine.

PR #1641, opened 2024-11-08
PYIC-7613: Add new alarm for 80% 5XX

Wynndow:
### What changed

Add new alarm for 80% 5XX

### Why did it change

This adds a new alarm that will send a PagerDuty alert if the core-front REST apigateway starts to return more than 80% of 5XX responses.

### Issue tracking

- [PYIC-7613](https://govukverify.atlassian.net/browse/PYIC-7613)

[PYIC-7613]: https://govukverify.atlassian.net/browse/PYIC-7613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1462,6 +1462,7 @@ Resources:
 ####################################################################
 
   FrontLoadBalancer5xxErrors:
+    # This is monitoring 5XXs originating from the load balancer, and not from the target group
     Type: AWS::CloudWatch::Alarm
     Condition: IsNotDevelopment
     Properties:
@@ -1495,6 +1496,7 @@ Resources:
             Stat: Sum
 
   FrontTargetGroup5xxErrors:
+    # This is monitoring 5XXs originating from the target group, and not from the load balancer
     Type: AWS::CloudWatch::Alarm
     Condition: IsNotDevelopment
     Properties:
@@ -1528,6 +1530,7 @@ Resources:
             Stat: Sum
 
   FrontTargetGroup5xxPercentErrors:
+    # This is used by the canary deployment stack
     Type: AWS::CloudWatch::Alarm
     Condition: UseCanaryDeployment
     Properties:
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -1527,6 +1527,55 @@ Resources:
             Period: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, tg500ErrorWindow ]
             Stat: Sum
 
+  FrontRestApiGateway5xxErrorsPercentage:
```
Wynndow: There appears to be some duplication with the above alarms, although they're informed by metrics from the load balancer and target groups. They use the SUM of 5XX metrics over a much longer window.
Without really knowing the importance of these alarms being in this configuration I've opted to add a new one rather than adjust. Up for debate.

shivanshuit914 (dismissed):

deploy/template.yaml
Joe-Edwards-GDS: I think it would be good to assess across all of these holistically... I'd be keen to see a clear list of each alarm and what it's purpose is - either as comments here, or in a runbook somewhere.

At the moment it looks like we have:
- `FrontLoadBalancer5xxErrors` - ALB 5XX errors (>2 in 5m) - triggers PD
- `FrontTargetGroup5xxErrors` - Target group 5XX errors (>50 in 5m) - triggers PD
- `FrontTargetGroup5xxPercentErrors` - Target group 5XX errors (>5% in 1m) - used by the canary deployments
- `FE5XXErrorAlarm` - API GW 5XX errors (>10% in 1m) - goes somewhere?
- `FrontRestApiGateway5xxErrorsPercentage` - API GW 5XX errors (>80% in 1m) - what you've just added

There might be a good reason to have lots of different ones - e.g. I can see value in having both %-based and absolute thresholds, and it seems reasonable that the canary alarm is a bit more sensitive than the P1 incident alarm; but at the moment it's not obvious what the rationale is.

Joe-Edwards-GDS: FWIW it's the first one that seems to trigger all the real incidents!

deploy/template.yaml
```diff
@@ -1568,6 +1571,60 @@ Resources:
...
+      EvaluationPeriods: 5
+      DatapointsToAlarm: 2
+      Threshold: 10
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      TreatMissingData: notBreaching
+      Metrics:
+        - Id: errorThreshold
+          Label: Threshold error percentage
+          ReturnData: true
+          Expression: IF(invocations<50,0,errorPercentage)
```
Wynndow, line 1598: Overnight invocations are around the 8-10 per mintue mark. These seem to be from some sort of smoke testing happening every 3 minutes. Setting this alarm to only invoke when invocations are greater than 50 will mean that this alarm will effectively ignore the traffic generate by the smoke tests.
If we have actualy users overnight, hopefullt the traffic would increase above 50 so it gets caught. We could reduce the limit here, but then we might get alerted overnight for a very low number of 500s.

deploy/template.yaml
```diff
@@ -1568,6 +1571,60 @@ Resources:
           Period: 60
           Stat: Sum
 
+  FrontRestApiGateway5XXErrorAlarm:
```
Wynndow, line 1574: Does this alarm render the `FrontTargetGroup5xxErrors` redundant?
In prod, we'd need to see two 5 minute periods back to back with more than 50 5xx errors from the target group for that alarm to trigger.
If we had 50 errors in minute 1 and then 50 in minute 6, the new alarm wouldn't trigger. So maybe this is still useful?

deploy/template.yaml
Joe-Edwards-GDS: I think the smoke test is hooked up to a separate alarm anyway, so if it starts failing we should know (but it may be less responsive than this one)

deploy/template.yaml
Joe-Edwards-GDS: I think `FrontTargetGroup5xxErrors` is used by the canary deployments (not incident alerts) - it might make sense to be a little more sensitive for canaries since it's a short window and won't receive as much traffic.

Joe-Edwards-GDS (approved):
Seems reasonable to me, I guess we'll see how often it alerts...

Is it possible to see 'when would this have alerted historically'? Might be tricky if the metrics have lost resolution...

Wynndow:
> Is it possible to see 'when would this have alerted historically'? Might be tricky if the metrics have lost resolution...

Good question - I've not seen that functionality before. Could trawl a graph of the metrics but might be a bit of a PITA...

PR #1643, opened 2024-11-11
BAU: Update coverage exclusions

Joe-Edwards-GDS:
### What changed

Update coverage exclusions

### Why did it change

We were excluding some things we shouldn't, and not including some things we should

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -2,7 +2,8 @@ sonar.projectKey=ipv-core-front
 sonar.organization=govuk-one-login
 
 sonar.sources=src/
-sonar.exclusions=**/assets/**,**/*.test.js,**/*.test.ts,**/service-unavailable-s3/**,**/config/**,**/lib/**,**/*router.js,**/axiosHelper.js,**/app.ts,**/development/**
+# Remove tests, config, routing and development-only code from code coverage
+sonar.exclusions=**/*.test.ts,**/config/**,**/*router.ts,**/app.ts,**/development/**
 
 sonar.tests=src/
 sonar.test.inclusions=**/*.test.js,**/*.test.ts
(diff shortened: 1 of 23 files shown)


sonar-project.properties
```diff
@@ -2,7 +2,8 @@ sonar.projectKey=ipv-core-front
 sonar.organization=govuk-one-login
 
 sonar.sources=src/
-sonar.exclusions=**/assets/**,**/*.test.js,**/*.test.ts,**/service-unavailable-s3/**,**/config/**,**/lib/**,**/*router.js,**/axiosHelper.js,**/app.ts,**/development/**
+# Remove tests, config, routing and development-only code from code coverage
+sonar.exclusions=**/*.test.ts,**/config/**,**/*router.ts,**/app.ts,**/development/**
```
Joe-Edwards-GDS, line 6: Previously we were excluding `lib` which doesn't feel right - this introduces some uncovered code, but it's probably accurate.

Joe-Edwards-GDS: It also hides these files from Sonar analysis, which isn't really desirable!

src/config/config.ts
```diff
@@ -9,25 +9,25 @@ const coreBackApiUrl = appEnv.isLocal
   : appEnv.getServiceURL("core-back-api");
 
 export default {
-  ANDROID_APP_ID: process.env.ANDROID_APP_ID || "uk.gov.documentchecking",
+  ANDROID_APP_ID: process.env.ANDROID_APP_ID ?? "uk.gov.documentchecking",
```
thebauSoftwire, line 12: just out of curiosity, do we not care if these were empty strings or some other falsey value?

thebauSoftwire (approved):
Nice! Just one comment for my own learning

src/config/config.ts
Joe-Edwards-GDS: I think we potentially care - but in general empty string might be a valid value!

We could rewrite this to validate config a bit more effectively, like we do for the api test config (for example), but trying to avoid too much refactoring at once...

Joe-Edwards-GDS: (I changed it because Sonar was flagging it as an issue)

PR #1507, opened 2024-07-30
IPS-954 Perf Test Environment Changes

vam-gds:
### What changed
IPS-954 Perf Test Environment Changes

### Issue tracking

- [PYIC-XXXX](https://govukverify.atlassian.net/browse/PYIC-XXXX)

### Environment variables or secrets

- [ ] No environment variables or secrets were added or changed

- [ ] Documented in the [README](./blob/main/README.md)
- [ ] Added to deployment repository
- [ ] Added to local startup repository

### Other considerations

- [ ] Update [README](./blob/main/README.md) with any new instructions or tasks

PR #1644, opened 2024-11-11
PYIC-7612: configure healthcheck for ECS

thebauSoftwire:
### What changed
Configured a healthcheck in TaskDefinition for ECS
Adds a healthcheck to the DockerFile with matching config

| logs in /aws/ecs/core-front-theab-CoreFront-ECS | Screenshot 2024-11-12 at 11 21 44 |
|-|-|

### Why did it change
FE scaling doesn't work unless healthchecks are properly configured

### Issue tracking

- [PYIC-7612](https://govukverify.atlassian.net/browse/PYIC-7612)

[PYIC-7612]: https://govukverify.atlassian.net/browse/PYIC-7612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,10 @@ COPY --from=khw46367.live.dynatrace.com/linux/oneagent-codemodules-musl:nodejs /
 ENV LD_PRELOAD /opt/dynatrace/oneagent/agent/lib64/liboneagentproc.so
 
 ENV PORT 8080
+
+HEALTHCHECK --interval=5s --timeout=2s --retries=10 \
+  CMD curl -f http://localhost:8080/healthcheck || exit 1
+
 EXPOSE 8080
 
 ENTRYPOINT ["tini", "--"]
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -857,7 +857,7 @@ Resources:
           HealthCheck:
             Command:
               - "CMD-SHELL"
-              - "curl -f <http://localhost:8080/healthcheck> || exit 1"
+              - "curl -f http://localhost:8080/healthcheck || exit 1"
             Interval: 5
             Retries: 10
             Timeout: 2
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -854,6 +854,13 @@ Resources:
           PortMappings:
             - ContainerPort: 8080
               Protocol: tcp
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -43,6 +43,10 @@ COPY --from=khw46367.live.dynatrace.com/linux/oneagent-codemodules-musl:nodejs /
 ENV LD_PRELOAD /opt/dynatrace/oneagent/agent/lib64/liboneagentproc.so
 
 ENV PORT 8080
+
+HEALTHCHECK --interval=5s --timeout=2s --retries=10 \
```
thebauSoftwire, line 47: The 5s is from the open [perf test PR](https://github.com/govuk-one-login/ipv-core-front/pull/1507/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R35)

deploy/template.yaml
```diff
@@ -854,6 +854,13 @@ Resources:
           PortMappings:
             - ContainerPort: 8080
               Protocol: tcp
+          HealthCheck:
+            Command:
+              - "CMD-SHELL"
+              - "curl -f <http://localhost:8080/healthcheck> || exit 1"
```
Joe-Edwards-GDS: I think those angle brackets are probably placeholders - you probably just want the URL?

Dockerfile
Joe-Edwards-GDS: Thanks for the link.

This is going to be very noisy (every 5s from docker + every 15s from ALB + every 30s from NLB), so I wonder if we should lower the healthcheck log level to `debug`, to allow us to tune it out.

(@blakeyp any thoughts? I know we've flip-flopped on this quite a bit!)

Joe-Edwards-GDS (approved):
Can you check what the deployed logs look like in your dev env, to see what it looks like?

thebauSoftwire:
Have updated the PR overview with ss of the logs in AWS

Dockerfile
blakeyp: Sorry, missed this, yeah that sounds if we could reduce the log level

PR #1647, opened 2024-11-11
PYIC-7616: add overload-protection to FE

thebauSoftwire:
### What changed
Add `overload-protection` to FE to block express from trying to handle too many requests

### Why did it change
Node can put itself into a blocked state bc it has accepted too many requests. Instead, we want the to reject with a 503
### Issue tracking

- [PYIC-7616](https://govukverify.atlassian.net/browse/PYIC-7616)

[PYIC-7616]: https://govukverify.atlassian.net/browse/PYIC-7616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app.ts
```diff
@@ -70,6 +71,19 @@ const sessionStore: SessionStore | undefined =
...
 
+const protectConfig: ProtectionConfig = {
+  production: process.env.NODE_ENV === "production",
+  clientRetrySecs: 1,
+  sampleInterval: 5,
+  maxEventLoopDelay: 400,
+  maxHeapUsedBytes: 0,
+  maxRssBytes: 0,
+  errorPropagationMode: false,
+  logging: "info"
```
thebauSoftwire: There's an extra `logStatsOnReq` that I didn't add here. It was failing bc of type errors (for some reason it could only be `false | undefined`. Not sure how to resolve this other than maybe declaring a custom type file?

thebauSoftwire: There is also a warning with setting [this config to true](https://www.npmjs.com/package/overload-protection#logstatsonreq-false) that it could potentially increase pressure on the event loop

Joe-Edwards-GDS: I don't think we should set `logStatsOnReq` - that would be very noisy and limited use. The `frontend-vital-signs` thing should capture relevant metrics instead 🤞 

(If you do need to override the library type definitions there are a couple of options depending on how it's been implemented - mainly through 'declaration merging'. You can see a couple of examples at the top of `app.ts`)

thebauSoftwire: oh good point, didn't think of those but yeah I'll leave it out 👍

src/app.ts
```diff
@@ -70,6 +71,19 @@ const sessionStore: SessionStore | undefined =
...
 
+const protectConfig: ProtectionConfig = {
+  production: process.env.NODE_ENV === "production",
+  clientRetrySecs: 1,
+  sampleInterval: 5,
+  maxEventLoopDelay: 400,
+  maxHeapUsedBytes: 0,
+  maxRssBytes: 0,
+  errorPropagationMode: false,
+  logging: "info",
```
Joe-Edwards-GDS: I think this should probably be error?

Joe-Edwards-GDS (commented):
Looks sensible, but I think the log should be an error, since it's returning one?

Joe-Edwards-GDS (approved):

PR #1648, opened 2024-11-12
BAU: Require ts-node/register in .mocharc.yml

MikeCollingwood:
### What changed

- Require ts-node/register for mocha to run unit tests in IntelliJ

### Why did it change

- So we can run unit tests in IntelliJ

### DIFF
--- a/.mocharc.yml
+++ b/.mocharc.yml
@@ -1,4 +1,6 @@
 fail-zero: false
+require:
+  - ts-node/register
 spec:
   - 'src/**/*.test.ts'
 ignore:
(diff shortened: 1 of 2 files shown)


.mocharc.yml
```diff
@@ -1,4 +1,6 @@
 fail-zero: false
+require:
+  - ts-node/register
```
Joe-Edwards-GDS, line 3: Does this mean we can remove it from the `package.json` test commands?

MikeCollingwood: Yes it does! Will update `package.json`

AmritSidhu (approved):

PR #1653, opened 2024-11-14
PYIC-6245: Use clientOAuthSessionId returned from core-back.

DanCorderIPV:
### What changed

Handling of clientOAuthSessionId for cross-browser callbacks

### Why did it change

As part of the new strategic app routing.

### Issue tracking

- [PYIC-6245](https://govukverify.atlassian.net/browse/PYIC-6245)

[PYIC-6245]: https://govukverify.atlassian.net/browse/PYIC-6245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/mobile-app/middleware.ts
+++ b/src/app/mobile-app/middleware.ts
@@ -19,6 +19,7 @@ export const checkMobileAppDetails: RequestHandler = async (req, res) => {
 
   // If we don't have a clientOAuthSessionId in our session then we're dealing with a cross browser callback and core back will give us a clientOAuthSessionId to use instead.
   if (
+    !req.session.ipvSessionId &&
     !req.session.clientOauthSessionId &&
     apiResponse.data.clientOAuthSessionId
   ) {
--- a/src/app/mobile-app/middleware.ts
+++ b/src/app/mobile-app/middleware.ts
@@ -17,5 +17,13 @@ export const checkMobileAppDetails: RequestHandler = async (req, res) => {
     res.status(apiResponse.status);
   }
 
+  // If we don't have a clientOAuthSessionId in our session then we're dealing with a cross browser callback and core back will give us a clientOAuthSessionId to use instead.
+  if (
+    !req.session.clientOauthSessionId &&
+    apiResponse.data.clientOAuthSessionId
+  ) {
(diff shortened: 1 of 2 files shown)


src/app/mobile-app/middleware.ts
```diff
@@ -17,5 +17,13 @@ export const checkMobileAppDetails: RequestHandler = async (req, res) => {
     res.status(apiResponse.status);
   }
 
+  // If we don't have a clientOAuthSessionId in our session then we're dealing with a cross browser callback and core back will give us a clientOAuthSessionId to use instead.
```
DanCorderIPV, line 20: I'm not very happy about having to have this special case handling in core-front.
It might be better to find a way of having the app callback be handled as a route in core-back so that this can be hidden from core-front. However that would mean changes to journey handling to cope with the cross-browser case having no IpvSession.

DanCorderIPV: @blakeyp do you have any thoughts about this? Is it worth the time to remove this?
I think it would require changes to core-back to:
- Accept a parameter to a journey request (for the state)
- Cope with the lack of IPV session - at least for this journey

It would also have the advantage of making this special case handling visible in the journey map rather than hidden 'magic'

blakeyp: I think it's definitely worth exploring, could you raise a ticket or spike? It might be something we'd do post go-live as an improvement

src/app/mobile-app/middleware.ts
```diff
@@ -17,5 +17,13 @@ export const checkMobileAppDetails: RequestHandler = async (req, res) => {
...
+  if (
+    !req.session.clientOauthSessionId &&
```
blakeyp, line 23: should this be if we don't have an ipvSessionId rather than clientOauthSessionId ?

DanCorderIPV: I'll add that as well. I'd like to keep this check so there's no way we can accidentally overwrite a clientOAuthSessionId

blakeyp (approved):

PR #1663, opened 2024-11-21
PYIC-7609: export dynatrace host id in dockerfile

thebauSoftwire:
### What changed
The default DYNATRACE_HOST_ID in ECS is set to the same value across all containers, so we only get an aggregate view of performance in Dynatrace. To fix this, we export `DT_HOST_ID` with a unique value

### Why did it change
exports unique dynatrace host id in dockerfile

### Issue tracking

- [PYIC-7609](https://govukverify.atlassian.net/browse/PYIC-7609)

[PYIC-7609]: https://govukverify.atlassian.net/browse/PYIC-7609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,8 +42,7 @@ COPY --chown=appuser:appgroup --from=builder /app/package-lock.json ./
 COPY --from=khw46367.live.dynatrace.com/linux/oneagent-codemodules-musl:nodejs / /
 ENV LD_PRELOAD /opt/dynatrace/oneagent/agent/lib64/liboneagentproc.so
 
-ENV DT_HOST_ID="CORE-FRONT-`openssl rand -base64 12`"
 ENV PORT 8080
 EXPOSE 8080
 
-ENTRYPOINT ["sh", "-c", "export DT_HOST_ID=$DT_HOST_ID && tini npm start"]
+ENTRYPOINT ["sh", "-c", "export DT_HOST_ID=CORE-FRONT-$RANDOM && tini npm start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,4 @@ ENV LD_PRELOAD /opt/dynatrace/oneagent/agent/lib64/liboneagentproc.so
 ENV PORT 8080
 EXPOSE 8080
 
-ENTRYPOINT ["tini", "--"]
-
-CMD ["npm", "start"]
+ENTRYPOINT ["sh", "-c", "export DT_HOST_ID=CORE-FRONT-$RANDOM && tini npm start"]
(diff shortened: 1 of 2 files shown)


shivanshuit914 (dismissed):

Dockerfile
```diff
@@ -42,9 +42,8 @@ COPY --chown=appuser:appgroup --from=builder /app/package-lock.json ./
...
 ENV LD_PRELOAD /opt/dynatrace/oneagent/agent/lib64/liboneagentproc.so
 
+ENV DT_HOST_ID="CORE-FRONT-`openssl rand -base64 12`"
 ENV PORT 8080
 EXPOSE 8080
 
-ENTRYPOINT ["tini", "--"]
-
-CMD ["npm", "start"]
+ENTRYPOINT ["sh", "-c", "export DT_HOST_ID=$DT_HOST_ID && tini npm start"]
```
Joe-Edwards-GDS: Why is this one different to the `dev-deploy` version? Where did the openssl bit come from?

thebauSoftwire: Oh, I must have forgotten to update the dev-deploy one too. I took the openssl bit from [here](https://github.com/govuk-one-login/ipv-core-front/blob/1c0471ae87723dc44787b93b1a67a16c16923b05/Dockerfile#L31). I know in the ticket it looks more like this: `export DT_HOST_ID=CORE-FRONT-$RANDOM` but I thought $RANDOM was an env variable for a randomly generated string which I think is what that openssl bit is doing. 

I couldn't find any docs which said that `$RANDOM` was anything special for DT_HOST_ID. I couldn't test this (except that making sure the container is still able to be built + deployed with dev-deploy) in dynatrace for dev though...

Joe-Edwards-GDS: `$RANDOM` is a bash thing, which generates a random number (try `echo $RANDOM` a few times in your terminal)

The uncertainty on the ticket is over whether this will actually work in the docker container's shell (it depends what shell it's running...). Using openssl rand is an alternative, but again - I'm not sure whether the backtick expansion will actually work in that context!

I would suggest following what's on the ticket (i.e. `$RANDOM` and no openssl), so we match everyone else. If it doesn't work then they'll need to update the tickets for every other team as well!

thebauSoftwire: ooo didn't know that, thanks!

Joe-Edwards-GDS: I didn't know until this ticket!

Joe-Edwards-GDS (approved):

PR #1733, opened 2024-12-06
PYIC-7819: update content for ipv success page

shivanshuit914:
## Proposed changes

- Update content on IPV core success screen on P2 journeys and COI

### What changed

- Updated the content for the  IPV core success screen on P2 journeys and COI

### Issue tracking

- [PYIC-7819](https://govukverify.atlassian.net/browse/PYIC-7819)

[PYIC-7819]: https://govukverify.atlassian.net/browse/PYIC-7819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


locales/en/translation.json
```diff
@@ -959,6 +959,7 @@
       "header": "Continue to the service you need to use",
       "content": {
         "paragraph1": "We’ve checked your details. You can now continue to the service you need to use.",
+        "paragraph1RepeatFraudCheck": "You’ve successfully proved your identity. You can now continue to the service you want to use.",
```
Joe-Edwards-GDS: I think this is the wrong way round?

There should be two versions of this content:
- If you've proved a new identity, it should say 'You've successfully proved your identity'
- If you're updating/6MFCing an existing identity, it should say 'We've checked your details'

If you want to leave the default as the existing content then the new context is probably something like `newIdentity`

locales/en/translation.json
```diff
@@ -958,7 +958,8 @@
       "title": "Continue to the service you need to use",
       "header": "Continue to the service you need to use",
       "content": {
-        "paragraph1": "We’ve checked your details. You can now continue to the service you need to use.",
+        "paragraph1": "You’ve successfully proved your identity. You can now continue to the service you want to use.",
+        "paragraph1RepeatFraudCheck": "We’ve checked your details. You can now continue to the service you need to use.",
```
Joe-Edwards-GDS, line 963: Are you sure this is only for repeat fraud checks? Are there any other cases, e.g. update name/address?

shivanshuit914: Thanks Joe, I have created PR to add context for the update name, we already have context for the address 
https://github.com/govuk-one-login/ipv-core-back/pull/2780

Joe-Edwards-GDS: What I mean is, why are we using `repeatFraudCheck` as a context for update name/address journeys - shouldn't we use something like `updateIdentity` or something that captures what the criterion really is?

shivanshuit914: Ok, I thought to use existing context, I will update it with more scenario specific context. Thanks :)

Joe-Edwards-GDS: Ah, I see - it already existed (even though it didn't do anything!)

Thanks, I think updating it will still be better overall :)

Joe-Edwards-GDS: I kinda meant using `updateIdentity` for both - since repeat fraud check is also updating your identity (with a fresh fraud check)

I'm happy for you to keep them separate now you've got this, if you prefer.

Joe-Edwards-GDS (approved):

PR #1739, opened 2024-12-10
PYIC-7532: Strategic app another way page

AmritSidhu:
### What changed

For when the Strategic App goes live. The user may want to exit or prove their identity another way when given the option across the pages on the international address journey.

Subsequent PR for routing: https://github.com/govuk-one-login/ipv-core-back/pull/2787

### Issue tracking

- [PYIC-7532](https://govukverify.atlassian.net/browse/PYIC-7532)

[PYIC-7532]: https://govukverify.atlassian.net/browse/PYIC-7532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


locales/cy/translation.json
```diff
@@ -491,6 +491,23 @@
...
       }
     },
+    "pyiTriageAnotherWay": {
+      "title": "You need to use the GOV.UK One Login app to prove your identity",
+      "header": "You need to use the GOV.UK One Login app to prove your identity",
+      "content": {
+        "paragraph1": "If you do not live in the UK, you can only use the app and a smartphone to prove your identity with GOV.UK One Login.",
+        "subHeading1": "What would you like to do?",
+        "formRadioButtons": {
+          "continueToServiceButtonText": "Go to the service you were trying to use to look for other ways to access it",
```
DanCorderIPV, line 501: Figma has `Continue` instead of `Go`

locales/cy/translation.json
DanCorderIPV, line 1: I assume that Welsh translations will be in another ticket?

views/ipv/page/pyi-triage-another-way.njk
```diff
@@ -0,0 +1,47 @@
...
+    {% set radiosConfig = {
+      idPrefix: "journey",
+      name: "journey",
+      items: [
+        {
+          value: "returnToRp",
+          text: 'pages.pyiTriageAnotherWay.content.formRadioButtons.continueToServiceButtonText' | translate
+        },
+        {
+          value: "back",
```
DanCorderIPV: I suspect using `back` here will hit the automatic back functionality. Safer to use something like `useApp`

DanCorderIPV (changes requested):

views/ipv/page/pyi-triage-another-way.njk
AmritSidhu: I believe it would only automatically use the back functionality if we dont specifically define back step. These actions are going back to previous page so does make sense to be called back. The existing buffer page also uses the keyword back so keep consistency with that. Happy to change it if you still think we should?

locales/cy/translation.json
AmritSidhu: The figma actually uses `Go`, the screenshot which tends to get outdated quick uses continue. So I think `Go` is correct here

views/ipv/page/pyi-triage-another-way.njk
AmritSidhu: After more thought, if we decide to add a back button to these pages its best to avoid this now, I will make the change

MikeCollingwood (approved):
LGTM

DanCorderIPV (approved):

PR #1754, opened 2024-12-17
PYIC-7279: display page context for dev/all-templates page

thebauSoftwire:
### What changed
- display relevant context for pages on `dev/all-templates` screen
- add error handling when page isn't selected and a context option isn't chosen

|Context | Error |
|-|-|
| Screenshot 2024-12-17 at 08 48 08 | Screenshot 2024-12-17 at 08 47 36 |

### Why did it change
Makes it easier to select a context (from a known set) when using the dev page

### Issue tracking

- [PYIC-7279](https://govukverify.atlassian.net/browse/PYIC-7279)

[PYIC-7279]: https://govukverify.atlassian.net/browse/PYIC-7279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/development/middleware.ts
```diff
@@ -13,36 +13,78 @@ import { getAppStoreRedirectUrl } from "../shared/appDownloadHelper";
...
+  RadioOption[]
+> = {};
 
 export const allTemplatesGet: RequestHandler = async (req, res) => {
   const directoryPath = path.resolve("views/ipv/page");
 
   // Load available templates and convert into radio option objects for the GOV.UK Design System nunjucks template
-  if (!config.TEMPLATE_CACHING || !templateRadioOptions) {
+  if (!config.TEMPLATE_CACHING || !templates) {
     const templateFiles = await fs.readdir(directoryPath);
```
Joe-Edwards-GDS: Do you think we actually need to read this from the filesystem any more?

It looks like we'll throw an error if we find a page that isn't present in `pages-and-contexts`, so we could just use that as the source of truth?

That should simplify things as well, since we don't need caching if it's just a JS module.

src/app/development/middleware.ts
```diff
@@ -13,36 +13,78 @@ import { getAppStoreRedirectUrl } from "../shared/appDownloadHelper";
...
+    if (!pagesAndContexts[page]) {
+      throw new Error(
+        `Page ${page} does not exist in the template and context mapping.`,
+      );
+    }
+
+    templatesWithContextRadioOptions[page] = pagesAndContexts[page].map(
+      (context) => ({
+        text: context || "No context",
+        value: context || "",
```
Joe-Edwards-GDS: Prefer `??` if you're detecting null/undefined (it would be pretty silly to have `""` as a valid context, but still)

src/app/development/middleware.ts
```diff
@@ -13,36 +13,78 @@ import { getAppStoreRedirectUrl } from "../shared/appDownloadHelper";
...
   }
 
   res.render(getTemplatePath("development", "all-templates"), {
-    templateRadioOptions: templateRadioOptions,
+    templatesWithContextRadioOptions: templatesWithContextRadioOptions,
     csrfToken: req.csrfToken?.(true),
   });
 };
 
+export const checkRequiredOptionsAreSelected: RequestHandler = (
```
Joe-Edwards-GDS: If simplifies the logic, I'd also be happy to put this into the `allTemplatesPost` handler - I'm not sure how I feel about this approach of breaking everything into tiny middlewares

views/development/all-templates.njk
```diff
@@ -6,10 +6,34 @@
...
+        html: govukRadios({idPrefix: "pageContext",
+        name: "pageContext",
+        classes: "govuk-radios--small",
+        fieldset: {
+            legend: {
+            text: 'Choose a context:',
+            classes: "govuk-fieldset__legend--s"
+            }
+        },
+        items: contexts})} if contexts|length else undefined %}
```
Joe-Edwards-GDS: The indentation here is a bit funky - can we lay it out like we would JS?

views/development/all-templates.njk
```diff
@@ -6,10 +6,34 @@
...
+{% endif %}
+
 {% block content %}
   <form id="pageSelectForm" action="/dev/all-templates" method="POST">
     <input type="hidden" name="_csrf" value="{{ csrfToken }}">
 
+    {% set templateRadioConfig = [] %}
+    {% for page, contexts in templatesWithContextRadioOptions %}
+      {% set conditional = {
+        html: govukRadios({idPrefix: "pageContext",
```
Joe-Edwards-GDS: Looks like the idPrefix isn't enough to make the context radios unique, which is breaking a couple of things, e.g. being able to click the label to select the button.

You might need to add the page id into this prefix as well

src/test-utils/pages-and-contexts.ts
```diff
@@ -0,0 +1,82 @@
...
+  "live-in-uk": [],
+  "no-photo-id-abandon-find-another-way": [],
+  "no-photo-id-exit-find-another-way": [],
+  "no-photo-id-security-questions-find-another-way": ["dropout", undefined],
+  "non-uk-no-app-options": [],
+  "non-uk-app-intro": [],
+  "non-uk-no-app": [],
+  "non-uk-no-passport": [],
+  "non-uk-passport": [],
+  "page-dcmaw-success": ["coiNoAddress"],
```
Joe-Edwards-GDS: Not related to your change (so feel free to do separately), but should this have `undefined` as well?

src/test-utils/pages-and-contexts.ts
Joe-Edwards-GDS, line 1: One thing to consider as a follow-up, I wonder if we can group these in some way?

The templates page is quite long now, and you have to do a lot of scrolling to get to the submit button 😅

Joe-Edwards-GDS (commented):
Looks good, I think we can make a few small enhancements...

src/app/development/middleware.ts
thebauSoftwire: yeah I kept it in for added safety for the dev to know w/out having to run the browser tests but that's a fair point! And makes it simpler, thanks :)

src/test-utils/pages-and-contexts.ts
thebauSoftwire: nice spot!

src/test-utils/pages-and-contexts.ts
thebauSoftwire: yeah I had a bit of think. I think the simplest way of reducing the scrolling would be to use something like the existing [tabs](https://design-system.service.gov.uk/components/tabs/) component or maybe the [accordion](https://design-system.service.gov.uk/components/accordion/) (though there might still be quite a bit of vertical scrolling for this depending on how many groups there are). The biggest problem with these would then be that it'll be harder to do something like cmd+f to search for a page quickly - instead having to go through each tab if you didn't know which group it belonged to.

The ideal solution to tackle both problems would be to have something like a search bar - not sure how this would be done with nunjucks though. 

Or we could force the page to scroll to the next section when a template (and context if applicable) have been selected. Again, not sure how this would work with nunjucks.

DanCorderIPV: If the scrolling is to get to the submit button, then maybe we could have a floating submit that's always on screen - this is a dev page so we can rely on JS if we want to

thebauSoftwire: ooo yeah that sounds good too

Joe-Edwards-GDS: Yeah, I think a dedicated search probably isn't that much better than browser search, but some way of quickly submitting would be nice.

Floating submit could work (but hard to make nice!) - or maybe some way of doing it inline, but can't quite picture it...

In any case, happy to leave this as a separate PR

Joe-Edwards-GDS (approved):
Noice

I notice Joe U has had a look as well - might be worth getting his input as a sign-off, but this is a definitely better than what we have!

PR #1761, opened 2024-12-19
PYIC-7869: Spike - render service-unavailable page on 503

Wynndow:
### What changed

Spike - render service-unavailable page on 503

### Why did it change

The overload protection lib doesn't seem to be configurable to specify a template or html page to return when it kicks in. We can however set it to hand control back to the framework and handle a 503 ourselves.

The downside of this is that it will require the app to do work when it's already overloaded.

### Issue tracking

- [PYIC-7869](https://govukverify.atlassian.net/browse/PYIC-7869)

[PYIC-7869]: https://govukverify.atlassian.net/browse/PYIC-7869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/internal-server-error-handler.ts
+++ b/src/handlers/internal-server-error-handler.ts
@@ -3,7 +3,7 @@ import { isAxiosError } from "axios";
 import { ErrorRequestHandler } from "express";
 import { isHttpError } from "http-errors";
 import PAGES from "../constants/ipv-pages";
-import { getIpvPageTemplatePath, getErrorPageTemplatePath } from "../lib/paths";
+import {getIpvPageTemplatePath, getErrorPageTemplatePath, getHtmlPath} from "../lib/paths";
 import ERROR_PAGES from "../constants/error-pages";
 import HttpError from "../errors/http-error";
 import { HANDLED_ERROR } from "../lib/logger";
@@ -29,12 +29,16 @@ const serverErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
     return next(err);
   }
 
+  if (res.statusCode === 503) {
+      return res.render(getHtmlPath("errors", "service-unavailable-s3", ERROR_PAGES.SERVICE_UNAVAILABLE))
+  }
+
   const status = getErrorStatus(err);
 
   req.log?.error({
     message: {
       description: err?.constructor?.name ?? "Unknown error",
-      errorMessage: err?.message ?? "Unkown error",
+      errorMessage: err?.message ?? "Unknown error",
       errorStack: err?.stack,
     },
   });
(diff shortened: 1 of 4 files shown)


src/handlers/internal-server-error-handler.ts
```diff
@@ -29,12 +29,16 @@ const serverErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
     return next(err);
   }
 
+  if (res.statusCode === 503) {
+      return res.render(getHtmlPath("errors", "service-unavailable-s3", ERROR_PAGES.SERVICE_UNAVAILABLE))
```
Joe-Edwards-GDS, line 33: Ah cool - I guess it's hard to tell whether routing through the error handlers will be substantially slower than the 'basic' response, but it feels like a worthwhile hit.

Joe-Edwards-GDS, line 33: I didn't even realise we had this 503 page..! Perhaps once we get designs etc., we can build the page statically during the build process - something we'll worry about later.

Joe-Edwards-GDS, line 33: Do you know if this will cache the file or load it from disk each time? That's something we probably want to watch out for when implementing for real

Wynndow: By the look of it Express will cache views in production by default

Wynndow: Which is interesting because it implies that lower envs won't cache the views. Which might not help the results of the perf tests.

Wynndow: http://expressjs.com/en/5x/api.html#res.render

Joe-Edwards-GDS: Thanks - I suspect it's looking for NODE_ENV=production, which should always be the case in deployed environments. :)

I wonder if we should be using `res.send` or `res.sendFile` instead? I'm not totally sure what the view engine will do with raw HTML 🤔 - presumably it works though!

Wynndow: Ah you're right - i just looked and it is production across the envs.

And yeah render works fine - but I guess there might be some perf gains for using send... I'll have a quick look

Wynndow: So it looks like sendFile _won't_ cache the file in memory - it'll read it from disk each time. So render might actually be a better choice, unless you want to introduce some other sort of caching.

Although I notice that we're globally setting `Cache-control private no-store` (among others) on all responses. Maybe if we allow caching, cloudfront will handle it for us?

Joe-Edwards-GDS: Ah - but we probably don't want CloudFront caching `/page/ipv-success-page` as a 503 error screen

If render 'just works' then that might be easiest. Otherwise we can load the file on startup and serve it from memory.

PR #1779, opened 2025-01-08
PYIC-7910 Preload templates for all pages

Joe-Edwards-GDS:
### What changed

Pre-render all pages on startup, to ensure they are cached before serving requests

### Why did it change

We see a high rate of 503s when spinning up new containers during a load test, suggesting that new containers are seeing a high event loop latency.

One possible explanation is that this is occurring when loading uncached templates from disk (as Nunjucks does this synchronously). By preloading the templates, we should be able to keep the event loop latency low once serving requests.

### Issue tracking

- [PYIC-7910](https://govukverify.atlassian.net/browse/PYIC-7910)

[PYIC-7910]: https://govukverify.atlassian.net/browse/PYIC-7910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS:
This doesn't appear to have had the desired effect, so parking this for now

PR #1787, opened 2025-01-13
BAU: Support perf environment

Joe-Edwards-GDS:
### What changed

Some of these changes already exist in https://github.com/govuk-one-login/ipv-core-front/pull/1507 but these are unmerged and out of date.

### Why did it change

We want to be able to deploy any branch to perf without manually adjusting the cloudformation template

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -77,7 +77,6 @@ Mappings:
       tg500ErrorLimit: 10
       tg500ErrorWindow: 300
       lambdaInvokeCompareWindow: 300
-      nodeEnv: "production"
       languageToggle: true
       ga4Disabled: false
       uaDisabled: false
@@ -94,7 +93,6 @@ Mappings:
       tg500ErrorLimit: 10
       tg500ErrorWindow: 300
       lambdaInvokeCompareWindow: 300
-      nodeEnv: "production"
       languageToggle: true
       ga4Disabled: false
       uaDisabled: false
@@ -111,7 +109,6 @@ Mappings:
       tg500ErrorLimit: 10
       tg500ErrorWindow: 300
       lambdaInvokeCompareWindow: 300
-      nodeEnv: "production"
       languageToggle: true
       ga4Disabled: false
       uaDisabled: false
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -74,10 +77,7 @@ Mappings:
       tg500ErrorLimit: 10
       tg500ErrorWindow: 300
       lambdaInvokeCompareWindow: 300
-      fargateCPUsize: "256"
-      fargateRAMsize: "512"
-      nodeOldSpaceLimit: "384"
-      nodeEnv: "development"
+      nodeEnv: "production"
```
Joe-Edwards-GDS: I can't see any good reason for this to be `development` during normal workflow - it's probably more useful to have it include production optimisations.

deploy/template.yaml
```diff
@@ -137,9 +128,6 @@ Mappings:
       tg500ErrorLimit: 10
       tg500ErrorWindow: 300
       lambdaInvokeCompareWindow: 300
-      fargateCPUsize: "256"
-      fargateRAMsize: "512"
-      nodeOldSpaceLimit: "256"
```
Joe-Edwards-GDS, line 142: This is the biggest change - staging and integration will be bumped to match build and prod. It will slightly increase costs, but it feels less surprising than the current approach.

For reference, in staging ECS is about $40/month, and this will increase it ~4x. (Integration is already halfway between, so will increase ~2x).

deploy/template.yaml
```diff
@@ -786,9 +768,9 @@ Resources:
               Value: !Join
                 - ''
                 - - '--max-old-space-size='
-                  - !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, nodeOldSpaceLimit ]
+                  - !If [ IsPerformanceSensitive, 1792, 384 ]
             - Name: NODE_ENV
-              Value: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, nodeEnv ]
+              Value: "production"
```
blakeyp, line 767: don't need `nodeEnv` in the mappings anymore?

Joe-Edwards-GDS: Ah - I hardcoded it here and then flip-flopped about whether to leave it in the mapping (ending up with neither one or the other...)

Do you have a preference? I'll remove from the mapping unless you think otherwise.

blakeyp: Fine to remove yeah

blakeyp (approved):

PR #1806, opened 2025-01-22
PYIC-7933: Update content on F2F handoff (booking confirmation) page

shivanshuit914:
## Proposed changes

- Updated content for the f2f handoff Page

### What changed

- Content changes and uses dynamic date to visit Post Office

### Issue tracking

- [PYIC-7933](https://govukverify.atlassian.net/browse/PYIC-7933)

[PYIC-7933]: https://govukverify.atlassian.net/browse/PYIC-7933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/ipv/middleware.ts
```diff
@@ -357,6 +357,10 @@ export const handleJourneyPageRequest = async (
     } else if (pageId === PAGES.PYI_TRIAGE_MOBILE_DOWNLOAD_APP) {
       validatePhoneType(context);
       renderOptions.appDownloadUrl = getAppStoreRedirectUrl(context);
+    } else if (pageId === PAGES.PAGE_FACE_TO_FACE_HANDOFF) {
+      renderOptions.postOfficeVisitByDate = new Date().setDate(
+        new Date().getDate() + 15,
+      );
```
MikeCollingwood, line 363: The snapshots are always going to be updated with this...

src/app/development/middleware.ts
```diff
@@ -104,6 +103,7 @@ export const templatesDisplayGet: RequestHandler = async (req, res) => {
     );
   }
   const phoneType = context ? (context as string) : undefined;
+  /* istanbul ignore else  */
```
MikeCollingwood, line 116: Is this to ignore for code coverage? Think it's valuable to test this code

shivanshuit914: This is the development middleware that we use for the snapshot testing and all templates url :)

src/app/ipv/middleware.ts
shivanshuit914: It will use the date from the development middleware, that I have hardcoded so the snapshot will be same :) 


```
if (process.env.NODE_ENV === "local") {
    today = new Date("2025-01-01");
}
```

src/app/development/middleware.ts
MikeCollingwood: Oh apologies, though still think we want testing of the ipv/middleware code update

src/app/ipv/middleware.ts
MikeCollingwood: Cool, I understand now :)

MikeCollingwood (dismissed):
LGTM though would be good to add some testing for ipv/middleware

src/app/ipv/middleware.ts
```diff
@@ -357,6 +357,10 @@ export const handleJourneyPageRequest = async (
     } else if (pageId === PAGES.PYI_TRIAGE_MOBILE_DOWNLOAD_APP) {
       validatePhoneType(context);
       renderOptions.appDownloadUrl = getAppStoreRedirectUrl(context);
+    } else if (pageId === PAGES.PAGE_FACE_TO_FACE_HANDOFF) {
+      renderOptions.postOfficeVisitByDate = new Date().setDate(
+        new Date().getDate() + 15,
```
blakeyp: We probably want to pull `15` out into config - it doesn't need to be an env var but we could define it in `config.ts` perhaps?

src/app/development/middleware.ts
```diff
@@ -112,6 +112,14 @@ export const templatesDisplayGet: RequestHandler = async (req, res) => {
...
     validatePhoneType(phoneType);
     renderOptions.appDownloadUrl = getAppStoreRedirectUrl(phoneType);
+  } else if (templateId === PAGES.PAGE_FACE_TO_FACE_HANDOFF) {
+    let today = new Date();
+    if (process.env.NODE_ENV === "local") {
+      today = new Date("2025-01-01");
+    }
+    renderOptions.postOfficeVisitByDate = new Date().setDate(
+      today.getDate() + 15,
+    );
```
blakeyp: could we have this just always hardcode the date (rather than relying on NODE_ENV being set to local) ?
should be fine for just a preview of the page

blakeyp (commented):
Looks good, HOLD until we get the green light

blakeyp (approved):

PR #1814, opened 2025-01-27
PYIC-7964: Reinstate core-front backend response logging

AmritSidhu:
### What changed

Reinstate core-front backend response logging
Exclude data logging for the proven-identity-details request only

### Why did it change

In core-front we were logging the response data we get back for all core-back requests. This is useful for debugging as it includes for example CRI ids, redirect URIs and page/journey responses. Unfortunately we discovered we were inadvertently logging PII (name/dob/address) for the proven-identity-details request, and the quick incident fix was just to drop the data property for all API requested completed logs

### Issue tracking

- [PYIC-7964](https://govukverify.atlassian.net/browse/PYIC-7964)

[PYIC-7964]: https://govukverify.atlassian.net/browse/PYIC-7964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/shared/axiosHelper.ts
+++ b/src/app/shared/axiosHelper.ts
@@ -12,7 +12,6 @@ declare module "axios" {
     startTime?: number;
   }
 }
-
 interface RequestLog {
   description: string;
   endpoint: string;
@@ -20,7 +19,6 @@ interface RequestLog {
   cri?: string;
   duration?: number;
 }
-
 const extractCredentialIssuerId = (
   response: AxiosResponse,
 ): string | undefined => {
@@ -35,11 +33,18 @@ const extractCredentialIssuerId = (
   return undefined;
 };
 
-// Client/CRI redirect URLs may contain sensitive data, and we shouldn't log them
 const sanitiseResponseData = (response: AxiosResponse): object | undefined => {
   try {
     if (typeof response.data === "object") {
       const body = { ...response.data };
+
+      const endpoint = response.config?.url ?? "";
+      if (endpoint.includes("proven-identity-details")) {
+        // Completely  data logging for this endpoint
+        return undefined;
+      }
+
+      // Sanitize redirect URLs for other endpoints
       if (body.cri?.redirectUrl) {
         body.cri = {
           ...body.cri,
@@ -56,7 +61,7 @@ const sanitiseResponseData = (response: AxiosResponse): object | undefined => {
     }
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
   } catch (err) {
-    // Ignore
+    // Safely catch and ignore errors
   }
   return undefined;
 };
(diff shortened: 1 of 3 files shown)


src/app/shared/axiosHelper.ts
```diff
@@ -34,37 +33,70 @@ const extractCredentialIssuerId = (
   return undefined;
 };
 
+const sanitiseResponseData = (response: AxiosResponse): object | undefined => {
+  try {
+    if (typeof response.data === "object") {
+      const body = { ...response.data };
+
+      const endpoint = response.config?.url ?? "";
+      if (endpoint.includes("proven-identity-details")) {
```
blakeyp: Could we make this an allow list rather than deny list ie. specify which requests we should allow data logging for (which I think should be all except proven-identity-details)? so we don't inadvertently start leaking PII if we add new requests in the future

Also could we pull the values from config eg. `API_SESSION_INITIALISE` rather than hardcoding?

src/app/shared/axiosHelper.ts
```diff
@@ -12,13 +13,25 @@ declare module "axios" {
...
   endpoint: string;
   data?: object;
   cri?: string;
   duration?: number;
 }
+
+// Helper function to determine if an endpoint is in the allow list
+const isEndpointAllowedForDataLogging = (url: string): boolean => {
+  const ALLOWED_ENDPOINTS = Object.values(API_URLS).filter(
+    (endpoint) => endpoint !== API_URLS.API_BUILD_PROVEN_USER_IDENTITY_DETAILS,
```
blakeyp: Hmm to me this is still a deny list as we're allowing anything except this request - I think we want to explicitly define the allowed requests? so that if we introduce a new API endpoint in the future, we need to explicitly add it to the allowlist here - otherwise we may start calling a new endpoint that logs PII without us realising

src/constants/api-constants.ts
```diff
@@ -0,0 +1,9 @@
+const API_URLS = Object.freeze({
+  API_CRI_CALLBACK: "/cri/callback",
+  API_MOBILE_APP_CALLBACK: "/app/callback",
+  API_JOURNEY_EVENT: "/journey",
+  API_SESSION_INITIALISE: "/session/initialise",
+  API_BUILD_PROVEN_USER_IDENTITY_DETAILS: "/user/proven-identity-details",
+});
+
```
blakeyp: could we pull these directly from config.ts rather than having to redefine them here?

src/app/shared/axiosHelper.ts
```diff
@@ -22,9 +22,18 @@ interface RequestLog {
   duration?: number;
 }
 
+const ALLOWED_ENDPOINTS: string[] = [
```
blakeyp: could name this is a bit more explicitly eg. `ALLOWED_RESPONSE_DATA_LOGGING_ENDPOINTS` ?

blakeyp (approved):

PR #1822, opened 2025-01-31
Pyic 4434 update govuk frontend

DanCorderIPV:
### What changed

Updated GOVUK Frontend to v5

### Why did it change

To keep up to date

### Issue tracking

- [PYIC-4434](https://govukverify.atlassian.net/browse/PYIC-4434)

[PYIC-4434]: https://govukverify.atlassian.net/browse/PYIC-4434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


views/shared/sign-out-header.njk
```diff
@@ -3,26 +3,19 @@
...
+        </svg>
```
blakeyp: Based on the updated snapshots I think this is the old crown logo?

We probably considered this before but is there any way we can avoid having to override the header ourselves while still adding the sign out link? 🤔

views/errors/service-unavailable-s3/service-unavailable.html
```diff
@@ -48,29 +47,19 @@
...
+          </svg>
```
blakeyp, line 62: Is this also the old crown? (not that it matters so much on this page)

views/development/all-templates.njk
```diff
@@ -87,4 +87,6 @@
       text: "Go"
     }) }}
   </form>
+
+  <a class="govuk-link" href="/dev/template/service-unavailable">Service Unavailable static page</a>
```
blakeyp: Nice 👍 

I did find where this page is actually being referenced - https://github.com/govuk-one-login/ipv-core-common-infra/blob/7a692a8c0f4125713fa83a94227c2ad20fe7534f/cloudformation/shield-monitor/template.yaml#L19 which is used on DNS failover. But the url is very out of date 🤦

assets/scss/application.scss
```diff
@@ -1,5 +1,5 @@
-@use "../../node_modules/govuk-frontend/govuk/all";
-@import "../../node_modules/govuk-frontend/govuk/base";
+@use "../../node_modules/govuk-frontend/dist/govuk/all";
+@import "../../node_modules/govuk-frontend/dist/govuk/base";
 @import "../../node_modules/@govuk-one-login/frontend-language-toggle/build/stylesheet/styles";
 @import "service-header";
```
blakeyp, line 5: The upgrade guide also talks about adding an override for the phase banner, is that something we need too?
https://govukverify.atlassian.net/wiki/spaces/DIFC/pages/4523851778/V5+Upgrade+Guidance+for+the+One-Login+Programme#%E2%9A%A0%EF%B8%8F-Breaking-Change-%235---Tag-and-Phase-Banner

views/shared/sign-out-header.njk
DanCorderIPV: Good point, I'll revert this and below

assets/scss/application.scss
DanCorderIPV: Oh, I missed that, that might actually get rid of a lot of the differences

DanCorderIPV:
I squashed the snapshots and force pushed this to keep the repo size down

blakeyp (dismissed):

blakeyp (approved):

PR #1831, opened 2025-02-06
PYIC-7928: Update dev/all-templates page

AmritSidhu:
### What changed

Used the govUkTable to make information inline
This includes adding a submit button for every template 
This reduces the need to scroll to the bottom every time

Before:
Screenshot 2025-02-06 at 09 37 49

After:
Screenshot 2025-02-06 at 11 32 35

### Issue tracking

- [PYIC-7928](https://govukverify.atlassian.net/browse/PYIC-7928)

[PYIC-7928]: https://govukverify.atlassian.net/browse/PYIC-7928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/views/development/all-templates.njk
+++ b/views/development/all-templates.njk
@@ -2,89 +2,69 @@
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/select/macro.njk" import govukSelect %}
 
 {% extends "shared/base.njk" %}
 {% set pageTitleKey = "All templates" %}
+{% set rowClasses = "govuk-grid-column-full" %} 
 
 {% if errorState %}
     {% set errorText = 'Select a page and a context if appropriate.' %}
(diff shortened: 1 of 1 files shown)


views/development/all-templates.njk
```diff
@@ -2,89 +2,69 @@
...
-    {{ govukButton({
-      text: "Go"
-    }) }}
+          ]
+        }) }}
+      </td>
+      <td class="govuk-table__cell">
+        <div style="display: flex; gap: 10px;">
+          <button name="language" value="en" class="govuk-button">En</button>
+          <button name="language" value="cy" class="govuk-button govuk-button--secondary">Cy</button>
```
thebauSoftwire: Might be a personal preference but on first glance at the page, it's not immediately obvious to me that we have to click on one of the language buttons to view the template. I think this is because the En button is green which makes it look like it's already been selected between the two options.

I'd lean towards just making them both the same colour since there's no need to highlight a "default" option now that there isn't a shared submit button.

thebauSoftwire (dismissed):
Neat! Looks way nicer :) Just one optional comment

thebauSoftwire (approved):

PR #1847, opened 2025-02-19
PYIC-7287: spinner component

shivanshuit914:
## Proposed changes

- Add spinner component for the dcmaw response page
- Add polling endpoint for spinner component

### What changed

- Created spinner with polling endpoint for dcmaw page.

### Issue tracking

- [PYIC-7287](https://govukverify.atlassian.net/browse/PYIC-7287)

[PYIC-7287]: https://govukverify.atlassian.net/browse/PYIC-7287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env.sample
```diff
@@ -8,3 +8,5 @@ DT_RUM_URL=
 SESSION_SECRET=dev_secret
 ENABLE_PREVIEW="development"
 LANGUAGE_TOGGLE=false
+SPINNER_REQUEST_INTERVA=3000
```
blakeyp: `INTERVAL` ?
and could we have these prefixed with `MAM` as we'll want to be able to configure the values for the DAD screen separately

locales/cy/translation.json
```diff
@@ -1357,10 +1357,43 @@
...
+            "messageText": "We cannot prove your identity right now.",
+            "whatYouCanDo": {
+              "heading": "What you can do",
+              "message": {
+                "text1": "Go back to the service you were trying to use and start again. You can look for service using your search engine or find it from the GOV.UK homepage.",
+                "link": {
+                  "href": "",
+                  "text": ""
+                },
+                "text2": ""
```
blakeyp, line 1392: Do we need to fill these?

locales/cy/translation.json
```diff
@@ -1357,10 +1357,43 @@
...
+          "longWait": {
+            "spinnerStateText": "ydym yn dal i wirio’ch manylion. Peidiwch â chau nac adnewyddu’r dudalen hon."
+          },
+          "error": {
+            "heading": "Sorry, there is a problem",
+            "messageText": "We cannot prove your identity right now.",
+            "whatYouCanDo": {
+              "heading": "What you can do",
+              "message": {
+                "text1": "Go back to the service you were trying to use and start again. You can look for service using your search engine or find it from the GOV.UK homepage.",
```
blakeyp: We should already have the translations for these from other error pages?

locales/cy/translation.json
```diff
@@ -1357,10 +1357,43 @@
       }
     },
     "checkMobileAppResult": {
-      "title": "Checking your details",
-      "header": "Checking your details",
-      "content": {
-        "paragraph1": "Polling to see your mobile app results."
+      "title": "Rydym yn gwirio eich manylion",
+      "content": {
+        "paragraph1": "Polling to see your mobile app results.",
```
blakeyp: Remove?

locales/en/translation.json
```diff
@@ -1357,10 +1357,43 @@
       }
     },
     "checkMobileAppResult": {
-      "title": "Checking your details",
-      "header": "Checking your details",
-      "content": {
-        "paragraph1": "Polling to see your mobile app results."
+      "title": "We’re checking your details",
+      "content": {
+        "paragraph1": "Polling to see your mobile app results.",
```
blakeyp: Remove?

src/config/config.ts
```diff
@@ -17,6 +17,7 @@ export default {
   API_JOURNEY_EVENT: "/journey",
   API_SESSION_INITIALISE: "/session/initialise",
   API_BUILD_PROVEN_USER_IDENTITY_DETAILS: "/user/proven-identity-details",
+  API_DCMAW_POLL: "/app/check-vc-receipt",
```
blakeyp: Can we align this with the name of the function ie. `API_CHECK_MOBILE_APP_VC_RECEIPT`

src/app/identity-status/router.ts
```diff
@@ -0,0 +1,8 @@
+import express from "express";
+import { proveIdentityStatusCallbackGet } from "./middleware";
+
+const router = express.Router();
+
+router.get("/prove-identity-status", proveIdentityStatusCallbackGet);
```
blakeyp: We probably want to rename this (and the handler) for our purposes, maybe `app-vc-receipt-status` ?

src/app/identity-status/middleware.ts
```diff
@@ -0,0 +1,27 @@
+import { Request, Response } from "express";
+import { NextFunction, RequestHandler } from "express-serve-static-core";
+import { getDcMawPoll } from "../../services/coreBackService";
+
+enum IdentityProcessingStatus {
+  COMPLETED = "COMPLETED",
+  ERROR = "ERROR",
+  PROCESSING = "PROCESSING",
+  INTERVENTION = "INTERVENTION",
```
blakeyp: Are we reusing auth's statuses because they're hardcoded in the spinner component? 
That's probably ok if we can't avoid it, although I'd probably rename this `AppVcReceiptStatus` and drop `INTERVENTION`

src/services/coreBackService.ts
```diff
@@ -144,3 +144,26 @@ export const getProvenIdentityUserDetails = (
...
+      return true;
+    }
+
+    await axiosInstance.get(
+      config.API_DCMAW_POLL,
+      generateAxiosConfig(
+        `${config.API_BASE_URL}${config.API_DCMAW_POLL}`,
+        req,
+      ),
+    );
```
blakeyp: Might be out of scope of this ticket but I understand this endpoint returns a `journeyType` response which will tell us where to take the user next - we'll need to use this to patch the 'continue' action on the spinner page so the user can continue, which might require storing it in the session or something 🤔

src/services/coreBackService.ts
```diff
@@ -144,3 +144,26 @@ export const getProvenIdentityUserDetails = (
     ),
   );
 };
+
+export const getDcMawPoll = async (req: Request): Promise<boolean> => {
```
blakeyp: Could rename simply `appVcReceived` ?

src/services/coreBackService.ts
```diff
@@ -144,3 +144,26 @@ export const getProvenIdentityUserDetails = (
...
+    }
+
+    await axiosInstance.get(
+      config.API_DCMAW_POLL,
+      generateAxiosConfig(
+        `${config.API_BASE_URL}${config.API_DCMAW_POLL}`,
+        req,
+      ),
+    );
+    return true;
```
blakeyp: To be safe we might want to explicitly check that the status code is exactly 200 as we expect (and not a different success code) ?

blakeyp (commented):
Nice job 🎉 just left mostly naming comments.
I did think for the DAD download page we might need to lift and shift the spinner component entirely into our code given how customised the designs are, but let's cross that bridge when we get to it 🤞.

locales/cy/translation.json
shivanshuit914: Yes, Otherwise it shows undefined with key name :)

locales/cy/translation.json
shivanshuit914: I messed up this while rebasing my branch I will remove it thanks

src/app/identity-status/middleware.ts
shivanshuit914: Yes, Its been used by the spinner.

src/services/coreBackService.ts
shivanshuit914: YEs It returns journeyType response. Hm yes we might need to store in session.

blakeyp: Reading the ticket ACs it sounds like the continue action is meant to be in scope, could you take a look? @MikeCollingwood may be able to help as he originally worked on this endpoint

src/services/coreBackService.test.ts
```diff
@@ -162,4 +165,63 @@ describe("CoreBackService", () => {
       },
     );
   });
+
+  it("should appVcReceived to retrieve user identity details", async () => {
```
blakeyp: probably needs rewording?

src/services/coreBackService.ts
```diff
@@ -144,3 +147,31 @@ export const getProvenIdentityUserDetails = (
...
+    );
+
+    if (isJourneyResponse(response.data)) {
+      req.session.journey = response.data.journey;
+    }
+
+    return response.status === 200;
+  } catch (error) {
+    if (isAxiosError(error) && error.response?.status === 404) {
+      return false;
```
blakeyp: Hmm I think it will be neater if we handle this in the caller i.e. [vc-receipt-status/middleware.ts](https://github.com/govuk-one-login/ipv-core-front/pull/1847/files#diff-30e30dd051e06c268b4f244c482cf4091d2f289fbf413a8d7e51614237389b22).
coreBackService probably shouldn't be setting things in the session as a side effect (it's really just a thin wrapper around the API) so I'd probably have this just return the core back response and handle the session and 404 error in the middleware, if this makes sense?

shivanshuit914: Yes agree thanks

src/app/vc-receipt-status/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+      return;
+    }
+
+    const appVcResponse = await appVcReceived(req);
+
+    if (isJourneyResponse(appVcResponse.data)) {
+      req.session.journey = appVcResponse.data.journey;
+    }
+
+    res.status(200).json({ status: AppVcReceiptStatus.COMPLETED });
```
blakeyp, line 35: for clarity best to add a `return` after this?

src/app/vc-receipt-status/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  } catch (error) {
+    if (isAxiosError(error) && error.response?.status === 404) {
+      res.status(200).json({ status: AppVcReceiptStatus.PROCESSING });
+      return;
+    }
+    res.status(500).json({ status: AppVcReceiptStatus.ERROR });
+    next(error);
```
blakeyp: How will the spinner component handle a 500 response - does it show its dynamic error content? I think if so we probably want to just log the error here and then return res.status(500) rather than calling `next`?

src/constants/ipv-pages.ts
```diff
@@ -50,6 +50,7 @@ const IPV_PAGES = Object.freeze({
   UPDATE_NAME_DATE_BIRTH: "update-name-date-birth",
   WE_MATCHED_YOU_TO_YOUR_ONE_LOGIN: "we-matched-you-to-your-one-login",
   YOU_CAN_CHANGE_SECURITY_CODE_METHOD: "you-can-change-security-code-method",
+  PAGE_CHECK_MOBILE_APP_RESULT: "check-mobile-app-result",
```
MikeCollingwood: I think we have kept this alphabetically-ordered. Also, I think it doesn't necessarily need the `PAGE_` prefix, as we already know it's a page because it's in this object.

src/app/vc-receipt-status/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (config.ENABLE_PREVIEW) {
+      res.status(200).json({ status: AppVcReceiptStatus.COMPLETED });
+      return;
+    }
+
+    const appVcResponse = await appVcReceived(req);
+
+    if (isJourneyResponse(appVcResponse.data)) {
+      req.session.journey = appVcResponse.data.journey;
+    }
```
MikeCollingwood: I think this should be made stricter, if there is no journey response the eventual continue button won't work.

It _should_ return a journey response for a successful response, so I think it should fail if not:

```suggestion
    if (!isJourneyResponse(appVcResponse.data)) {
      throw new Exception("Journey response expected from successful check app vc receipt response.");
    }
    
    req.session.journey = appVcResponse.data.journey;
```

src/app.ts
```diff
@@ -208,6 +210,7 @@ router.use("/oauth2", oauthRouter);
 router.use("/credential-issuer", criRouter);
 router.use("/app", mobileAppRouter);
 router.use("/ipv", ipvRouter);
+router.use("/", vcReceiptStatusRouter);
```
MikeCollingwood, line 213: I think it could make sense to put it under the `/app` umbrella, and move the logic into the mobileAppRouter

src/app/vc-receipt-status/middleware.ts
shivanshuit914: Yes It shows the error content, same as technical error page.

src/app/ipv/tests/handleJourneyActionRequest.test.ts
```diff
@@ -92,6 +92,29 @@ describe("handleJourneyActionRequest", () => {
...
+
+    const res = createResponse();
+
+    // Act
+    await middleware.handleJourneyActionRequest(req, res, next);
+
+    // Assert
+    expect(req.body.journey).to.equal("mobile-app-journey-event");
+  });
+
```
MikeCollingwood, line 117: We probably want a positive test case here as well

blakeyp (approved):

src/services/coreBackService.test.ts
```diff
@@ -162,4 +165,35 @@ describe("CoreBackService", () => {
       },
     );
   });
+
+  it("appVcReceived should retrieve user identity details and save journey to session", async () => {
```
MikeCollingwood, line 169: This test name implies the journey is being saved to the session but I don't see a check for that. Also, "retrieve user identity details" might be confusing as the journey/next result is more of a "retrieve a successful VC result", I think

PR #1860, opened 2025-02-27
PYIC-7287:  implement no JS jouney for the mobile app

shivanshuit914:
## Proposed changes

- Added middleware check to call API before process

### What changed

- Middleware implementation of radio button check

### Issue tracking

- [PYIC-7287](https://govukverify.atlassian.net/browse/PYIC-7287)

[PYIC-7287]: https://govukverify.atlassian.net/browse/PYIC-7287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/ipv/middleware.ts
+++ b/src/app/ipv/middleware.ts
@@ -48,7 +48,7 @@ import { HANDLED_ERROR } from "../../lib/logger";
 import { sanitiseResponseData } from "../shared/axiosHelper";
 import {
   AppVcReceiptStatus,
-  getAppVcReceiptStatus,
+  getAppVcReceiptStatusAndStoreJourneyResponse,
 } from "../vc-receipt-status/middleware";
 
 const directoryPath = path.resolve("views/ipv/page");
@@ -428,7 +428,7 @@ export const checkFormRadioButtonSelected: RequestHandler = async (
 };
 
 export const checkVcReceiptStatus: RequestHandler = async (req, res, next) => {
-  const status = await getAppVcReceiptStatus(req);
+  const status = await getAppVcReceiptStatusAndStoreJourneyResponse(req);
   if (status === AppVcReceiptStatus.PROCESSING) {
     await handleJourneyPageRequest(req, res, next, true);
   } else if (status === AppVcReceiptStatus.ERROR) {
--- a/src/app/ipv/middleware.ts
+++ b/src/app/ipv/middleware.ts
@@ -46,7 +46,10 @@ import NotFoundError from "../../errors/not-found-error";
 import UnauthorizedError from "../../errors/unauthorized-error";
 import { HANDLED_ERROR } from "../../lib/logger";
 import { sanitiseResponseData } from "../shared/axiosHelper";
-import { getAppVcReceipt } from "../vc-receipt-status/middleware";
+import {
+  AppVcReceiptStatus,
+  getAppVcReceipt,
+} from "../vc-receipt-status/middleware";
 
 const directoryPath = path.resolve("views/ipv/page");
 
@@ -428,8 +431,10 @@ export const checkFormRadioButtonSelected: RequestHandler = async (
   ) {
     // Special case scnario for check-mobile-app-result page to handle non JS browsers
     const status = await getAppVcReceipt(req);
-    if (status === "PROCESSING") {
+    if (status === AppVcReceiptStatus.PROCESSING) {
       await handleJourneyPageRequest(req, res, next, true);
+    } else if (status === AppVcReceiptStatus.ERROR) {
+      throw new Error("Failed to get VC response status");
     } else {
       return next();
     }
--- a/src/app/ipv/middleware.ts
+++ b/src/app/ipv/middleware.ts
@@ -420,24 +420,19 @@ export const checkFormRadioButtonSelected: RequestHandler = async (
   res,
   next,
 ) => {
-  if (
-    req.body.journey === undefined &&
-    req.params.pageId !== PAGES.CHECK_MOBILE_APP_RESULT
-  ) {
+  if (req.body.journey === undefined) {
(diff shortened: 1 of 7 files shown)


src/app/ipv/middleware.ts
```diff
@@ -416,11 +420,19 @@ export const checkFormRadioButtonSelected: RequestHandler = async (
...
+
+export const checkVcReceiptStatus: RequestHandler = async (req, res, next) => {
+  const status = await getAppVcReceipt(req);
+  if (status === AppVcReceiptStatus.PROCESSING) {
     await handleJourneyPageRequest(req, res, next, true);
+  } else if (status === AppVcReceiptStatus.ERROR) {
+    throw new Error("Failed to get VC response status");
   } else {
     return next();
   }
```
MikeCollingwood: I think separating out methods from `getAppVcReceipt` for its two uses makes sense. 


Just a suggestion but with some shuffling around, if `getAppVcReceipt` handler only sets the status code and `req.session.journey`, these two (and `checkFormRadioButtonSelected` after `processVcReceiptForPost`) could handle most of this logic?

```typescript
export const processVcReceiptForPost: RequestHandler = async (req, res, next) => {
  req.body.journey = req.session.journey;
  return next();
}

export const processVcReceiptForSpinner: RequestHandler = async (req, res, next) => {
  if (config.ENABLE_PREVIEW && process.env.NODE_ENV === "local") {
    res.status(200).json({ status: AppVcReceiptStatus.COMPLETED });
    return;
  }

  switch(res.statusCode) { 
    case 200: { 
      res.json({ status: AppVcReceiptStatus.COMPLETED });
      break; 
    } 
    case 404: { 
      res.status(200).json({ status: AppVcReceiptStatus.PROCESSING });
      break; 
    } 
    default: { 
      res.json({ status: AppVcReceiptStatus.ERROR });
      break; 
    } 
  } 
  return next();
}
```

src/app/vc-receipt-status/middleware.ts
```diff
@@ -6,40 +6,49 @@ import { isAxiosError } from "axios";
...
-) => {
+): Promise<AppVcReceiptStatus> => {
   try {
-    // For browser tests
-    if (config.ENABLE_PREVIEW && process.env.NODE_ENV === "local") {
-      res.status(200).json({ status: AppVcReceiptStatus.COMPLETED });
-      return;
-    }
-
     const appVcResponse = await appVcReceived(req);
```
MikeCollingwood, line 27: I think we should put in a check for if req.session.journey already exists, because in the JS Post scenario, it will have been set.

src/app/vc-receipt-status/middleware.ts
```diff
@@ -6,40 +6,49 @@ import { isAxiosError } from "axios";
...
+};
+
+export const getAppVcReceiptStatus: RequestHandler = async (
+  req: Request,
+  res: Response,
+) => {
+  if (config.ENABLE_PREVIEW && process.env.NODE_ENV === "local") {
+    res.status(200).json({ status: AppVcReceiptStatus.COMPLETED });
+    return;
   }
```
blakeyp: do we need this? I think it's reasonable for this to just not work for the all-templates preview page, given it's a special case

shivanshuit914: Hm yes I added this for the browser tests.

src/app/vc-receipt-status/middleware.ts
```diff
@@ -6,40 +6,49 @@ import { isAxiosError } from "axios";
...
+
+  const status = await getAppVcReceipt(req);
```
blakeyp: Can we tidy up the naming a little, maybe `getAppVcReceipt` -> `getAppVcReceiptStatus` given it's getting the status, and maybe the handler `pollVcReceiptStatus` ?

blakeyp: `getAppVcReceiptStatus` could even be `getAppVcReceiptStatusAndStoreJourneyResponse` so as not to hide its side effect

src/app/ipv/middleware.ts
```diff
@@ -416,11 +420,19 @@ export const checkFormRadioButtonSelected: RequestHandler = async (
...
+  }
+};
+
+export const checkVcReceiptStatus: RequestHandler = async (req, res, next) => {
+  const status = await getAppVcReceipt(req);
+  if (status === AppVcReceiptStatus.PROCESSING) {
     await handleJourneyPageRequest(req, res, next, true);
+  } else if (status === AppVcReceiptStatus.ERROR) {
+    throw new Error("Failed to get VC response status");
   } else {
```
blakeyp: to be safe, can we explicitly check the status is COMPLETED before allowing continue

src/app/vc-receipt-status/middleware.ts
blakeyp: Ah ok, could you add a comment?

src/app/vc-receipt-status/middleware.ts
```diff
@@ -6,40 +6,55 @@ import { isAxiosError } from "axios";
...
-      return;
+      return AppVcReceiptStatus.PROCESSING;
     }
     logger.error(error, "Error getting app vc receipt status");
-    res.status(500).json({ status: AppVcReceiptStatus.ERROR });
+    return AppVcReceiptStatus.ERROR;
   }
 };
+
+export const getAppVcReceiptStatusAndStoreJourneyResponse: RequestHandler =
```
blakeyp: I was thinking the above method would be getAppVcReceiptStatusAndStoreJourneyResponse (since it's actually storing the response) and this could be something like pollVcReceiptStatus ?

blakeyp (approved):

PR #1867, opened 2025-03-03
PYIC-7911: serve static service unavailable page

thebauSoftwire:
### What changed
- added script to build a static service-unavailable page during build process
- serve this page when overload protection is hit

| Page |
|-|
| Screenshot 2025-03-03 at 11 34 29 |
### Why did it change
Before, we were just serving a "Server experiencing heavy load: (event loop)" to the user. After a 503, we'll now display a more user-friendly page.

### Issue tracking

- [PYIC-7911](https://govukverify.atlassian.net/browse/PYIC-7911)

[PYIC-7911]: https://govukverify.atlassian.net/browse/PYIC-7911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thebauSoftwire:
Tested locally and in dev env 🍏

build-scripts/build-service-unavailable-html.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Get positional arguments to script
+htmlFilePath=$1 # destination file path for html file
+cssFilePath=$2 # path to compiled css file
+
+echo "Building service-unavailable html file at $htmlFilePath..."
+ts-node ./build-scripts/build-service-unavailable-html.ts $htmlFilePath
+
+echo "Extracting css from $cssFilePath..."
+styling=$(grep -o '.*' $cssFilePath)
```
thebauSoftwire: oops, will remove this

Dockerfile
```diff
@@ -7,16 +7,20 @@ COPY package-lock.json ./
...
 
-# Build assets
+# Get assets
 COPY /assets ./assets
-RUN npm run build
 
-# Build code
+# Get build scripts
+COPY /build-scripts ./build-scripts
+RUN chmod +x ./build-scripts/build-service-unavailable-html.sh
```
Joe-Edwards-GDS: In theory Docker should copy file permissions from the source, and it looks like you've got +x - is it not working?

Dockerfile
```diff
@@ -7,16 +7,20 @@ COPY package-lock.json ./
...
+
+# Compile TS and build code
 COPY /src ./src
 COPY /locales ./locales
 COPY /views ./views
 COPY tsconfig.json ./
 RUN npm run tsc
+RUN npm run build
```
Joe-Edwards-GDS: Note - one small consequence of this is that we'll now rebuild the JS and CSS whenever any source file changes (previously it would only do so if the package.json/npmrc/assets changed).

Fine to leave like this, but if you notice slow repeat builds during local dev, you could consider splitting the npm commands, so we still build the JS/CSS further up, and build the 503 page later.

(Essentially, docker will cache the result of all commands above this one, but only if they are unchanged)

src/handlers/error-handler.test.ts
```diff
@@ -143,6 +143,35 @@ describe("Error handlers", () => {
...
+      serverErrorHandler(err, req, res, next);
+
+      // Assert
+      expect(res.status).to.have.been.calledOnceWith(
+        HttpStatusCode.ServiceUnavailable,
+      );
+      expect(res.render).to.have.been.calledWith("service-unavailable.html");
+    });
+
+    it("should render a service-unavailable page when status is 503", () => {
```
DanCorderIPV: 404?

build-scripts/build-service-unavailable-html.ts
```diff
@@ -0,0 +1,37 @@
+import fs from "fs";
+import i18next from "i18next";
+import {configureNunjucks} from "../src/config/nunjucks";
+
+const buildServiceUnavailableHtml = (outputFile: string) => {
+  const enJson = JSON.parse(fs.readFileSync("./locales/en/translation.json", 'utf-8'))
+  const cyJson = JSON.parse(fs.readFileSync("./locales/cy/translation.json", 'utf-8'))
```
Joe-Edwards-GDS: (It's fine for a build script, but I'd probably still use fs/promises anyway out of habit!)

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "skipLibCheck": true,
     "outDir": "build",
     "inlineSourceMap": true,
-    "types": ["node", "mocha"]
+    "types": ["node", "mocha"],
```
Joe-Edwards-GDS: N.B. this isn't actually valid JSON - Typescript will allow it (for convenience) but it can upset other parsers/linters/etc.

src/handlers/internal-server-error-handler.ts
```diff
@@ -29,12 +33,17 @@ const serverErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
     return next(err);
   }
 
+  if (res.statusCode === HTTP_STATUS_CODES.SERVICE_UNAVAILABLE) {
```
DanCorderIPV: This is probably fine, but it looks like it would be technically slightly better to look at the status in the error object: https://github.com/davidmarkclements/overload-protection?tab=readme-ov-file#errorpropagationmode-false

build-scripts/build-service-unavailable-html.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Create copy of css file
+# The css needs to stay in a file as extracting it into an env variable results in "Arguments too long" errors
+cp $cssFilePath tempStyle.css
+
+# Add style tags to beginning and end of temp style file
+echo '<style>' | cat - tempStyle.css > temp && mv temp tempStyle.css
```
DanCorderIPV: The template has `style` tags too, does this not give us double tags?

views/errors/service-unavailable.njk
```diff
@@ -0,0 +1,29 @@
...
+
+{% block bodyStart %}{% endblock %}
+
+{% block head %}
+  <style>This will be replaced in the build script.</style>
+{% endblock %}
+
+{% block content %}
+  <div role="region" lang="en">
+    <h1 class="govuk-heading-l">Sorry, there is a problem</h1>
```
DanCorderIPV: It's arguably a bit pointless here as this page is only rendered in the "English" version, but it would be more consistent to have this text in the translation file(s) and bring it in that way.
As a side effect you would have had a test failure for the wrong apostrophe used in the Welsh text :)

DanCorderIPV (changes requested):

views/errors/service-unavailable.njk
```diff
@@ -0,0 +1,29 @@
+{% extends "shared/base.njk" %}
+{% set pageTitleKey = "Sorry, there is a problem" %}
+{% set googleTagManagerPageId = "Sorry, there is a problem" %}
+
+{% block bodyStart %}{% endblock %}
+
+{% block head %}
+  <style>This will be replaced in the build script.</style>
```
Joe-Edwards-GDS: I wonder if you could make this work with either an `{% include 'application.css' %}` (you might need to add it to the nunjucks config, so it knows where to find it) or `{{ raw applicationCss }}` (reading the file in TS and setting as a 'local')? Feels a little bit more robust than the `sed` command...

src/handlers/error-handler.test.ts
thebauSoftwire: 🤦‍♀️ and I got the page wrong in the description

src/handlers/internal-server-error-handler.ts
```diff
@@ -29,12 +33,17 @@ const serverErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
...
+    res.status(HTTP_STATUS_CODES.SERVICE_UNAVAILABLE);
+    return res.render(getHtmlPath(ERROR_PAGES.SERVICE_UNAVAILABLE));
```
Joe-Edwards-GDS, line 41: Should this be `res.render`? We don't need to invoke the view engine if you use `res.sendFile`... although I'm not 100% sure what the difference between the two will be(!)

Joe-Edwards-GDS (commented):
Is there any reasonable way of adding a snapshot test for this?

src/handlers/internal-server-error-handler.ts
thebauSoftwire: 😮 thanks!

build-scripts/build-service-unavailable-html.sh
thebauSoftwire: the `'d'` flag in the sed command on line 20:
```
sed -e '/This will be replaced in the build script./ {' -e 'r tempStyle.css' -e 'd' -e '}' $htmlFilePath > tmpHtml.html
```
deletes the line the match was found on including the `<script>` tags. Preserving the text around the matched text was tricky and I couldn't use sed's replace command since it couldn't handle the long css string so settled for this :/

views/errors/service-unavailable.njk
thebauSoftwire: good point, thanks!

src/handlers/internal-server-error-handler.ts
thebauSoftwire: Used `res.render` instead of `res.sendFile` based on [this comment](https://github.com/govuk-one-login/ipv-core-front/pull/1761#discussion_r1893795967) from Chris' spike PR

Dockerfile
thebauSoftwire: ah you're right, put it in originally as I had to update file permissions when I was testing manually

views/errors/service-unavailable.njk
thebauSoftwire: :big: 🧠 
this would've saved me so much pain making sed/perl/awk work 🤦‍♀️

thebauSoftwire: Thanks! Much nicer :)

build-scripts/build-service-unavailable-html.sh
thebauSoftwire: ^removed with Joe's [comment](https://github.com/govuk-one-login/ipv-core-front/pull/1867#discussion_r1977871565) 🎉

views/errors/service-unavailable.njk
Joe-Edwards-GDS: Ah, that's neat!

Knowing how sed etc. work is definitely a handy skill to have, but I always have to look them up each time anyway 😅

src/app/development/middleware.ts
```diff
@@ -113,15 +117,14 @@ export const templatesDisplayGet: RequestHandler = async (req, res) => {
   } else if (templateId === PAGES.PYI_TRIAGE_MOBILE_DOWNLOAD_APP) {
     validatePhoneType(phoneType);
     renderOptions.appDownloadUrl = getAppStoreRedirectUrl(phoneType);
-  } else if (templateId === PAGES.CHECK_MOBILE_APP_RESULT) {
```
DanCorderIPV: Is this a merge conflict issue?

src/handlers/internal-server-error-handler.ts
```diff
@@ -29,12 +33,18 @@ const serverErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
     return next(err);
   }
 
+  // Overload protection will propagate the 503 through the err object
+  if (err.statusCode === HTTP_STATUS_CODES.SERVICE_UNAVAILABLE) {
+    res.status(HTTP_STATUS_CODES.SERVICE_UNAVAILABLE);
+    return res.render(getHtmlPath(ERROR_PAGES.SERVICE_UNAVAILABLE));
```
DanCorderIPV, line 41: Probably worth a comment on why we're not using `res.sendFile` in case someone else has the same thought in the future

DanCorderIPV (changes requested):

DanCorderIPV (dismissed):

src/config/nunjucks.ts
```diff
@@ -35,6 +35,17 @@ export const configureNunjucks = (
     },
   );
 
+  nunjucksEnv.addFilter(
+    "translateWithoutEscape",
```
DanCorderIPV: Why do we need this for the static page, but not the other places we have contact us links?

thebauSoftwire: Not entirely sure, I think it's something to do with how I've initialised `i18n` for the build process vs in the app and the difference in underlying config... I couldn't get the exact same set-up working between both so have just purely loaded the translations into the i18n in the build script

DanCorderIPV (approved):

PR #1910, opened 2025-03-31
PYIC-8194: Fix core-front deployment

MikeCollingwood:
### What changed

- Fix frontend deployment
  - Remove bucket specification which was causing issues at least during first deployment & it's not necessary when AWS will generate a globally unique value by default.
  - Remove redundant steps from secure-post-merge

- Related: Updated CodeDeploy SecurePipeline permissions to be able to create the VpcLink/ NLB/ ALB resources

### Issue tracking

- [PYIC-8194](https://govukverify.atlassian.net/browse/PYIC-8194)

[PYIC-8194]: https://govukverify.atlassian.net/browse/PYIC-8194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -314,10 +314,6 @@ Resources:
           ignore_checks:
             - W3045 # Using a legacy access control setup for now
     Properties:
-      BucketName: !If
-        - IsDevelopment # existing bucket names exist in the current dev env
-        - !Sub ipv-core-access-logs-${Environment}
-        - !Sub ipv-core-${Environment}-access-logs
```
MikeCollingwood, line 320: The bucket name was somehow being used elsewhere, and this is supposed to be globally unique, so removing this specification ensures the bucket name is unique as is determined by AWS

MikeCollingwood:
One of the difficulties with this ticket was in trying to deploy core-front SecurePipelines doesn't include the required permissions for creating a VpcLink by default. I have raised this with devplatform as caused a significant amount of confusion and could affect other first-time deployments of Load Balancers

.github/workflows/secure-post-merge.yml
```diff
@@ -37,17 +31,10 @@ jobs:
           cp .npmrc.template .npmrc && \
           sed -i s/TOKEN_WITH_READ_PACKAGE_PERMISSION/${{ secrets.GITHUB_TOKEN }}/ .npmrc
 
-      - name: npm build assets, zip and sign
-        uses: govuk-one-login/github-actions/govuk/upload-assets@main
-        with:
-          signing-key-arn: ${{ secrets.UPLOAD_ASSETS_ZIP_SIGNING_KEY }}
-          stack-name: 'core-front'
-          destination-bucket-name: ${{ secrets.UPLOAD_ASSETS_ARTIFACT_SOURCE_BUCKET_NAME }}
```
MikeCollingwood, line 45: It seems as though these two secrets are no longer needed. This step, I believe, was made redundant by the `devplatform-upload-action-ecr` step

browser-tests/snapshot-tests/snapshot.spec.ts-snapshots/page-face-to-face-handoff-cy-chromium-linux.jpeg
MikeCollingwood, line 1: Changed because of the month - think this probably wants to be hardcoded for dev templates

deploy/template.yaml
AmritSidhu: Would this cause a new bucket to be created every deploy?

MikeCollingwood: Not per deploy, since it wouldn't need to be re-created once made. It's just for the first-time deployment

AmritSidhu (approved):

PR #1922, opened 2025-04-04
PYIC-8210: Add device intelligence

DanCorderIPV:
### What changed

Add device intelligence

### Why did it change

New feature

### Issue tracking

- [PYIC-8210](https://govukverify.atlassian.net/browse/PYIC-8210)

[PYIC-8210]: https://govukverify.atlassian.net/browse/PYIC-8210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


build-scripts/build-service-unavailable-html.ts
```diff
@@ -31,6 +31,7 @@ const buildServiceUnavailableHtml = async (outputFile: string) => {
       language: "en"
     },
     showLanguageToggle: false,
+    useDeviceIntelligence: false,
```
DanCorderIPV, line 34: I'm assuming we'd prefer not to have extra JS file load in this case.

DanCorderIPV:
In local running I get the new cookie set

build-scripts/build-service-unavailable-html.ts
DanCorderIPV: In fact, I'm not sure where this file lives, so the JS path might not work anyway?

MikeCollingwood: Not sure I follow, but I would've thought we wouldn't need to set this because of the truthy check in the Nunjucks base

MikeCollingwood (approved):
LGTM! Follows the implementation guide

build-scripts/build-service-unavailable-html.ts
DanCorderIPV: True, false is the default anyway. The other comments are about this being the static error page that is generated at build time and used if the system gets overloaded. That has reminded me that I think it does sit with the other routes (rather than being hosted separately on AWS as it used to be).

PR #1926, opened 2025-04-08
PYIC-8204: Improve page banner to allow multiple banners to be configured for the same page with different contexts

DanCorderIPV:
### What changed

The same page can have multiple banners configured for different contexts

### Why did it change

Previously multiple configs for the same page would overwrite each other.

### Issue tracking

- [PYIC-8204](https://govukverify.atlassian.net/browse/PYIC-8204)

[PYIC-8204]: https://govukverify.atlassian.net/browse/PYIC-8204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/notification-banner-handler.ts
```diff
@@ -31,15 +31,16 @@ const notificationBannerHandler: RequestHandler = async (req, res, next) => {
...
+          req.session.context === data.context)
       ) {
         res.locals.displayBanner = true;
-        res.locals.bannerContext = data.context;
         res.locals.bannerType = data.bannerType;
         res.locals.bannerMessage =
-          req.i18n.language === "en"
-            ? data.bannerMessage
-            : data.bannerMessageCy;
+          req.i18n.language === "cy"
```
DanCorderIPV, line 41: Defaulting to English makes the tests slightly nicer

src/test-utils/pages-and-contexts.ts
```diff
@@ -1,6 +1,6 @@
 export const pagesAndContexts: Record<string, (string | undefined)[]> = {
   "check-mobile-app-result": [],
-  "confirm-your-details": ["test", undefined],
```
DanCorderIPV, line 3: I don't think we should be inventing contexts to test this with - it adds unrealistic snapshots

src/handlers/notification-banner-handler.ts
```diff
@@ -31,15 +31,16 @@ const notificationBannerHandler: RequestHandler = async (req, res, next) => {
       if (
         req.path === data.pageId &&
         currentTime >= data.startTime &&
-        currentTime <= data.endTime
+        currentTime <= data.endTime &&
```
DanCorderIPV, line 34: I considered moving this to where we set up the other `renderOptions` values, but that is in more than one place. Here `req.session.context` is set up before this middleware is called so this works and only has to be in a single place.

browser-tests/compose.yaml
```diff
@@ -26,7 +26,8 @@ services:
       LANGUAGE_TOGGLE: true
       USE_DEVICE_INTELLIGENCE: true
       SERVICE_URL: http://localhost:4601
-      NOTIFICATION_BANNER: '[{"pageId":"/dev/template/confirm-your-details/en","bannerMessage":"<h3>Test banner</h3> <p>This is a test banner</p>","bannerMessageCy":"<h3>Welsh Test banner</h3> <p>This is a Welsh test banner</p>","startTime":"2024-10-25T06:14:40.162Z","endTime":"2030-10-25T06:14:40.162Z", "context": "test"}]'
+# Note that due to how snapshot tests work the pageIds here include the language so the banners only appear in one language version of the snapshot. We also can't test banner contexts in snapshot tests as the all-templates page shortcuts a lot of stuff.
+      NOTIFICATION_BANNER: '[{"pageId":"/dev/template/live-in-uk/en","bannerMessage":"<h3>Test banner</h3> <p>This is a test banner</p>","bannerMessageCy":"Not seen","startTime":"2024-10-25T06:14:40.162Z","endTime":"2030-10-25T06:14:40.162Z"}]'
```
thebauSoftwire, line 30: does this mean the ss test for this will fail in 2030? Is there a way to make this dynamic?

thebauSoftwire (approved):
Nice! One optional comment - from a quick search it doesn't look like docker has a nice way of doing this so might be a problem not worth tackling as part of this ticket

browser-tests/compose.yaml
DanCorderIPV: It will fail in 2030 - we could make it 2099 or later. As a general rule I prefer to not have dynamic times etc in tests, as it can lead to inconsistent failures.

PR #1937, opened 2025-04-11
PYIC-8171: Display country name for international users

AmritSidhu:
### What changed

Display country name for international users
Address CRI just keep a list of country codes to names which we have copied over
This data is taken from ipv-cri-address-front/blob/main/src/app/address/data/countries.json & https://github.com/govuk-one-login/ipv-cri-address-front/blob/main/src/locales/en/fields.yml

### Issue tracking

- [PYIC-8171](https://govukverify.atlassian.net/browse/PYIC-8171)

[PYIC-8171]: https://govukverify.atlassian.net/browse/PYIC-8171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/shared/addressHelper.ts
```diff
@@ -71,5 +68,12 @@ export const generateHTMLofAddress = (address: PostalAddressClass): string => {
   if (address.addressRegion) html += `${address.addressRegion}<br>`;
   html += `${address.postalCode}`;
 
+  const countryName = address.addressCountry
+    ? countryMap[address.addressCountry]
+    : undefined;
+  if (countryName && address.addressCountry !== "GB") {
```
blakeyp: could we check if country code is non GB _before_ doing the lookup? slight optimisation so we only do the lookup for intl' addresses

blakeyp (approved):

PR #1943, opened 2025-04-15
PYIC-8049: Add missing pages + contexts to dev/all-templates

AmritSidhu:
### What changed

From talking to UCD there was a few pages they could not find on the all-template routes

Not all pages are under the ipv directory where the list of pages gets pulled from

Pages added to all-templates:

- pyi-technical (unrecoverable)
- pages under errors/ directory

### Issue tracking

- [PYIC-8049](https://govukverify.atlassian.net/browse/PYIC-8049)

[PYIC-8049]: https://govukverify.atlassian.net/browse/PYIC-8049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/development/middleware.ts
```diff
@@ -102,14 +104,25 @@ export const templatesDisplayGet: RequestHandler = async (req, res) => {
...
+  const errorTemplates = new Set<string>(
+    Object.values(ERROR_PAGES).filter(
+      (p) => p !== ERROR_PAGES.SERVICE_UNAVAILABLE,
+    ),
+  );
+
+  if (errorTemplates.has(templateId)) {
+    return res.render(getErrorPageTemplatePath(templateId), renderOptions);
+  }
+
```
thebauSoftwire: I wonder if we can combine the `serviceUnavailableGet` middleware with this one so we don't have to filter out the service_unavailable page?
```
if (errorTemplates.has(templateId)) {
    if (templateId === ERROR_PAGES.SERVICE_UNAVAILABLE) {
         res.render(getHtmlPath(ERROR_PAGES.SERVICE_UNAVAILABLE));
    } 
    return res.render(getErrorPageTemplatePath(templateId), renderOptions);
 }
```
Then we can also remove the explicit routing for the service_unavailable page in router.ts (I think I wrote this in before we were displaying the error pages and didn't think far enough ahead 🙏 )

thebauSoftwire (commented):
Nice! Just one comment on potentially combing some middleware

src/app/development/middleware.ts
AmritSidhu: I did initially try something similar to this but where the SERVICE_UNAVAILABLE page loads in its own css it behaves slightly different. I will give it a go in the way you suggested above

thebauSoftwire (approved):

PR #1996, opened 2025-05-12
PYIC-8283: Fix device detection to route iPads to desktop flow instead of mobile app download

AmritSidhu:
### What changed

Fix device detection to route iPads to desktop flow Previously, iPads were incorrectly detected as iPhones due to relying solely on the OS name (iOS) without checking device type. 
This caused users to be routed to the mobile app download page instead of the select device page. 
Updated the sniffPhoneType logic to prioritise device type detection, treating only 'mobile' devices with iOS or Android as phones. 

### Why did it change

All other device types, including 'tablet' (iPads), now correctly fall back to the desktop flow. 
This restores expected behaviour after the recent OS version handling changes.

### Issue tracking

- [PYIC-8283](https://govukverify.atlassian.net/browse/PYIC-8283)

## Checklists

[PYIC-8283]: https://govukverify.atlassian.net/browse/PYIC-8283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/shared/deviceSniffingHelper.ts
+++ b/src/app/shared/deviceSniffingHelper.ts
@@ -41,16 +41,22 @@ export const sniffPhoneType = (
   fallback?: OsType,
 ): OsType | null => {
   const parser = new UAParser(req.headers["user-agent"]);
+  const deviceType = parser.getDevice().type;
+  const osName = parser.getOS().name;
+  const version = parser.getOS().version;
 
-  const version = parser.getOS()["version"];
-  switch (parser.getOS()["name"]) {
-    case OS_TYPES.IOS:
+  // Only treat "mobile" + OS as phone
+  if (deviceType === "mobile") {
+    if (osName === OS_TYPES.IOS) {
       return { name: PHONE_TYPES.IPHONE, version: parseVersion(version) };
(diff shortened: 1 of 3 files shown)


src/app/shared/deviceSniffingHelper.ts
```diff
@@ -41,16 +41,22 @@ export const sniffPhoneType = (
...
+  const deviceType = parser.getDevice().type;
+  const osName = parser.getOS().name;
+  const version = parser.getOS().version;
 
-  const version = parser.getOS()["version"];
-  switch (parser.getOS()["name"]) {
-    case OS_TYPES.IOS:
+  // Only treat "mobile" + OS as phone
+  if (deviceType === "mobile") {
+    if (osName === OS_TYPES.IOS) {
```
thebauSoftwire, line 50: Do we need to add/update any unit or browser tests for this? E.g. a case for an ipad

thebauSoftwire (commented):
Nice! Just one comment around tests

thebauSoftwire (approved):

PR #2011, opened 2025-05-19
BAU: Add DT trace id to logs

Joe-Edwards-GDS:
### What changed

- Commonised additional log properties
- Added Dynatrace trace id via the DT SDK

### Why did it change

- So we can connect logs with DT traces

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -175,14 +175,6 @@ app.use((req, res, next) => {
 
 app.use(notificationBannerHandler);
 
-app.use((req, res, next) => {
-  req.log = logger.child({
-    requestId: req.id,
-    ipvSessionId: req.session?.ipvSessionId,
-    sessionId: req.session?.id,
-  });
-  next();
-});
 app.use(loggerMiddleware);
 
 // Must be added to the app after the session and logging, and before the routers.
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -200,15 +192,6 @@ app.set("etag", false);
...
 const router = express.Router();
 
-router.use((req, res, next) => {
-  req.log = logger.child({
-    requestId: req.id,
-    ipvSessionId: req.session?.ipvSessionId,
-    sessionId: req.session?.id,
-  });
-  next();
-});
```
Joe-Edwards-GDS, line 210: Not sure where this came from:
- We seem to do this _twice_ - once on everything (above) and once on the router only, but it's identical
- The pino-http middleware already sets `req.log` for us

If there's a good reason then happy to keep it...

hugh-emerson: Git says you're the one who added both those blocks... so I think it's up to you to say whether there's a reason.

Is there a reason this PR is still a draft?

Joe-Edwards-GDS: I suspect they appeared via a merge conflict but I'm not sure - happy to say we only need one, and perhaps that should be extracted to a separate PR.

This specific PR is still in draft because we've been directed not to tie ourselves into Dynatrace-specific libraries. Ideally we'd use OpenTelemetry instead (as an open standard) but I couldn't get that to work.

Fine to close this if you're not missing the functionality!

Joe-Edwards-GDS:
Should be updated to use OpenTelemetry if desired

PR #2023, opened 2025-05-23
PYIC-8330: handle unknown route errors for core-front

thebauSoftwire:
### What changed
- updated error handling so that
  - POST/GET to unknown routes result in 404
  - POST requests with missing body results in 400
  - POST requests with missing or invalid CSRF token results in 403
  - PUT/DELETE requests to known or unknown routes return 404 as they are not defined routes
- moved `csrfSynchronisedProtection` handler to individual `dev` and `ipv` middleware instead of using directly on app router
- updated PR template checklist to include making sure `csrfSynchronisedProtection` middleware is added to added/updated routes where appropriate

### Why did it change
Our frontend 5xx alarm gets triggered occasionally because bots spamming the service with unknown routes result in 500s from core-front. This is because for unknown routes, we were failing with a `TypeError: Cannot read properties of undefined (reading '_csrf')` as a result of csrf protection middleware.

### Issue tracking

- [PYIC-8330](https://govukverify.atlassian.net/browse/PYIC-8330)

## Checklists

- [x] Browser/unit tests have been written/updated

[PYIC-8330]: https://govukverify.atlassian.net/browse/PYIC-8330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app.ts
```diff
@@ -186,9 +185,6 @@ app.use((req, res, next) => {
 });
 app.use(loggerMiddleware);
 
-// Must be added to the app after the session and logging, and before the routers.
-app.use(csrfSynchronisedProtection);
-
```
thebauSoftwire, line 191: Adding the csrf protection middleware at this level meant that any unknown routes passed through this middleware first instead of hitting `pageNotFoundHandler` so I've moved this middleware to the individual routers. A downside is having to define it for each of the routes manually but from speaking with Patrick, have tried to negate risks by adding an extra check in the PR template

AmritSidhu (dismissed):

src/lib/csrf.ts
```diff
@@ -1,10 +1,21 @@
...
+    if (!req.body) {
+      throw new BadRequestError("Missing request body");
+    }
+
+    const csrfToken = req.body["_csrf"];
+    if (!csrfToken) {
+      throw new ForbiddenError("Missing csrf token");
+    }
+
+    return csrfToken;
```
blakeyp: is this not handled by the csrf package itself if we just return `csrfToken` here?

thebauSoftwire: ah you're right, thanks!

AmritSidhu (dismissed):

src/handlers/journey-event-error-handler.ts
```diff
@@ -28,10 +28,7 @@ const journeyEventErrorHandler: ErrorRequestHandler = (err, req, res, next) => {
     // Set this to avoid pino-http generating a new error in the request log
     res.err = HANDLED_ERROR;
 
-    return res.render(getIpvPageTemplatePath(pageId), {
-      pageId: pageId,
-      csrfToken: req.csrfToken?.(true),
-    });
+    return res.redirect(getIpvPagePath(pageId));
```
thebauSoftwire: The `getTokenFromRequest` won't detect a csrf even if we generate one here because the csrf middleware wasn't used before this res.render. Using a redirect here instead forces express to use the csrf protected `get` route rather than this res.render for rendering the page

thebauSoftwire:
Ran e2e tests against current changes in dev environment and all tests passing 🍏

src/handlers/journey-event-error-handler.ts
blakeyp: Ah I don't think we can make this change - we're deliberately using render so we reflect the error status code in the response rather than it being a redirect and serving from a new route

Is there another way we can handle this 🤔

thebauSoftwire: ahh that makes sense, thanks. I've replaced it now with `generateToken` from csrf-sync which is equivalent, I didn't realise that the above only applies if you use the csrf on the req

blakeyp (dismissed):

src/app/ipv/tests/validatePageId.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import { expect } from "chai";
+import NotFoundError from "../../../errors/not-found-error";
+
+const createRequest = specifyCreateRequest();
+const createResponse = specifyCreateResponse();
+
+describe("validatePageId", () => {
+  it("should throw if invalid pageId is provided", () => {
+    // Arrange
+    const req = createRequest({ params: { pageId: "unkownPageId" } });
```
aleemriaz: minor typo in the pageId: "unknownPageId"

src/app/ipv/tests/validatePageId.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const next = sinon.fake();
+
+    // Act/Assert
+    expect(() => validatePageId(req, res, next)).to.throw(
+      NotFoundError,
+      "Invalid page id",
+    );
+  });
+
+  it("should throw if invalid pageId is provided", () => {
```
aleemriaz: Please update this test name to reflect that this is checking with a valid pageId.

thebauSoftwire: ah thanks 🤦‍♀️

aleemriaz (approved):
Approved

PR #2037, opened 2025-05-30
PYIC-8329: Fix date comparisons and validate date format in config

DanCorderIPV:
### What changed

Fixed date comparisons for displaying banners

### Why did it change

It was broken when using timezones

### Issue tracking

- [PYIC-8329](https://govukverify.atlassian.net/browse/PYIC-8320)

## Checklists

- [ x ] Browser/unit tests have been written/updated

[PYIC-8329]: https://govukverify.atlassian.net/browse/PYIC-8329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/notification-banner-handler.ts
```diff
@@ -41,11 +52,16 @@ const notificationBannerHandler: RequestHandler = async (req, res, next) => {
       bannerConfigs1Parsed.concat(bannerConfigs2Parsed);
 
     bannerConfigsParsed.forEach((data: BannerConfig) => {
-      const currentTime = new Date().toISOString();
+      validateDate(data.startTime);
+      validateDate(data.endTime);
+
+      const bannerStartTime = new Date(data.startTime);
+      const bannerEndTime = new Date(data.endTime);
+      const currentTime = new Date();
```
thebauSoftwire, line 60: could we instead convert the validated date string into UTC with `new Date(data.startTime).toUTCString()` and ensure that currentTime is also in UTC so that we don't have to add a timezone offset to the config?

thebauSoftwire, line 60: Instead of explicitly adding the timezone offset in the config, could we enforce the timezone here, convert the validated date string into UTC with new Date(data.startTime).toUTCString() and ensure that currentTime is also in UTC?

src/handlers/notification-banner-handler.test.ts
```diff
@@ -383,10 +443,11 @@ describe("Notification banner handler", () => {
     parameterServiceStub.getParameter = sinon.fake.resolves("not JSON");
 
     // Act
-    await notificationBannerHandler(req, res, next);
+    await underTest(req, res, next);
 
     // Assert
     expect(res.locals.displayBanner).to.be.false;
     expect(next).to.have.been.calledOnce;
+    expect(loggerStub.logger.error).to.have.been.calledOnce;
```
thebauSoftwire, line 451: nice!

src/handlers/notification-banner-handler.ts
thebauSoftwire: though I guess it means this would be less flexible if we enforce the timezone here

DanCorderIPV: Maybe I'm misunderstanding, but my question would be which timezone would we enforce? The UK switches between two timezones during the year so picking one of them would seem odd to me.

thebauSoftwire (approved):

PR #2103, opened 2025-07-08
Pyic 8471 Add timestamp stored in sessionStorage and initTime

undermad:
### What changed

- Spinner component 
    - Removed longWait and abortTime setTimeouts.
    - Added logic that update state according to time elapsed
    - Introduced initTime stored in sessionStorage.
    - Add abortController to cancel fetch on page unload.
    - Added updateDomTimer interval.

### Why did it change

Spinner still uses setTimeout for polling and setInterval for DOM update to avoid usage of loops. Each of those call first check time elapsed from the initialisation time. This make it resistant against power saving and sleep modes.   

It also prevent spinner from updating to abort or long wait state even if spinner already end in completed state.

"During the v2 app switch-ons in prod it’s been noticed that in some cases the MAM/DAD spinner polling for DCMAW VC receipt appears not to timeout within the configured window and/or stops but then resumes some time later. Because the EVCS access token has an expiry of 1 hour, eventually the polling calls to EVCS fail with 401s.

It’s possible that this is being caused by browser throttling (background tab, device sleep, power saving) because of users leaving the tab open running and abandoning their journeys."

### Issue tracking

- [PYIC-8471](https://govukverify.atlassian.net/browse/PYIC-8471)

## Checklists

[PYIC-8471]: https://govukverify.atlassian.net/browse/PYIC-8471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/javascript/spinner/init.js
+++ b/assets/javascript/spinner/init.js
@@ -28,7 +28,7 @@ class Spinner {
   };
 
   updateAccordingToTimeElapsed = () => {
-    const elapsedMilliseconds = this.initTime - new Date().getTime();
+    const elapsedMilliseconds = new Date().getTime() - this.initTime;
     if (elapsedMilliseconds >= this.config.msBeforeAbort) {
       this.reflectError();
     } else if (elapsedMilliseconds >= this.config.msBeforeInformingOfLongWait) {
--- a/assets/javascript/spinner/init.js
+++ b/assets/javascript/spinner/init.js
@@ -41,16 +41,6 @@ class Spinner {
     if (this.domRequirementsMet) {
       this.spinnerState = "pending";
       this.button.setAttribute("disabled", true);
-
-      let spinnerInitTime = sessionStorage.getItem('spinnerInitTime')
-      if(spinnerInitTime === null) {
-        spinnerInitTime = Date.now();
-        sessionStorage.setItem('spinnerInitTime', spinnerInitTime.toString());
-      } else {
-        spinnerInitTime = parseInt(spinnerInitTime, 10);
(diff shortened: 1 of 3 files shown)


assets/javascript/spinner/init.js
```diff
@@ -134,30 +152,39 @@ class Spinner {
...
+      .then(response => response.json())
+      .then(data => {
+        if (data.status === "COMPLETED") {
+          this.reflectCompletion();
+        } else if (data.status === "ERROR") {
+          this.reflectError();
+        } else if (data.status === "PROCESSING") {
+          setTimeout(async () => {
+            if ((Date.now() - this.initTime) >= this.config.msBeforeAbort) {
+              this.reflectError();
```
DanCorderIPV: This should match spinner library version - either delete this line, or add in a call to `updateDom()` as you have there.

DanCorderIPV (changes requested):

assets/javascript/spinner/init.js
undermad: Main difference in reflectError is that core-front version navigate to error page and spinner version just change DOM content but It can just simply return as DOM update timer is introduced in this PR and will handle correct behaviour.

DanCorderIPV: Good point, I missed that

DanCorderIPV (dismissed):

DanCorderIPV (approved):

PR #2111, opened 2025-07-10
PYIC-8413: Add flag and turn on for dev and build

DanCorderIPV:
NOTE: Because we use low quality JPEGs for the snapshots the cookie banner and footer have grey backgrounds in the snapshots. This is a JPEG artifact and on the site they are actually the correct subtle blue.
We could increase the JPEG quality level, but as these parts of the page are now coming from an external library they shouldn't change so it may not be worth the extra storage space.

### What changed

Turn on rebrand in dev and build

### Why did it change

New rebrand is going live in production soon

### Issue tracking

- [PYIC-8413](https://govukverify.atlassian.net/browse/PYIC-8413)

[PYIC-8413]: https://govukverify.atlassian.net/browse/PYIC-8413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV:
Note that the service unavailable page has not changed in this PR. There is only one static version of that page and it needs to match the production branding.

thebauSoftwire (approved):

PR #2131, opened 2025-07-21
BAU Downgrade ua-parser-js following licence change

blakeyp:
### What changed

Downgrade ua-paser-js to latest 1.x version

### Why did it change

[Since the release of version 2](https://github.com/faisalman/ua-parser-js/blob/master/CHANGELOG.md) this package is now under an AGPL licence which is incompatible with our licence (MIT). We likely need to source an alternative package given no guarantee of future patches under v1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "pino-http": "10.3.0",
         "qrcode": "1.5.4",
         "sanitize-filename": "1.6.3",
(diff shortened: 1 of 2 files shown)


blakeyp:
https://security.snyk.io/package/npm/ua-parser-js/1.0.40

AmritSidhu (approved):

PR #2203, opened 2025-08-29
feat(692): upgrading to GA4

DavidMcMichael-GDS:
### What changed

Updating dependencies and adding the ANALYTICS_DATA_SENSITIVE enviroment variable

GA4 and content ID's have now been set on all pages
-

### Why did it change

To enable GA4 on page load tracking
-

### Issue tracking

- [DFC-692](https://govukverify.atlassian.net/browse/DFC-692)

## Checklists

[DFC-692]: https://govukverify.atlassian.net/browse/DFC-692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -222,7 +227,7 @@ Mappings:
       may2025RebrandEnabled: true
       alwaysShowBanners: false
       ga4Disabled: false
-      uaDisabled: false
```
blakeyp, line 225: Is this removal intentional? I think CloudFormation will complain about it being missing

package.json
```diff
@@ -80,11 +80,11 @@
     "@aws-sdk/client-dynamodb": "3.868.0",
     "@aws-sdk/client-ssm": "3.864.0",
     "@govuk-one-login/data-vocab": "1.9.3",
-    "@govuk-one-login/frontend-analytics": "3.3.0",
+    "@govuk-one-login/frontend-analytics": "^4.0.3",
     "@govuk-one-login/frontend-device-intelligence": "1.2.0",
     "@govuk-one-login/frontend-language-toggle": "2.2.1",
     "@govuk-one-login/frontend-passthrough-headers": "1.3.0",
-    "@govuk-one-login/frontend-ui": "1.5.0",
+    "@govuk-one-login/frontend-ui": "^2.0.0",
```
blakeyp: Is the bump of this package required for the GA4 rollout? only because we have another ticket which will bump this soon (so would rather leave it for now if it's not needed)

deploy/template.yaml
```diff
@@ -222,7 +227,7 @@ Mappings:
...
+      analyticsDataSensitive: false
```
blakeyp: Shouldn't this be set to true initially in int and prod?
> Identity Pod must disable the repo flag in staging but enable the repo flag in production when they merge the changes. This is very important for allowing us to test the implementation in a non-production environment without the risk of releasing changes to production prematurely.

package.json
DavidMcMichael-GDS: Yeah, moving frontend-ui to v2 is required for the GA4 rollout as the basefile is updated within it. However I can change this to be pinned to just version 2 if you'd prefer?

deploy/template.yaml
DavidMcMichael-GDS: Good catch, will fix now

deploy/template.yaml
DavidMcMichael-GDS: I believe if this is set to true it redacts all data from the service and doesn't allow the per page overwrite to be used. [https://www.npmjs.com/package/@govuk-one-login/frontend-analytics]


The base template now sets the default of each page to have 

`{% set isPageDataSensitive = true %}`

and any page where we wish to use GA4 now contains
`{% set isPageDataSensitive = false %}`

package.json
blakeyp: Yes please pin the version, thanks

deploy/template.yaml
blakeyp: I think the intention is it should be set to true in int and prod initially until testing is completed in staging?

blakeyp:
@DavidMcMichael-GDS frontend-ui v3 has a breaking change (which we're handling in a separate PR)

package.json
```diff
@@ -80,11 +80,11 @@
     "@aws-sdk/client-dynamodb": "3.868.0",
     "@aws-sdk/client-ssm": "3.864.0",
     "@govuk-one-login/data-vocab": "1.9.3",
-    "@govuk-one-login/frontend-analytics": "3.3.0",
+    "@govuk-one-login/frontend-analytics": "^4.0.3",
```
blakeyp: Could you pin the version please

blakeyp (approved):

PR #2258, opened 2025-09-24
Pyic 8272 - Playwright E2E POC

aleemriaz:
### What changed

- A working proof of concept is created in Playwright that successfully replicates the functionality of the identified Selenium end-to-end test.

### Why did it change

- POC for spike

### Issue tracking

- [PYIC-8272](https://govukverify.atlassian.net/browse/PYIC-8272)

## Checklists

[PYIC-8272]: https://govukverify.atlassian.net/browse/PYIC-8272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


browser-tests/e2e-tests/pages/doc-checking-page.ts
```diff
@@ -0,0 +1,26 @@
+import { StubPage } from './stub-page';
+
+export class DocCheckingPage extends StubPage {
+  async processAliceParkerValidInitial(): Promise<void> {
+    await this.page.locator('#test_data').click();
+    await this.selectTestData('Alice Parker (Valid) DVLA Licence');
+    await this.setEvidenceScores({
+      strengthHours: '3',
```
DanCorderIPV: Why do these field names end in `hours`, `minutes`, and `seconds`?

browser-tests/e2e-tests/config/test-config.ts
```diff
@@ -0,0 +1,87 @@
...
+  // API Keys
+  static getTicfApiKey(): string {
+    return this.getEnvironmentVariableOrError('TICF_MANAGEMENT_API_KEY');
+  }
+
+  // Feature Flags
+  static getFeatureFlagsUrl(): string {
+    const baseUrl = this.getIdentityBuildUrl();
+    const featureSet = this.getEnvironmentVariableOrDefault(
+      'FEATURE_FLAGS',
```
DanCorderIPV: I don't think this makes sense as an environment variable - we'll want different flags per tests

browser-tests/e2e-tests/config/test-config.ts
```diff
@@ -0,0 +1,87 @@
...
+  URLS: {
+    ORCHESTRATOR_STUB: ConfigurationReader.getOrchestratorUrl(),
+    IDENTITY_BUILD: ConfigurationReader.getIdentityBuildUrl(),
+    TICF_MANAGEMENT_API: ConfigurationReader.getTicfManagementUrl(),
+  },
+  API: {
+    TICF_API_KEY: ConfigurationReader.getTicfApiKey(),
+  },
+  FEATURE_FLAGS: {
+    ENABLE_URL: ConfigurationReader.getFeatureFlagsUrl(),
```
DanCorderIPV: Again, not sure a single feature flags URL will be useful for most tests.

browser-tests/e2e-tests/fixtures/page-fixtures.ts
```diff
@@ -0,0 +1,44 @@
...
+  drivingLicencePage: async ({ page }, use) => {
+    await use(new DrivingLicenceStubPage(page));
+  },
+  addressPage: async ({ page }, use) => {
+    await use(new AddressStubPage(page));
+  },
+  fraudPage: async ({ page }, use) => {
+    await use(new FraudStubPage(page));
+  },
+  apiService: async ({ request }, use) => {
```
DanCorderIPV: This should be `ticfApiService`?

browser-tests/e2e-tests/pages/base-page.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export abstract class BasePage {
+  constructor(protected page: Page) {}
+
+  protected async navigateTo(url: string): Promise<void> {
+    await this.page.goto(url);
+  }
+
+  protected async clickButton(name: string): Promise<void> {
+    await this.page.getByRole('button', { name }).click();
```
DanCorderIPV: Would these be better off targetting an ID rather than a name?

browser-tests/e2e-tests/pages/base-page.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  protected async selectOption(selector: string, value: string): Promise<void> {
+    await this.page.locator(selector).selectOption(value);
+  }
+
+  protected async expectHeading(name: string): Promise<void> {
+    await expect(this.page.getByRole('heading', { name })).toBeVisible();
+  }
+
+  protected async expectText(text: string): Promise<void> {
```
DanCorderIPV: I've never been particularly keen on this method in the tests. Should we really be looking for specific text on a page in end to end tests?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
+import { expect } from '@playwright/test';
+import { BasePage } from './base-page';
+import { CONFIG } from '../config/test-config';
+
+export class IdentityPage extends BasePage {
+  async enableFeatureFlags(): Promise<void> {
+    await this.navigateTo(CONFIG.FEATURE_FLAGS.ENABLE_URL);
+    await this.page.goBack();
+    await this.page.goBack();
```
DanCorderIPV: Why go back twice?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
...
+    await this.selectRadio('Yes');
+    await this.clickButton('Continue');
+  }
+
+  async navigateToConfirmDetails(): Promise<void> {
+    await this.navigateTo(`${CONFIG.URLS.IDENTITY_BUILD}/ipv/page/confirm-your-details`);
+  }
+
+  async selectUpdateDetails(): Promise<void> {
+    await this.selectRadio('No - I need to update my');
```
DanCorderIPV: This seems fragile, can we use an ID or Value instead of text?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
...
+    await this.selectCheckbox('Given names');
+    await this.clickButton('Continue');
+  }
+
+  async selectUpdateNameMethod(): Promise<void> {
+    await this.selectRadio('Update your name using the');
+    await this.clickButton('Continue');
+  }
+
+  async navigateToDcmawSuccess(): Promise<void> {
```
DanCorderIPV: Why do we need to navigate to this page? The end user will get there by clicking a button or link won't they?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
+import { expect } from '@playwright/test';
+import { BasePage } from './base-page';
+import { CONFIG } from '../config/test-config';
+
+export class IdentityPage extends BasePage {
```
DanCorderIPV: This class seems to contain methods relating to more than one page - should they be split into separate page classes?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
...
+  async navigateToIPVSuccess(): Promise<void> {
+    await this.navigateTo(`${CONFIG.URLS.IDENTITY_BUILD}/ipv/page/page-ipv-success`);
+  }
+
+  async continueFromDcmaw(): Promise<void> {
+    await this.clickButton('Continue');
+  }
+
+  async expectIPVSuccess(): Promise<void> {
+    await this.expectHeading('Continue to the service you need to use');
```
DanCorderIPV: Would it be better to confirm that we're on the success page URL rather than look for text that might change in the future?

browser-tests/e2e-tests/pages/identity-page.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  async continueToService(): Promise<void> {
+    await this.clickButton('Continue to the service');
+  }
+
+  async expectReuseScreen(): Promise<void> {
+    await expect(this.page.locator('#header')).toContainText('You have already proved your identity');
+    await this.expectText('ALISON JANE PARKER');
+    await this.expectText('80TYEOMAN WAYTROWBRIDGEBA14');
```
DanCorderIPV: Why are there missing spaces here?

browser-tests/e2e-tests/pages/orchestrator-stub-page.ts
```diff
@@ -0,0 +1,43 @@
...
+  async navigate(): Promise<void> {
+    await this.navigateTo(CONFIG.URLS.ORCHESTRATOR_STUB);
+  }
+
+  async startFullJourney(): Promise<void> {
+    await this.clickButton('Full journey route');
+  }
+
+  async getUserId(): Promise<string> {
+    return await this.page.getByRole('textbox', { name: 'Enter userId manually' }).inputValue();
```
DanCorderIPV: Again can we use an ID or similar rather than text?
(This applies to other place too, I won't comment on all of them)

DanCorderIPV (changes requested):

browser-tests/e2e-tests/config/test-config.ts
aleemriaz: Yeah you’re right. This is just a POC for this specific test. Once we start adding more tests, we can parameterise this variable so each test can pass in the feature flags it needs.

browser-tests/e2e-tests/fixtures/page-fixtures.ts
aleemriaz: thanks, I'll update.

browser-tests/e2e-tests/pages/doc-checking-page.ts
thebauSoftwire: this feels unintentional, I've removed these unless @aleemriaz you think they should be kept?

browser-tests/e2e-tests/pages/base-page.ts
thebauSoftwire: I think we have e2e tests which test that we've got the correct details on a reuse screen (after the user comes back from a new identity proving journey or an update name journey) and that we return the correct VCs to orch after orch calls `build-user-identity` which I think both use steps which check that the correct text is displayed on the screen.

Saying that though, we do have a [browser test](https://github.com/govuk-one-login/ipv-core-front/blob/4cfe3c6849a8b700607d82fc973f0cfe29ee7a71/browser-tests/functional-tests/functional.spec.ts#L125-L147) which tests whether we can display the correct user details returned by core-back when core-front calls the `build-proven-user-details` lambda and we have [api tests](https://github.com/govuk-one-login/ipv-core-back/blob/d235a25858b55d0365e59e3b6cc1a9cd70ac538e/api-tests/features/reuse-update-details/p2-reuse-update-details.feature#L41-L42) which test the updated details. So maybe we don't need these checks in the e2e tests?
cc @SreeBandla01 @A-Rashid9  @aleemriaz

browser-tests/e2e-tests/.env.template
thebauSoftwire, line 1: The current tests don't really use any sensitive details (except from the orch stub url which would contain the username and password for it) so there wasn't really a need to integrate with a secret manager. 

However, I think this would probably look somewhat like how we're passing in secret values to the api-tests. 
- use a .env file that is never commited
- for running in docker, we could copy these over into the container or explicitly provide as env variables (e.g. running in a container for the CI pipieline)
- for running in secure-pipeline, we can pull from secrets manager with aws cli commands within a bash script

browser-tests/e2e-tests/features/coi-fraud-check.feature
```diff
@@ -0,0 +1,26 @@
+Feature: COI Fraud Check - Given Name Change
+  As a user with an identity verification
+  I want to update my name details
+  So that I can verify my new identity details through the fraud check process
+
+  Background: Initial identity proving journey
+    Given the user starts a new journey in 'build'
+    And the user completes an initial P2 identity journey with expired Alice Parker details
+
```
aleemriaz, line 9: @Build @QualityGateRegressionTest tags missing from the feature and scenario.

browser-tests/e2e-tests/features/web-journey-passport.feature
```diff
@@ -0,0 +1,27 @@
+Feature: E2E Passport Journey
+
+  @Build @Lamdatesta @PYIC-5477 @PYIC-6863 @PYIC-7016
```
aleemriaz: should that be @LambdaTest?

thebauSoftwire: I'm not really sure 🤔 I can't find another instance of it in the ipv-core-tests repo. Do you know if we still need this tag? cc @SreeBandla01 @A-Rashid9

SreeBandla01: I dont think it needed @thebauSoftwire

browser-tests/e2e-tests/features/coi-fraud-check.feature
thebauSoftwire: good spot! Thanks :)

browser-tests/e2e-tests/pages/base-page.ts
aleemriaz: I agree we shouldn't duplicate detailed rendering checks that are already covered in browser and API tests. However, I do think our E2E tests should assert a clear, user-visible outcome to confirm that the journey completed successfully and that data flowed correctly through the system end to end.

Perhaps we can reduce the assertions to focus only on key outcome signals (e.g. the verified identity name or the achieved VOT), rather than checking arbitrary bits of page text that could change over time without affecting behaviour. That way we preserve meaningful end-to-end confidence while avoiding brittleness and unnecessary duplication.

browser-tests/e2e-tests/clients/dcmaw-async-client.ts
```diff
@@ -0,0 +1,77 @@
...
+  userId: string,
+  scenario: string,
+): Promise<string> => {
+  const payload = {
+    user_id: userId,
+    credential_subject: JSON.parse(
+      fs.readFileSync(
+        path.join(
+          __dirname,
+          `../data/async-dcmaw-stub-data/${scenario}/credentialSubject.json`,
```
blakeyp: Thoughts on defining these in TS objects so we can export/import without the overhead of disk I/O during test runs? would also gives us type safety (could use `data-vocab`?)

If this is a big job we could move it to a separate ticket :)

browser-tests/e2e-tests/data/cri-stub-data.ts
```diff
@@ -0,0 +1,72 @@
...
+    },
+    "claimed-identity": {
+      stubData: "Kenneth Decerqueira",
+    },
+    f2f: {
+      stubData: "Kenneth Decerqueira (Valid Passport)",
+      evidenceScores: {
+        strength: "4",
+        validity: "2",
+        biometricVerification: "3",
```
blakeyp: Is there a reason to call it `biometricVerification` over just `verification`? it's ultimately just the "verification" score (it's just that for DCMAW VCs store it in the `biometricVerificationProcessLevel` field)

browser-tests/e2e-tests/features/coi-fraud-check.feature
```diff
@@ -0,0 +1,27 @@
+Feature: COI Fraud Check - Given Name Change
+  As a user with an existing identity and expired fraud details
+  I want to update my name details
+  So that I can verify my new identity details through the fraud check process
+
+  Background: Initial identity proving journey
+    Given the user starts a new journey in 'build'
```
blakeyp: Is there a way we could control the environment through config rather than embedding this in a feature step?

browser-tests/e2e-tests/features/coi-fraud-check.feature
```diff
@@ -0,0 +1,27 @@
...
+    And the user completes an initial P2 identity journey with expired Alice Parker details
+
+  @Build @QualityGateRegressionTest
+  Scenario: Pass successfully for a given name change and show reuse screen
+    When the user starts a new journey in 'build'
+    Then the user should see the 'confirm-your-details' page
+    When the user chooses to update their given name via the app
+    And the user goes through 'DAD' 'iphone' triage
+    And the user submits 'alice-parker-changed-first-name-dvla' details and continues from the 'DAD' journey
+    And the user submits 'alice-parker-changed-first-name' details to the 'driving-licence' CRI stub
```
blakeyp: I'd love to ultimately make these feature scenarios environment-agnostic such that you could run them against dev/build or staging alike, and the lower-level implementation of the step works out whether to submit details via a stub or real CRI pages (based on config). Then this step just becomes something like
`the user completes the Driving Licence CRI` 😍 

I think we're probably a way off from this but maybe we can think about/create some tickets around this as part of our migration of the tests cc @aleemriaz @SreeBandla01 @A-Rashid9

browser-tests/e2e-tests/steps/repeat-fraud-check.steps.ts
```diff
@@ -0,0 +1,96 @@
...
+      "alice-parker-valid",
+      "driving-licence",
+    );
+    // On page-dcmaw-success
+    await pageUtils.getContinueButton().click();
+
+    // submit address details
+    await criStubUtils.submitDetailsToCriStub("alice-parker-valid", "address");
+
+    // submit fraud details
```
blakeyp: Some of these comments are probably not necessary? the page ones may be useful but the other steps look self-descriptive

aleemriaz: yeah I agree, some of those comments do seem unnecessary

browser-tests/e2e-tests/.env.template
blakeyp: Mostly it's just config which we might eventually want to put in AppConfig and have the tests suite integrate with AppConfig directly (with the exception of actual secrets which can go in SecretsManager). (Be nice to move away from the current E2E tests script approach which pulls from SSM via CLI commands)

But agree this env var approach is fine for now :)

browser-tests/e2e-tests/guidance.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Running and Debugging
+
+After modifying any test file, regenerate the spec files before running:
+
+```bash
+npx playwright-bdd --config playwright-bdd.config.ts test
+```
+
+For more information on running the E2E tests, see [these docs](readme.md#-running-tests).
```
blakeyp, line 72: Nice ✨

browser-tests/functional-tests/readme.md
```diff
@@ -0,0 +1,44 @@
+# Functional Tests
+
+The functional tests run Core Front against a browser using Playwright and a fake Core Back provided by imposter.
```
blakeyp, line 3: Be good to add something to distinguish these from the E2E tests e.g. "check that the core-front application runs and handles isolated FE concerns properly (e.g. redirects, language toggle, special case handling)"

browser-tests/package.json
```diff
@@ -3,15 +3,28 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "test": "npx playwright test",
-    "update-snapshots": "npx playwright test --update-snapshots"
+    "test:functional": "npx playwright test --config playwright.config.ts",
+    "update-snapshots": "npx playwright test --config playwright.config.ts --update-snapshots",
+    "test:functional-docker": "docker compose build && docker compose up --exit-code-from browser-tests-tests",
+    "test:e2e": "npx playwright-bdd --config playwright-bdd.config.ts test && npx playwright test --config playwright-bdd.config.ts",
```
blakeyp: Is `npx` necessary if dependencies are added below (and installed) ?

browser-tests/package.json
```diff
@@ -3,15 +3,28 @@
...
   "author": "",
   "license": "MIT",
   "dependencies": {
     "@axe-core/playwright": "^4.11.0",
+    "@cucumber/cucumber": "^10.0.0",
     "@playwright/test": "1.56.1",
-    "dotenv": "^17.3.1"
+    "dotenv": "^17.3.1",
+    "esbuild": "^0.27.2",
+    "pirates": "^4.0.7",
```
blakeyp: is this used?

eslint.config.js
```diff
@@ -49,5 +49,12 @@ module.exports = tsEslint.config(
       "@typescript-eslint/no-explicit-any": "off", // any is convenient for tests
       "@typescript-eslint/no-unused-expressions": "off", // chai assertions are often 'unused' expressions
     }
+  },
+  {
+    // This is to allow console statements when running the E2E tests which is helpful for debugging.
+    files: ["browser-tests/e2e-tests/**/*.ts"],
+    rules: {
+      "no-console": ["error", { allow: ["info", "warn", "error"] }],
+    },
```
blakeyp, line 58: We'll probably want to use a proper logging library eventually (but fine for now)

browser-tests/e2e-tests/data/cri-stub-data.ts
thebauSoftwire: ah I thought it was biometricVerificationProcessLevel for all of the CRIs, I'll just make it verification like you said then, thanks :)

browser-tests/e2e-tests/features/coi-fraud-check.feature
thebauSoftwire: ooh that'd be really nice 🙌 it'd cut down the amount of tests we have defined and make sure we're covering scenarios across multiple environments so we have less gaps in our coverage

browser-tests/package.json
thebauSoftwire: hmm no I don't think so, not sure how they got there, have removed!

browser-tests/e2e-tests/fixtures/index.ts
```diff
@@ -0,0 +1,68 @@
...
+interface Fixtures {
+  scenarioContext: ScenarioContext;
+  pageUtils: PageUtils;
+  criStubUtils: CriStubUtils;
+  orchStubUtils: OrchestratorStubUtils;
+  capturePageDetailsOnFailure: void;
+}
+
+export default baseTest.extend<Fixtures>({
+  scenarioContext: async ({}, use): Promise<void> => {
```
aleemriaz, line 26: Using (_, use) causes Playwright to throw:
`First argument must use the object destructuring pattern: _
   at ../e2e-tests/fixtures/index.ts:24`

Playwright requires fixture functions to use object destructuring for the first argument (even when no fixtures are used). The empty object pattern ({}, use) satisfies Playwright’s fixture API requirements.

aleemriaz: added the comment on the line before to not run the `no-empty-pattern` check for line 27

thebauSoftwire: thank you!!

blakeyp (dismissed):

browser-tests/e2e-tests/features/coi-fraud-check.feature
DanCorderIPV: Making the step decide on whether to use the CRI or stub would be nice, but I'm not sure a lot of the scenarios can be fully agnostic - don't we only have a limited number of users and scenarios that will work in staging due to the test data set up in the CRIs?

browser-tests/e2e-tests/fixtures/cri-stub-fixture.ts
```diff
@@ -0,0 +1,87 @@
...
+    return testDataConfig;
+  };
+
+  const setTestData = async (testDataIdentifier: string): Promise<void> => {
+    await page.locator(TEST_DATA_INPUT).selectOption(testDataIdentifier);
+  };
+
+  const setEvidenceScores = async (scores: EvidenceScores): Promise<void> => {
+    if (scores.strength) {
+      await page.locator("#strength").fill(scores.strength.toString());
```
DanCorderIPV: I know these ids are only used once, but they still feel like they should be constants - especially as some of them already are

browser-tests/e2e-tests/guidance.md
```diff
@@ -18,8 +18,29 @@ Guidelines for writing and maintaining E2E Playwright BDD tests in this director
 
 ## Reusing Step Definitions
 
-- Check `steps/common.steps.ts` before writing new steps — for example, it contains reusable steps for starting journeys, submitting CRI stub details and generic page interactions and assertions.
-- Only create a new step file when the step is **specific to a particular journey** (e.g. `repeat-fraud-check.steps.ts` for RFC-specific interactions, `f2f.steps.ts` for F2F credential assertions).
+The steps folder has the following structure:
+
+```
+steps/                          # Step definitions (.steps.ts) that implement each Gherkin step defined in /features/*.feature
+├── ipv-page-steps/             # Each file represents an IPV Core page. These contain the step deifinitions for all the interactions with that page.
```
blakeyp: typo *definitions

src/test-utils/pages-and-contexts.ts
```diff
@@ -107,4 +107,6 @@ export const pagesAndContexts: Record<string, (string | undefined)[]> = {
   ],
   "we-matched-you-to-your-one-login": [],
   "you-can-change-security-code-method": [],
-};
+} as const;
+
+export type PageName = keyof typeof pagesAndContexts;
```
blakeyp, line 112: Oo nice!

browser-tests/e2e-tests/readme.md
```diff
@@ -0,0 +1,125 @@
+# E2E Playwright BDD Tests
+
+This directory contains the **End-to-End (E2E) test suite** for IPV Core.
```
blakeyp, line 5: Just a thought, for now we should add a note in here saying this is WIP and these tests aren't actually run anywhere yet and that we're migrating from our existing core-tests repo

browser-tests/e2e-tests/steps/ipv-page-steps/ipv-page-reuse.steps.ts
aleemriaz, line 1: This appears to be a duplicate of the correctly named file `page-ipv-reuse.steps.ts`. Merge this file into the other one.

browser-tests/e2e-tests/steps/ipv-page-steps/page-pre-experian-kbv-transition.steps.ts
aleemriaz, line 1: Typo in the file name, per instead of page-pre, and transitions instead of transition. Please rename the file  to `page-pre-experian-kbv-transition.steps.ts`

thebauSoftwire: nice spot!

blakeyp (dismissed):

browser-tests/e2e-tests/steps/ipv-page-steps/ipv-page-reuse.steps.ts
aleemriaz: @thebauSoftwire 👀

thebauSoftwire: ahh thank you 🤦‍♀️

browser-tests/e2e-tests/steps/f2f.steps.ts
```diff
@@ -0,0 +1,29 @@
+import { createBdd } from "playwright-bdd";
+import { expect } from "@playwright/test";
+import fixtures from "../fixtures";
+
+const { Then } = createBdd(fixtures);
+
+// This is specific to the journey in scenario: F2F Passport claim is returned
+// in the user identity response.
+Then(
```
DanCorderIPV: This seems like it should be in the orch-stub steps file?

browser-tests/e2e-tests/steps/repeat-fraud-check.steps.ts
```diff
@@ -0,0 +1,70 @@
...
+      "alice-parker-expired-fraud",
+      "fraud",
+    );
+  },
+);
+
+// This is specific to the journey in scenario: Pass successfully for a given
+// name change and show reuse screen
+Then(
+  "Alison Parker's credentials should be passed to the orch stub",
```
DanCorderIPV: This looks similar to the Kenneth step above and could be commonised in the orch-stub steps?

browser-tests/e2e-tests/steps/f2f.steps.ts
thebauSoftwire: yeah I did debate on putting it there but decided to leave it here since this collection of VCs is fairly specific to the user having gone through a f2f journey

thebauSoftwire: though actually saying that I can just make it a generic step function which takes in arguments 🤦‍♀️ Will make that update (sorry if that's what you were trying to say already!)

browser-tests/e2e-tests/steps/strategic-app.steps.ts
```diff
@@ -0,0 +1,78 @@
+import { createBdd } from "playwright-bdd";
```
DanCorderIPV: Maybe name this something more like `dcmaw-async-stub.steps.ts` ?

So it's clear that these are steps specific to the stub rather than the larger journey.

browser-tests/e2e-tests/steps/composite.steps.ts
```diff
@@ -0,0 +1,70 @@
...
+import { enqueueVcWithScenario } from "../clients/dcmaw-async-client";
+import { selectResidenceLocation } from "./ipv-page-steps/live-in-uk.steps";
+import { selectPhotoIdAvailability } from "./ipv-page-steps/page-ipv-identity-document-start.steps";
+import { selectDevice } from "./ipv-page-steps/pyi-triage-select-device.steps";
+import { selectSmartphone } from "./ipv-page-steps/pyi-triage-select-smartphone.steps";
+
+const { Given, Then } = createBdd(fixtures);
+
+Given(
+  "the user completes an initial P2 identity journey with expired Alice Parker details",
```
DanCorderIPV, line 12: Should this go into a file called something like `composite-steps` or similar?

If I wanted to set up Alice with an expired fraud VC for a different test I might not think to look here and so end up duplicating the logic.

DanCorderIPV (dismissed):

browser-tests/e2e-tests/steps/strategic-app.steps.ts
thebauSoftwire: I've decided to add this to cri.steps.ts since it didn't feel like it warranted its own file anymore

browser-tests/e2e-tests/steps/composite.steps.ts
thebauSoftwire: good point!

DanCorderIPV (approved):

PR #2282, opened 2025-10-06
PYIC-8648: Update triage page links in the login to app flow

AmritSidhu:
### What changed

- Currently, the app download flow includes content suggesting that users can "prove their identity another way" if they choose not to use the app.
- For regular ID proving journeys, this is correct – users may be able to continue via an alternative path.
- For app-only routes (e.g. COI, 6MFC, international addresses, MFA reset), this content is misleading because no alternative route exists.

- The `iphone` and `android` contexts determine which QR code to display. A quick win for adding another context is to append the new context to the existing one and search for keywords instead of matching the context exactly.

### Why did it change

- This creates user frustration and failure, as some users leave the journey assuming an alternative path exists, only to find they cannot complete their task.

### Issue tracking

- [PYIC-8648](https://govukverify.atlassian.net/browse/PYIC-8648)

## Checklists

[PYIC-8648]: https://govukverify.atlassian.net/browse/PYIC-8648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/shared/contextHelper.ts
```diff
@@ -1,13 +1,11 @@
...
-export function validatePhoneType(context?: string): asserts context is string {
-  if (
-    !context ||
-    !([PHONE_TYPES.IPHONE, PHONE_TYPES.ANDROID] as string[]).includes(context)
-  ) {
-    throw new TechnicalError(
-      `Context cannot be parsed as a phone type: ${context}`,
-    );
-  }
+export function validatePhoneType(context?: string): void {
```
blakeyp: It might be cleaner/clearer if we just make this a function that returns a `PHONE_TYPES` given context?

i.e. if context.startsWith("android") -> `PHONE_TYPES.ANDROID`

then `getAppStoreRedirectUrl` can still use an actual `PHONE_TYPES` rather than a context string pretending to be one

src/constants/device-constants.ts
```diff
@@ -1,7 +1,9 @@
 export const PHONE_TYPES = Object.freeze({
   IPHONE: "iphone",
   ANDROID: "android",
-});
+} as const);
+
+export type PhoneType = (typeof PHONE_TYPES)[keyof typeof PHONE_TYPES];
```
blakeyp: We should be able to use a Typescript string enum for this (which can also be used as a type) ?
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

Though we'd probably want to apply the same to all constants so happy for this to be a separate tech debt ticket

blakeyp (dismissed):

src/constants/device-constants.ts
```diff
@@ -1,13 +1,11 @@
...
+  ANDROID = "android",
+}
 
-export type PhoneType = (typeof PHONE_TYPES)[keyof typeof PHONE_TYPES];
-
-export const OS_TYPES = Object.freeze({
-  IOS: "iOS",
-  ANDROID: "Android",
-});
+export enum OS_TYPES {
```
blakeyp: Minor but the convention for enums/type is usually singular noun `PHONE_TYPE` rather than plural

blakeyp (dismissed):

blakeyp (dismissed):

blakeyp (approved):

PR #2286, opened 2025-10-08
PYIC-8674: Add backoff on unexpected polling errors

DanCorderIPV:
DO NOT MERGE UNTIL QA TESTED IN DEV

### What changed

Add backoff on unexpected spinner polling errors

### Why did it change

So we don't instantly fail the process if a user loses connectivity or our end point is rate limited

### Issue tracking

- [PYIC-8674](https://govukverify.atlassian.net/browse/PYIC-8674)

(Frontend changes here: https://github.com/govuk-one-login/govuk-one-login-frontend/pull/283)

[PYIC-8674]: https://govukverify.atlassian.net/browse/PYIC-8674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/javascript/app/dcmaw-async.js
+++ b/assets/javascript/app/dcmaw-async.js
@@ -14,11 +14,14 @@ function generatePollApiFunction(url) {
         throw new Error(`Unexpected status: ${data.status}`);
       })
       .catch((error) => {
-        if (error.name !== "AbortError") {
-          console.error("Unexpected error in pollFunction, backing off: ", error);
-          return 3; // Backoff
+        if (error.name === "AbortError") {
+          // If the error is an AbortError then the user is navigating away from the page (probably a refresh) so we
+          // don't want to trigger success or failure methods.
+          return 2; // Pending
         }
-        return 1; // Failure
+
+        console.error("Unexpected error in pollFunction, backing off: ", error);
+        return 3; // Backoff
       });
   }
 }
--- a/assets/javascript/app/dcmaw-async.js
+++ b/assets/javascript/app/dcmaw-async.js
@@ -14,7 +14,8 @@ function generatePollApiFunction(url) {
       })
       .catch((error) => {
         if (error.name !== "AbortError") {
-          console.error("Error in pollFunction:", error);
+          console.error("Unexpected error in pollFunction, backing off: ", error);
+          return 3; // Backoff
         }
         return 1; // Failure
       });
--- a/assets/javascript/app/dcmaw-async.js
+++ b/assets/javascript/app/dcmaw-async.js
@@ -6,11 +6,12 @@ function generatePollApiFunction(url) {
       .then((data) => {
         if (data.status === "COMPLETED") {
           return 0; // Success
-        } else if (data.status === "ERROR") {
-          return 1; // Failure
-        } else {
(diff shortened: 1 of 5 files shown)


assets/javascript/app/dcmaw-async.js
```diff
@@ -6,15 +6,17 @@ function generatePollApiFunction(url) {
...
         }
+        throw new Error(`Unexpected status: ${data.status}`);
       })
       .catch((error) => {
         if (error.name !== "AbortError") {
-          console.error("Error in pollFunction:", error);
+          console.error("Unexpected error in pollFunction, backing off: ", error);
+          return 3; // Backoff
         }
         return 1; // Failure
```
undermad: Hi, just spotted this.
When error.name is "AbortError", we shouldn’t return a failure, as that would clear session storage. - [here](https://github.com/govuk-one-login/govuk-one-login-frontend/blob/462feb762c503219b25e628cfb6fcab2b04943a1/packages/frontend-ui/frontend-src/spinner/spinner.ts#L205).

This error only occurs on page refresh when we abort a request, and it happens quite sporadically.

undermad (changes requested):

assets/javascript/app/dcmaw-async.js
undermad: It can be easily observed in [template](https://identity.build.account.gov.uk/dev/template/check-mobile-app-result/en) page.
If you open your dev-tools and jump into the session storage page and refresh 10-20 times it will reinitialise the value

undermad (approved):

assets/javascript/app/dcmaw-async.js
```diff
@@ -6,17 +6,22 @@ function generatePollApiFunction(url) {
       .then((data) => {
         if (data.status === "COMPLETED") {
           return 0; // Success
-        } else if (data.status === "ERROR") {
-          return 1; // Failure
-        } else {
+        } else if (data.status === "PROCESSING") {
           return 2; // Pending
+        } else if (data.status === "ERROR") {
+          return 3; // Backoff
```
blakeyp, line 12: ~~Could we use an enum for these responses?~~ it's plain js, ignore me

assets/javascript/app/dcmaw-async.js
```diff
@@ -6,17 +6,22 @@ function generatePollApiFunction(url) {
...
+        if (error.name === "AbortError") {
+          // If the error is an AbortError then the user is navigating away from the page (probably a refresh) so we
+          // don't want to trigger success or failure methods.
+          return 2; // Pending
         }
-        return 1; // Failure
+
+        console.error("Unexpected error in pollFunction, backing off: ", error);
+        return 3; // Backoff
       });
```
blakeyp, line 25: Are we covering network errors, something like `TypeError: Failed to fetch` ?

DanCorderIPV: I think this should catch/handle any error. Although fetch doesn't treat e.g. a 404 as an error, our `then` blocks will fail to parse the response and throw an error instead

blakeyp (approved):

PR #2362, opened 2025-11-05
DFC-692 Set GA4 global data sensitivity flag to false in int and prod

blakeyp:
### What changed

- Set GA4 global data sensitivity flag to false in int and prod

### Why did it change

- D&A have completed their testing in staging and are happy for us to apply the same to int and prod, completing the GA4 upgrade


DanCorderIPV (approved):

blakeyp:
HOLD until D&A confirm

PR #2372, opened 2025-11-10
PYIC-8778: Remove unused published-did-documents bucket and switch all envs to SPOT DID source

AmritSidhu:
### What changed

- Deleted SpotKeysS3Bucket and SpotKeysBucketPolicy resources referencing published-did-documents-* 
- Updated RestDidDocumentResourceS3Role to always reference govuk-one-login-spot-published-keys-${Environment}/did.json 
- Removed conditional logic and references to the deprecated DID bucket

### Why did it change

- The following bucket has been flagged for infringing ‘s3-bucket-ssl-requests-only': published-did-documents-production, because the bucket does not have an SSL deny policy in place.

### Issue tracking

- [PYIC-8777](https://govukverify.atlassian.net/browse/PYIC-8777)

## Checklists

[PYIC-8777]: https://govukverify.atlassian.net/browse/PYIC-8777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1313,6 +1313,11 @@ Resources:
   ApiGatewayStage:
     Type: AWS::ApiGateway::Stage
     Metadata:
+      # Ignore W1001 (conditional Ref) – both deployment resources are mutually exclusive and safely handled by CloudFormation.
+      cfn-lint:
+        config:
+          ignore_checks:
+            - W1001
       checkov:
         skip:
           - id: "CKV_AWS_120"
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -51,11 +51,14 @@ Conditions:
   IsBuild: !Equals [ !Ref AWS::AccountId, "457601271792"]
   IsNotDevelopment: !Not [ !Condition IsDevelopment ]
   IsProduction: !Equals [ !Ref Environment, production ]
-  IsSubscriptionEnvironment: !Or
+  IsStagingIntProd: !Or
     - !Equals [ !Ref Environment, staging ]
     - !Equals [ !Ref Environment, integration ]
     - !Equals [ !Ref Environment, production ]
-  IsNotSubscriptionEnvironment: !Not [ !Condition IsSubscriptionEnvironment ]
+  IsDevOrBuild: !Or
+    - !Equals [ !Ref Environment, build ]
+    - !Equals [ !Ref Environment, dev01 ]
+    - !Equals [ !Ref Environment, dev02 ]
   UsePermissionsBoundary: !Not
     - !Equals [ !Ref PermissionsBoundary, "none" ]
   UseCodeSigning: !Not
@@ -294,12 +297,13 @@ Resources:
     Type: AWS::ApiGatewayV2::ApiMapping
     Condition: IsDevelopment
     Metadata:
+      # Ignore E3005 (linter thinks dependency may not exist) and W1001 (ref may be missing under conditions) - Both safe since they are only created in dev
       cfn-lint:
         config:
           ignore_checks:
             - E3005
             - W1001
-    DependsOn: RestApiGwDeployment202502241200NonSub
+    DependsOn: RestApiGwDeploymentWithoutDidEndpoint202511111337
     Properties:
       DomainName: !Ref DevApiDomain
       ApiId: !Ref RestApiGateway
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -1135,12 +1089,7 @@ Resources:
               - Effect: Allow
                 Action:
                   - "s3:GetObject"
-                Resource: !Sub
-                  - "arn:aws:s3:::${BucketName}/did.json"
-                  - BucketName: !If
-                      - IsSubscriptionEnviroment
-                      - !Sub "govuk-one-login-spot-published-keys-${Environment}"
-                      - !Sub "published-did-documents-${Environment}"
+                Resource: !Sub "arn:aws:s3:::govuk-one-login-spot-published-keys-${Environment}/did.json"
```
blakeyp, line 1103: In the lower envs I suspect this might break as I'm not sure SPOT have this configured in dev/build. I don't _think_ we're using or have a need for the DID endpoint in our dev/build envs - if this is the case we probably want to just put a Condition on this resource itself so it's only created in staging and above (I think we also need to update RestDidDocumentMethod?)

deploy/template.yaml
```diff
@@ -51,10 +51,11 @@ Conditions:
   IsBuild: !Equals [ !Ref AWS::AccountId, "457601271792"]
   IsNotDevelopment: !Not [ !Condition IsDevelopment ]
   IsProduction: !Equals [ !Ref Environment, production ]
-  IsSubscriptionEnviroment: !Or
+  IsSubscriptionEnvironment: !Or
```
blakeyp: I think this was named because staging and above are the Splunk subscription envs, but given this is used in multiple use cases it would be clearer if it was just named what it actually is i.e. `IsStagingIntProd` ?

deploy/template.yaml
```diff
@@ -51,10 +51,11 @@ Conditions:
...
     - !Equals [ !Ref Environment, staging ]
     - !Equals [ !Ref Environment, integration ]
     - !Equals [ !Ref Environment, production ]
+  IsNotSubscriptionEnvironment: !Not [ !Condition IsSubscriptionEnvironment ]
```
blakeyp: Prefer this to be explicitly *is dev/build* ?

deploy/template.yaml
```diff
@@ -292,13 +293,18 @@ Resources:
   DevApiMapping:
     Type: AWS::ApiGatewayV2::ApiMapping
     Condition: IsDevelopment
-    DependsOn: RestApiGwDeployment202502241200
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks:
+            - E3005
+            - W1001
```
blakeyp, line 305: Can we add a comment explaining why we're ignoring

deploy/template.yaml
```diff
@@ -1213,13 +1169,26 @@ Resources:
         Types:
           - REGIONAL
 
-  RestApiGwDeployment202502241200:
+  # For environments that have the DID document (staging, integration, production)
+  RestApiGwDeployment202502241200Sub:
```
blakeyp: Maybe rename this `WithDidEndpoint` ? and we should probably update the `YYYYMMDDHHMM` timestamp

(plus same for "without")

deploy/template.yaml
```diff
@@ -1213,13 +1169,26 @@ Resources:
         Types:
           - REGIONAL
 
-  RestApiGwDeployment202502241200:
+  # For environments that have the DID document (staging, integration, production)
```
blakeyp: to clarify a little - "that host SPOT's DID endpoint" ?

deploy/template.yaml
```diff
@@ -1213,13 +1169,26 @@ Resources:
...
+
+  # For dev/build environments that don’t have the DID document
+  RestApiGwDeployment202502241200NonSub:
+    Condition: IsNotSubscriptionEnvironment
+    DependsOn:
+      - RestApiGatewayMethod
+    Type: AWS::ApiGateway::Deployment
+    Properties:
+      RestApiId: !Ref RestApiGateway
+      Description: "Deployment without DID document for non-subscription environments"
```
blakeyp: Again prefer to refer to actual envs rather than "subscription"/"non-subscription" which is not well known

blakeyp (approved):

PR #2418, opened 2025-11-25
PYIC-8776: add new page for cross-browser problem

thebauSoftwire:
### What changed

- added new problem-different-browser page
- add support for routing from problem-different-browser page to returning to the RP

### Why did it change

- we currently silently return the user back to the RP in a cross-browser scenario and the user is given little or misleading information about why this happened. To help ease frustration, we show them this new error page just before they go back to the RP

### Issue tracking

- [PYIC-8776](https://govukverify.atlassian.net/browse/PYIC-8776)

## Checklists

[PYIC-8776]: https://govukverify.atlassian.net/browse/PYIC-8776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


browser-tests/imposter/config/api-config.yaml
```diff
@@ -688,3 +688,40 @@ resources:
...
+      currentPage: "cross-browser-problem"
+    response:
+      statusCode: 200
+      # In reality, core-back will route to the RP, but
+      # for this test, we just want to check that
+      # core-front makes the appropriate request to core-back
+      content: |
+        {
+          "page": "page-multiple-doc-check"
+        }
```
hugh-emerson, line 729: Would it be clearer/more intuitive to do something like this? Maybe this is a misuse of "client" and there's something better though

```
          "client": {
            "redirectUrl": "https://example.com/return-to-rp"
          }
```

We seem to use example.com elsewhere in these tests so I've suggested that for consistency. However, I don't love the idea of the tests loading up a 3rd party site (example.com). Alternatively, can you do something like this?

```
"redirectUrl": "/page-ipv-success?test=crossBrowserProblem"
```

or even redirecting to something more clearly fake like /dev/template?

thebauSoftwire: good point, thanks! I've also updated the existing client redirect test to not use example.com

blakeyp (dismissed):
QA in dev before merging

hugh-emerson (dismissed):
👍

DanCorderIPV (approved):

PR #2443, opened 2025-12-08
PYIC-8790: Add progress spinner

AmritSidhu:
### What changed

- Replace 'Continue...' button with progress spinner component on 'page-ipv-success' and 'page-ipv-reuse' pages
- Default content and translations are already built in to the component. The component is actually a link masquerading as a button (it takes an href) so we’ll need to use our journey GET mechanism to simulate a ‘next’ event.
- gate progress button behind feature flag

### Why did it change

- To ensure we can handle any potential latency / wait time gracefully this story is to replace the current Continue to service button on those two pages with a progress button component that shows a dynamic ‘waiting’ progress element.

### Issue tracking

- [PYIC-8790](https://govukverify.atlassian.net/browse/PYIC-8790)

## Checklists

[PYIC-8790]: https://govukverify.atlassian.net/browse/PYIC-8790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blakeyp (dismissed):
Looks good but hold merge until QA complete in dev

blakeyp (dismissed):

deploy/template.yaml
```diff
@@ -228,6 +236,7 @@ Mappings:
       spinnerRequestLongWaitInterval: 60000
       mamSpinnerRequestTimeout: 600000
       dadSpinnerRequestTimeout: 1800000
+      enableProgressSpinnerButton: false
```
thebauSoftwire, line 239: is there a reason we don't want a progressSpinnerButtonTimeout in int/prod?

views/ipv/page/page-ipv-reuse.njk
```diff
@@ -65,5 +66,32 @@
       </div>
     </details>
 
-    {{ journeyNextForm(pageId, csrfToken, 'general.buttons.continueToService' | translate ) }}
+    {% if enableProgressSpinnerButton %}
+      {# New progress button (experiment) #}
+      {% set customTimeout = progressSpinnerButtonTimeout %}
```
thebauSoftwire, line 71: out of curiosity, if we don't set a value of `progressSpinnerButtonTimeout`, what value does `customTimeout` take and does the frontendUiProgressButton component handle it?

AmritSidhu: @thebauSoftwire If the value is undefined (because it hasn’t been set), it defaults to the normal timing of 5 seconds, then transitions and waits another 5 seconds. This isn’t long enough in lower environments, which is why there was pushback to add support for a custom timeout.

Because this is currently an untested feature, Patrick asked that it only be enabled up to staging. For now, int/prod will continue to use the new button while we test through staging.

There will be a follow-up ticket to clean this up and enable the functionality for int/prod. Hopefully that answers your above comment too 🙂

thebauSoftwire: thanks for clarifying :)

thebauSoftwire (approved):

package.json
```diff
@@ -76,14 +76,14 @@
...
-    "@aws-sdk/client-ssm": "3.981.0",
+    "@aws-sdk/client-dynamodb": "3.980.0",
+    "@aws-sdk/client-ssm": "3.980.0",
     "@govuk-one-login/data-vocab": "2.0.0",
     "@govuk-one-login/frontend-analytics": "4.0.5",
     "@govuk-one-login/frontend-device-intelligence": "1.2.0",
     "@govuk-one-login/frontend-language-toggle": "2.2.1",
     "@govuk-one-login/frontend-passthrough-headers": "1.3.2",
-    "@govuk-one-login/frontend-ui": "4.0.1",
+    "@govuk-one-login/frontend-ui": "4.2.2",
```
blakeyp, line 86: I assume the updates to all the snapshots are just an artefact of bumping this package?

views/ipv/page/page-ipv-reuse.njk
```diff
@@ -65,5 +66,32 @@
...
+        translations: translations.translation.progressButton,
+        href: "/ipv/journey/" + pageId + "/next",
+        errorPage: "/ipv/page/pyi-technical",
+        text: ('general.buttons.continueToService' | translate),
+        customDevErrorTimeout: customTimeout
+      }) }}
+    {% else %}
+      {# Original button (control) #}
+      {{ journeyNextForm(pageId, csrfToken, 'general.buttons.continueToService' | translate ) }}
+    {% endif %}
```
blakeyp, line 83: Before merging it's probably worth testing both a dev deployment with enableProgressSpinnerButton=true and a dev deployment with enableProgressSpinnerButton=false (for regression) ?

package.json
AmritSidhu: Yes this is a result of the package bump

views/ipv/page/page-ipv-reuse.njk
AmritSidhu: - I have tested true/false and undefined in my dev env for enableProgressSpinnerButton
- I have also tested different times and undefined in my dev env for progressSpinnerButtonTimeout

PR #2458, opened 2025-12-11
SPT-1693: Allow access to published keys KMS key

chris-dco:
### What changed

The published keys bucket is migrating from using `SSE-S3` to `KMS-S3` encryption. This change adds the ability for the `RestDidDocumentResourceS3Role` to access the encryption key in the published bucket account. A matching change is being made in the account to allow IPV Core to perform a Decrypt operation using the key.

### Why did it change

It has been highlighted that this change needs to be made on the published keys bucket in order to following security best practice and its been highlight by AWS Config that these parameters are not enabled on this bucket.

### Issue tracking

- [SPT-1693](https://govukverify.atlassian.net/browse/SPT-1693)

## Checklists

[SPT-1693]: https://govukverify.atlassian.net/browse/SPT-1693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -105,6 +105,7 @@ Mappings:
       mamSpinnerRequestTimeout: 300000
       dadSpinnerRequestTimeout: 300000
       frontendAutoScalingMinCount: 4
+      publishedKeysAccountId: ""
     "175872367215": # core-dev02
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
@@ -135,6 +136,7 @@ Mappings:
       mamSpinnerRequestTimeout: 300000
       dadSpinnerRequestTimeout: 300000
       frontendAutoScalingMinCount: 4
+      publishedKeysAccountId: ""
     "457601271792": # Build
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
@@ -165,6 +167,7 @@ Mappings:
       mamSpinnerRequestTimeout: 300000
       dadSpinnerRequestTimeout: 300000
       frontendAutoScalingMinCount: 6
+      publishedKeysAccountId: ""
     "335257547869": # Staging
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
@@ -195,6 +198,7 @@ Mappings:
       mamSpinnerRequestTimeout: 300000
       dadSpinnerRequestTimeout: 900000
       frontendAutoScalingMinCount: 4
+      publishedKeysAccountId: "444977453444"
     "991138514218": # Integration
       lb500ErrorLimit: 2
       lb500ErrorWindow: 300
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -1089,7 +1095,14 @@ Resources:
               - Effect: Allow
                 Action:
                   - "kms:Decrypt"
-                Resource: !GetAtt S3KmsKey.Arn
+                Resource:
+                  - !GetAtt S3KmsKey.Arn
+                  - !Sub
+                    - "arn:aws:kms:${AWS::Region}:${CoreAccountId}:key/*"
```
chrisclayson: I'm a little nervous about the `*` here. 

Can we, either:
a) split your other [PR](https://github.com/govuk-one-login/ipv-spot-key-management/pull/341) to deploy the key first, so we can put the explicit key ARNs in this policy
OR
b) add a condition to this policy that allows use only of keys with a specific tag or alias. I think you could use the `kms:ResourceAliases` condition here.

chrisclayson: Amend: I can see you've actually already deployed the keys. So you could just get the ARNs into here.

I think you could also add the `kms:ViaService` condition to ensure that the key only allowed when used with S3 ... I think .........

chrisclayson: Not sure you should use `CoreAccountId` here, thats a bit confusing, might be better to `PublishedKeysAccountId` to match the mapping name

chris-dco: Notes: Add condition kms:ViaService and add alias as well. Will need to add alias to the other side. Will need to add alias before either PR goes through.

chrisclayson (dismissed):

chrisclayson (dismissed):

deploy/template.yaml
```diff
@@ -1089,7 +1095,21 @@ Resources:
               - Effect: Allow
                 Action:
                   - "kms:Decrypt"
-                Resource: !GetAtt S3KmsKey.Arn
+                Resource:
+                  - !GetAtt S3KmsKey.Arn
+                  - !Sub
+                    - "arn:aws:kms:${AWS::Region}:${PublishedKeysAccountId}:key/*"
```
blakeyp: I'm wondering if the "" values for `PublishedKeysAccountId` in the lower envs is going to cause stack deployment validation to fail 🤔

chrisclayson: Good point. I'll wrap that statement in an `!If`

chrisclayson: Oh ... I don't need to, the whole role is set with the condition of `IsStagingIntProd`, so it won't try to get created in dev/build.

chrisclayson (approved):

deploy/template.yaml
blakeyp: ah yeah I missed that

blakeyp (approved):

PR #2479, opened 2025-12-18
bau: Add missing await key word.

undermad:
### What changed

- Added missing await to wait for backend response

### Why did it change
Over the last three days, we have observed an increased number of 502 errors returned from the ALB. An analysis of CloudWatch logs shows that all affected requests originate from core-front and report the same error message:

`
Error: Unrecognised response type received from core-back: {"errorMessage":"Missing ipv session id header","errorCode":1010,"statusCode":400}
at handleBackendResponse (/app/src/app/ipv/middleware.ts:157:9)
at processAction (/app/src/app/ipv/middleware.ts:100:37)
at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
`

These errors started appearing after the core-back [PR](https://github.com/govuk-one-login/ipv-core-back/pull/3598) that redirects users to the problem-different-browser page. Route responsible for displaying this page is missing away key word.

The presence of processTicksAndRejections in the stack trace indicates that a Promise is not being handled correctly. This likely creates a race condition, where the request to the backend is not awaited, allowing the Express response to be returned before the required headers are attached.

This PR addresses the issue by ensuring the Promise is properly awaited, which should prevent the race condition and hopefully resolve the observed 502 errors.

## Checklists

### DIFF
--- a/src/app/ipv/middleware.ts
+++ b/src/app/ipv/middleware.ts
@@ -422,7 +422,7 @@ export const handleCrossBrowserJourneyActionRequest: RequestHandler = async (
   }
   // We return directly to the Oauth client when continuing from
   // the problem-different-browser page
-  return processAction(req, res, BUILD_CLIENT_OAUTH_RESPONSE_ACTION, pageId);
+  await processAction(req, res, BUILD_CLIENT_OAUTH_RESPONSE_ACTION, pageId);
 };
 
 export const handleJourneyActionRequest: RequestHandler = async (req, res) => {


src/app/ipv/middleware.ts
```diff
@@ -422,7 +422,12 @@ export const handleCrossBrowserJourneyActionRequest: RequestHandler = async (
   }
   // We return directly to the Oauth client when continuing from
   // the problem-different-browser page
-  return processAction(req, res, BUILD_CLIENT_OAUTH_RESPONSE_ACTION, pageId);
+  return await processAction(
+    req,
+    res,
+    BUILD_CLIENT_OAUTH_RESPONSE_ACTION,
+    pageId,
+  );
```
thebauSoftwire: good spot! Now that I'm looking back, I don't actually think this needs a return statement since `processAction` returns a promise void and I don't think express does anything with that returned value.

(sorry, I missed this when I added this 🤦‍♀️ )

src/app/ipv/middleware.ts
```diff
@@ -74,7 +74,7 @@ const journeyApi = async (
     action = action.substring(8);
   }
 
-  return postJourneyEvent(req, action, currentPageId);
+  return await postJourneyEvent(req, action, currentPageId);
```
thebauSoftwire: I don't think we need this here since `journeyApi` isn't a middleware function (it's called in other middleware and is awaited within them) and it's already an async method meaning it'll wrap it's return value in a Promise anyway. So, typescript will just unwrap the Promise<void> returned from `postJourneyEvent` (bc of the `await`) then wrap it in a promise again (because of the `async`) and when this function is called, unwrap again with another `await`

undermad: Ah, you are right. Updated PR

src/app/ipv/middleware.ts
undermad: You are right here as well, it looks like it is the end of the middleware chain. But we also missed that in the PR review.

thebauSoftwire (approved):

PR #2481, opened 2025-12-18
BAU: add missing return statement inside handleCrossBrowserJourneyActionRequest to prevent calling core-back again

thebauSoftwire:
### What changed

- add missing return statement to exit the `handleCrossBrowserJourneyActionRequest` middleware when the user continues on from the problem-different-browser page but the sessionPageId is not equal

### Why did it change

- We have been seeing a lot of 502s from the ALB since the new problem-different-browser page was added. They are being caused by repeat (unintended) calls to core-back which corrupts the socket when a user has continued on from the page but the user goes back

## Checklists


undermad (approved):

blakeyp:
Linking incident INC0016299 for posterity

PR #2570, opened 2026-02-09
Revert "Revert "PYIC-8790: Add progress spinner (#2443)""

AmritSidhu:
Reverts govuk-one-login/ipv-core-front#2565


src/app/ipv/router.ts
```diff
@@ -53,11 +54,7 @@ router.get(
 router.get(
   getPagePath(IPV_PAGES.PYI_TECHNICAL),
   csrfSynchronisedProtection,
-  (req, res) => {
-    return res.render(getPagePath(IPV_PAGES.PYI_TECHNICAL), {
-      context: "unrecoverable",
-    });
-  },
+  renderPyiTechnicalWithContextOverride,
```
blakeyp: I remembered we have something similar to this for the DAD redirect to error page - https://github.com/govuk-one-login/ipv-core-front/blob/6fcee6785feecc1a3e977910b0adc135c8e9ea70/src/app/ipv/middleware.ts#L325. Could we handle this in the same way there?

src/app/ipv/middleware.ts
```diff
@@ -309,6 +309,23 @@ const validateSessionAndPage = async (
...
+    req.session.currentPage = pageId;
+
+    return (
+      res.render(getIpvPageTemplatePath(pageId), {
+        pageId,
+        csrfToken: req.csrfToken?.(true),
+        context: "unrecoverable",
+      }),
+      false
+    );
```
blakeyp: For consistency can we separate this out as just a res.render followed by return false? the comma operator probably isn't widely known so makes it less easy to read/understand this

blakeyp (approved):

PR #2575, opened 2026-02-10
PYIC-8790: add progressSpinnerButtonTimeout to int and prod mapping set to default 5s

thebauSoftwire:
### What changed

- add progressSpinnerButtonTimeout to int and prod mapping set to default 5s

### Why did it change

- deployment is failing in int and prod because CF is failing to get the mapping values for progressSpinnerButtonTimeout

### Issue tracking

- [PYIC-8790](https://govukverify.atlassian.net/browse/PYIC-8790)

## Checklists

[PYIC-8790]: https://govukverify.atlassian.net/browse/PYIC-8790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -55,6 +55,9 @@ Conditions:
     - !Equals [ !Ref AWS::AccountId, "175872367215"]
   IsNotDevelopment: !Not [ !Condition IsDevelopment ]
   IsProduction: !Equals [ !Ref Environment, production ]
+  IsIntOrProd: !Or
+    - !Equals [ !Ref Environment, production ]
+    - !Equals [ !Ref Environment, integration ]
   IsStagingIntProd: !Or
     - !Equals [ !Ref Environment, staging ]
     - !Equals [ !Ref Environment, integration ]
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -236,6 +236,7 @@ Mappings:
       spinnerRequestLongWaitInterval: 60000
       mamSpinnerRequestTimeout: 600000
       dadSpinnerRequestTimeout: 1800000
+      progressSpinnerButtonTimeout: 5000
```
thebauSoftwire: setting this to the default value for now

undermad (dismissed):

deploy/template.yaml
blakeyp: I think the default is actually 10s (not that it matters here) - could we add a comment that this value isn't used in int/prod?

blakeyp: https://github.com/govuk-one-login/govuk-one-login-frontend/blob/b6e88dc5f2510a9935fb18ff1d117c009957bebb/packages/frontend-ui/frontend-src/progress-button/progress-button.ts#L1

blakeyp (approved):

PR #2698, opened 2026-03-26
PYIC-8718: switch from context to pageContext object

thebauSoftwire:
❗❗ **This should be merged in AFTER the banner config has been removed (as it's not backwards compatible)** ❗ ❗ 
**Note: merging this PR in will make the switch in prod.**

### What changed

- update nunjucks templates to use new `pageContext` object instead of context
- replace context with pageContext everywhere in core-front
- update development middleware and `/dev/all-templates` to support new pageContext
- update notification banner handler to use new pageContext object (including updating the config shape)
- update e2e tests to work better with dev-deployed instances

| new /dev/all-templates context |
| -| 
| Screenshot 2026-03-26 at 14 00 25 |

**Note on documentation**: I'll update the journey map docs in the clean-up PR to remove the existing context on page states so I can get feedback on this PR first.

### Why did it change

core-back is now sending a `pageContext` object intended to replace the existing `context` for page step responses. 

We currently use a string context for routing to variants of a page. However, this has limitations because it is only a string. For example, [pyi-triage-desktop-download-app](https://identity.build.account.gov.uk/dev/template/pyi-triage-desktop-download-app/en?context=android) has four different context values: android, iphone, android-appOnly and iphone-appOnly which could make the logic in core-front messy/confusing. Instead, we can turn it into an object:

```
{
  isAppOnly: true/false,
  smartphone: iphone/android
}
```

Core-back PRs:
- https://github.com/govuk-one-login/ipv-core-back/pull/3821
- https://github.com/govuk-one-login/ipv-core-back/pull/3830

### Issue tracking

- [PYIC-8718](https://govukverify.atlassian.net/browse/PYIC-8718)

## Checklists

[PYIC-8718]: https://govukverify.atlassian.net/browse/PYIC-8718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/test-utils/pages-and-contexts.ts
thebauSoftwire, line 1: I've kept the original `context` values here as keys so that the browser tests don't create new screenshot files and we can properly check for regressions in the snapshot tests.

I'll be creating a new PR to do some clean up but I'm debating on whether we should keep this extra "descriptive" key or not though... I think it's useful for adding context to the different pageContext object variants for a page so useful for someone who isn't a dev. But, the pageContexts we have atm are quite simple and descriptive enough so might not need it anyway. Having to maintain a descriptive "key" might also not fit the dynamic journey engine 🤔 

Would be great to hear opinions on this!

src/config/nunjucks.ts
```diff
@@ -37,24 +60,100 @@ export const configureNunjucks = (
...
+    } catch {
+      return str;
+    }
+  });
+
+  // TODO PYIC-8718: remove this once the ipv-core-base.njk file for core in frontend-ui
+  // has been updated to remove the use of translateWithContextOrFallback
   nunjucksEnv.addFilter(
     "translateWithContextOrFallback",
     function (this: FilterContext, key, context, options) {
```
thebauSoftwire, line 150: I'm planning on replacing the `translateWithContextOrFallback` in the [govuk-one-login-frontend repo](https://github.com/govuk-one-login/govuk-one-login-frontend/blob/caec105bb1a73c40309bc270ba8ccb4f462e44ef/packages/frontend-ui/components/bases/ipv-core/ipv-core-base.njk#L49) with just a normal `translate` so we can remove this function. 

I've updated any pages with a "title" variant with something like
```
{% set pageTitleKey = 'pages.noPhotoIdSecurityQuestionsFindAnotherWay.title' | fullKeyWithContext(pageContext, 'reason') %}
```
so it's backwards compatible.

Ideally though, we don't have our base njk file depending on nunjucks filters in this repo or move our base.njk file to this repo instead 🤔 (and the base njk file in the govuk-one-login-frontend repo be bare bones to include all the analytics stuff)

src/test-utils/pages-and-contexts.ts
thebauSoftwire: If I got rid of the "descriptive key" it also means I can get rid of the weird way I'm trying to access the actual context object e.g.
```
const contextValue = Object.values(contextScenario)[0];
```

browser-tests/e2e-tests/steps/orch-stub.steps.ts
```diff
@@ -12,7 +12,7 @@ const sleep = (ms: number): Promise<void> =>
 // case where we're waiting for an async VC to be ready which might require multiple attempts.
 const retryUntil = async (
   action: () => Promise<void>,
-  maxWaitMs = 20000,
+  maxWaitMs = 30000,
```
thebauSoftwire, line 15: increasing this for dev-deployed instances which suffer more from cold starts

src/app/development/middleware.ts
```diff
@@ -37,32 +41,39 @@ export const allTemplatesGet: RequestHandler = async (req, res) => {
...
+    templatesWithContextRadioOptions[page] = contexts.map(
+      (contextLabelAndValue) => {
+        if (contextLabelAndValue === NO_CONTEXT_VARIANT) {
+          return { text: "No context", value: "" };
+        }
+
+        const label = Object.keys(contextLabelAndValue)[0];
+        return {
+          text: label,
+          value: JSON.stringify(contextLabelAndValue[label]),
```
thebauSoftwire, line 59: If I got rid of the "descriptive key" from the `pagesAndContexts` object, I'll either:
- just use the actual object as a label (which removes the need for a details sections showing the valid objects) but this doesn't style nicely the input stretches for big objects (since the string is longer) edit: I think I might be able to override styling to fix this
- have dropdowns for each context key and the valid values it can take e.g. dropdown for "smartphone" and dropdown for "isAppOnly". This will need a details components to show which combinations are valid. I think this fits more with the future dynamic routing stuff but might be too much right now and confusing to see which pages (+ variants) are currently being used.

But keen to get thoughts/ideas on this!

src/test-utils/pages-and-contexts.ts
thebauSoftwire: with the contexts now being objects, and us heading towards a dynamic journey engine, this'll be harder to keep in-sync with what we think core-back will return. It'd be nice to have a shared package between the two to keep these in-sync but I think this would bloat the ticket

thebauSoftwire: after speaking with Dan, we've decided to keep these "descriptive" keys for now since they're helpful for labelling the object variations, and it's easier for non-devs to know what those contexts represent. This can be updated much later to support/work better with the dynamic engine work

src/app/development/middleware.ts
thebauSoftwire: decided to stick with the "descriptive key" in the `pagesAndContexts` so I won't need to update this now, unless anyone thinks the /dev/all-templates page should be re-formatted!

src/config/nunjucks.ts
```diff
@@ -37,24 +62,102 @@ export const configureNunjucks = (
...
 
-      const pascalContext = kebabCaseToPascalCase(context);
+  /**
+   * If context:
+   *   - does not exist in the pageContext object
+   *   - has a boolean value, but it's false
+   *  then throw an error as there is no fallback
+   */
+  nunjucksEnv.addFilter(
+    "translateWithPageContext",
```
blakeyp, line 101: I'm a bit unclear how do we know when to use this vs translateWithPageContextOrFallback ?

src/app/development/middleware.ts
blakeyp, line 1: What will the all-templates page URLs look like now? it's quite nice for them to be human readable as a way of referencing page variants (page URL slug + context) but it looks like pageContext is going to be a URL-encoded JSON string? wonder if we can either
- add the descriptive `context` key too
- (perhaps more future proof) just flatten the page context key-values into the query params themselves e.g. `?isAppOnly=true&device=iphone` though this relies on page context always being simple key-value pairs (which I think it should be?)

src/app/shared/contextHelper.ts
```diff
@@ -1,18 +1,18 @@
...
-export function getPhoneType(context?: string): PHONE_TYPE {
-  if (!context || typeof context !== "string") {
-    throw new TechnicalError(`Invalid phone type context: ${context}`);
+export function getPhoneType(smartphone?: string): PHONE_TYPE {
+  if (!smartphone || typeof smartphone !== "string") {
+    throw new TechnicalError(`Invalid phone type context: ${smartphone}`);
   }
 
-  if (context.startsWith(PHONE_TYPE.IPHONE)) {
+  if (smartphone.startsWith(PHONE_TYPE.IPHONE)) {
```
blakeyp: I think the `startsWith` was to handle the composite context key so this can just be equals now?
Could probably even make this an assertion function `value is PHONE_TYPE` and then just use the value as is

src/app/development/middleware.ts
blakeyp: TIL a query param with no value is valid, so for booleans it could even be `?isAppOnly&device=iphone`

src/config/nunjucks.ts
thebauSoftwire: I was thinking it might be helpful for pages that require a context e.g. `pyi-triage-select-smartphone` which doesn't have a no-context-variant so the fallback wouldn't be valid

src/app/development/middleware.ts
thebauSoftwire: that's a good point, I'll flatten it out. I guess another thing is, do we think we'll need to add more query parameters in future that aren't page contexts? We pass the contexts in as a single parsed object e.g. something like this works for all pages:
```
contexts: JSON.parse(req.query.pageContexts)
```
If we flatten them out, we'll probably need to pass in all query params into the template (which I think is fine for now since we don't do anything else with query params) but if we added a new query param in future unrelated to the page contexts, we'll need to make sure it doesn't unintentionally affect the nunjucks templates

blakeyp (approved):

PR #2740, opened 2026-04-09
PYCI-8942: Bump progress spinner timeout to 40s - all envs.

undermad:
### What changed

- Configure the progress button custom timeout at 40 seconds in all environments.

### Why did it change

- The component has a deliberately built-in 10 seconds timeout for user accessibility, however in rare cases (such as at times of low traffic) we expect the redirect may take longer than 10s and we don’t want to unnecessarily block users.

### Issue tracking

- [PYIC-8942](https://govukverify.atlassian.net/browse/PYIC-8942)

## Checklists

[PYIC-8942]: https://govukverify.atlassian.net/browse/PYIC-8942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -107,7 +107,7 @@ Mappings:
       spinnerRequestLongWaitInterval: 60000
       mamSpinnerRequestTimeout: 300000
       dadSpinnerRequestTimeout: 300000
-      progressSpinnerButtonTimeout: 30000
+      progressSpinnerButtonTimeout: 40000
       enableProgressSpinnerButton: true
       frontendAutoScalingMinCount: 4
       publishedKeysAccountId: ""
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -272,7 +272,7 @@ Mappings:
       spinnerRequestLongWaitInterval: 60000
       mamSpinnerRequestTimeout: 600000
       dadSpinnerRequestTimeout: 1800000
-      progressSpinnerButtonTimeout: 10000 # This value isn't used in int or prod but CF requires these keys to be present in the mapping
+      progressSpinnerButtonTimeout: 40000
```
blakeyp, line 275: oo we also need to update the conditional so this value actually gets used: https://github.com/govuk-one-login/ipv-core-front/blob/d2b3f90682b6fbedeaf1bf21ddcd4e8a7fd27b96/deploy/template.yaml#L949-L955

blakeyp: ah I see you've done it in the other PR

blakeyp (approved):

deploy/template.yaml
undermad: Yes, then I check in staging if I can use network throttling in the browser tools to test it somehow.

PR #2773, opened 2026-04-24
PYIC-8942 Improve special handling for progress spinner button timeouts

blakeyp:
### What changed

- Set previousIpvSessionId when deleting ipvSessionId on Orch handoff and include previousIpvSessionId as ref to old session in timeout error log
- Update timeout error query value to explicitly denote spinner timeout
- 'Protect' the timeout error route by only handling if in an expected page state (i.e. reuse/success page or Orch handoff state) and include 'fromPage' in log
    - Timeout should only really occur after redirect to Orch (i.e. in Orch handoff state)
    - If it occurs directly from our reuse/success pages something is probably wrong with the component/timer

### Why did it change

- We didn't have a link to the ipvSessionId in the timeout error log nor protect the special route in any way, making it hard to see through the noise and properly trace user journeys. Users or bots may refresh/hit the error route directly, creating a lot of noise in the logging

### Issue tracking

- [PYIC-8942](https://govukverify.atlassian.net/browse/PYIC-8942)

## Checklists

[PYIC-8942]: https://govukverify.atlassian.net/browse/PYIC-8942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/views/ipv/page/page-ipv-success.njk
+++ b/views/ipv/page/page-ipv-success.njk
@@ -22,7 +22,7 @@
       id: "submitButton",
       translations: translations.translation.progressButton,
       href: "/ipv/journey/" + pageId + "/next",
-      errorPage: '/ipv/page/pyi-technical?isUnrecoverable',
+      errorPage: '/ipv/page/pyi-technical?spinnerTimeout',
       text: ('general.buttons.continueToService' | translate),
       customDevErrorTimeout: customTimeout
     }) }}
--- a/src/app.ts
+++ b/src/app.ts
@@ -51,6 +51,7 @@ declare global {
 declare module "express-session" {
   interface SessionData {
     ipvSessionId?: string;
+    previousIpvSessionId?: string;
     clientOauthSessionId?: string;
     currentPage?: string;
     pageContext?: Record<string, unknown>;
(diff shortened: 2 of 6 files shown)


views/ipv/page/page-ipv-success.njk
```diff
@@ -22,7 +22,7 @@
       id: "submitButton",
       translations: translations.translation.progressButton,
       href: "/ipv/journey/" + pageId + "/next",
-      errorPage: '/ipv/page/pyi-technical?isUnrecoverable',
+      errorPage: '/ipv/page/pyi-technical?spinnerTimeout',
```
undermad, line 25: Is there a scenario where this component throw error within the timeout window? If so, this would be inaccurate to set the spinnerTimeout param. Ideally if component accept two possible scenarios. Normal error and timeout error.

undermad (commented):
Looking great and I like the solution. Just a clarification question about the errorPage and if we should request additional field from component?

views/ipv/page/page-ipv-success.njk
blakeyp: As far as I can tell the component only uses the error page on timeout: https://github.com/govuk-one-login/govuk-one-login-frontend/blob/ac8ec352fcc09fc46f4eb1672eb6252b438b821b/packages/frontend-ui/frontend-src/progress-button/progress-button.ts#L147
(but of course as we've seen it may be a premature timeout)

undermad (approved):

PR #2790, opened 2026-04-30
Pyic 9050 add new mitigation screens

DanCorderIPV:
### What changed

Add new pages

As I got the translations very quickly I have applied them to this PR.

I created `check-translations.ts` with Kiro and I think it will be useful to keep around. UCD are discussing whether they want to make Welsh text consistent. If they decide not to then we should remove the consistency check from the script.

### Why did it change

We need them for upcoming mitigation routes

### Issue tracking

- [PYIC-9050](https://govukverify.atlassian.net/browse/PYIC-9050)

## Checklists

[PYIC-9050]: https://govukverify.atlassian.net/browse/PYIC-9050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.kiro/steering/tech.md
```diff
@@ -0,0 +1,18 @@
...
+The base template for `core-front` is actually stored in the `@govuk-one-login/frontend-ui`
+CSS is generated using SASS
+
+# Testing
+Unit tests use Mocaha, Sinon, and Chai.
+Browser tests have their own folder and use PlayWright
+
+# Misc
+Check package.json for the full list of packages used by the site.
+Check README files for more information about the project.
```
blakeyp, line 18: Are the .kiro files intentionally committed? Probably fine but may not be totally clear without the context of what kiro is/does - I assume this is basically prompt input?

views/ipv/page/retry-prove-identity-app.njk
```diff
@@ -0,0 +1,80 @@
...
+      fieldset: {
+        legend: {
+          text: 'pages.retryProveIdentityApp.content.subHeading' | translate,
+          classes: "govuk-fieldset__legend--m"
+        }
+      },
+      items: [
+        {
+          value: "useApp",
+          text: 'pages.retryProveIdentityApp.content.formRadioButtons.useApp' | translate
```
blakeyp, line 56: We're not in the habit of doing it but could we add ids for the radio items? then we can write less brittle e2e tests (currently most seem to rely on the order of radio items or text content itself)

.kiro/steering/tech.md
DanCorderIPV: Yes, this is deliberate, but if I'm the only one using Kiro, maybe we don't want them? They are supposed to give Kiro context about the project so it can make better suggestions. By committing them everyone can add things if they spot Kiro making mistakes, and everyone should benefit from the files being here.

blakeyp: I think it's fine to keep them, maybe just add some context on Kiro and the script in the project README and say it's a bit of a trial for now? Be great to see a demo in the next Tech Sync & Share!

blakeyp (approved):

PR #2862, opened 2026-05-29
Pyic 9063 new mitigation pages

DanCorderIPV:
### What changed

Add new pages

### Why did it change

For the new mitigation routes that are coming

### Issue tracking

- [PYIC-9063](https://govukverify.atlassian.net/browse/PYIC-9063)

[PYIC-9063]: https://govukverify.atlassian.net/browse/PYIC-9063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


locales/en/translation.json
```diff
@@ -359,6 +359,23 @@
...
       }
     },
+    "appPassportProveIdentity": {
+      "title": "You need to use the GOV.UK One Login app to prove your identity",
+      "header": "You need to use the GOV.UK One Login app to prove your identity",
+      "content": {
+        "paragraph1": "If you cannot use the app, you can go to the service you were trying to use to find out if there are other ways to access it.",
+        "subHeading1": "What would you like to do?",
+        "formRadioButtons": {
+          "useApp": "Prove your identity with the GOV.UK One Login app",
```
AmritSidhu, line 369: Seems to be two variants of this on the ticket, The screenshot/figma show different text to the one in the Notes column

views/ipv/page/need-biometric-passport.njk
```diff
@@ -11,14 +10,22 @@
...
 {% set isPageDataSensitive = false %}
-{% set contentID = '059cc6c8-6e04-4de9-a6c0-5d26cd9e53c0'%}
+{% if pageContext.appOnlyMitigation %}
+    {% set contentID = '044793a2-b5d8-45a1-add9-ed6780e45601'%}
+{% else %}
+    {% set contentID = '059cc6c8-6e04-4de9-a6c0-5d26cd9e53c0'%}
+{% endif %}
 
 {% set showBack = true %}
 {% set hrefBack = "/ipv/journey/need-biometric-passport/back" %}
```
AmritSidhu, line 20: Ticket says based on the URL's this page should be called passport-biometric-chip?

locales/check-translations.ts
```diff
@@ -10,6 +10,57 @@
...
+  "error.serviceUnavailable.content.paragraph2Cy",
+  "error.serviceUnavailable.content.paragraph3En",
+  "error.serviceUnavailable.content.paragraph3Cy",
+  "general.shared.govUKHomepageButtonHref",
+  "pages.pageF2fHandoff.content.firstCircle",
+  "pages.pageF2fHandoff.content.secondCircle",
+  "pages.pageF2fHandoff.content.thirdCircle",
+
+  // Text awaiting translation
+  "pages.appPassportProveIdentity.content.paragraph1",
```
AmritSidhu, line 32: Just a thought to save us having to do this, could we not just do "<PLACE-HOLDER-TEXT>", that way it doesnt pick up these. Unless there are situations where we go live with pages without the welsh text which is a possible so therefore probably can ignore this

locales/check-translations.ts
```diff
@@ -125,43 +176,28 @@ function findStructuralIssues(
...
+  const cyValue = getTranslationFromFullyQualifiedName(welshTranslations, key);
   if (
-    getTranslationFromFullyQualifiedName(englishTranslations, key) === undefined
+    typeof enValue === "string" &&
+    typeof cyValue === "string" &&
+    enValue !== cyValue
   ) {
     throw new Error(
-      `UNTRANSLATED_WHITELIST contains "${key}" which does not exist in English translations`,
+      `UNTRANSLATED_WHITELIST contains "${key}" which is now translated and should be removed from the whitelist`,
```
AmritSidhu, line 196: Nice to have!

views/README.md
```diff
@@ -45,12 +57,26 @@ From a number of sources:
...
     ```
-   - `translate` & variants are defined in `src/config/nunjucks.js`.
+   - `translate` & variants are defined in `src/config/nunjucks.ts`.
      - It takes the reference as a key to find the content saved in `locales` files for english/ welsh.
      - It can take secondary params, e.g.: `context`, which adds a suffix to the reference it's going to find.
-3. `locals.js`:
+   - Additional translation filters:
+     - `translateToEnglish` - always translates to English regardless of current language.
+     - `translateWithPageContext(pageContext, contextKey)` - translates using a context value from the pageContext object. Throws an error if the context key doesn't exist.
+     - `translateWithPageContextOrFallback(pageContext, contextKey)` - translates using a context value, falling back to the base translation if the context key doesn't exist.
```
AmritSidhu, line 66: Nice updating the docs!🫡

views/README.md
```diff
@@ -64,8 +90,8 @@ Breaking this example down:
     ```
   - Relevant details:
     - `pageTitleKey` is used for finding title from content, and `pageErrorState` may trigger an "Error: " prefix.
-    `showLanguageToggle` toggles the language toggle.
```
AmritSidhu, line 67: We still do have this, however never seen it be toggled before

locales/en/translation.json
AmritSidhu: `Prove your identity with the GOV.UK One Login app`
`Prove your identity using the GOV.UK One Login app`

views/ipv/page/need-biometric-passport.njk
DanCorderIPV: I think you maybe misread the bit that says this page appears after `passport-biometric-chip`?

locales/check-translations.ts
DanCorderIPV: Hmm, I think I'd prefer to add a check for blank translations so that we don't accidentally have any :)

Doing this is a little annoying, but it should ensure that we don't miss things by accident.

views/README.md
DanCorderIPV: It's not in `prove-identity-no-other-photo-id` though which is what this section is about

locales/en/translation.json
DanCorderIPV: I've asked Kim to confirm which is correct

views/ipv/page/need-biometric-passport.njk
AmritSidhu: Im referring to the D&A requirements of the ticket. Where it gives the Page URL's along with the contentID, pageSensitive info for these pages. Must just be a typo

DanCorderIPV: Oh good point, those must be wrong

locales/en/translation.json
DanCorderIPV: Kim has changed the Figma to use `with` and I've updated the ticket with a new screenshot

views/ipv/page/need-biometric-passport.njk
DanCorderIPV: I'll check with D&A whether the content ID is actually the one we want

DanCorderIPV: Rachael has confirmed that the content ID on the ticket is correct, but the URL isn't

AmritSidhu (approved):

PR #2875, opened 2026-06-04
Pyic 9092 update readme

DanCorderIPV:
### What changed

Readme files

### Why did it change

To make them clearer and more correct

### Issue tracking

- [PYIC-9092](https://govukverify.atlassian.net/browse/PYIC-9092)

## Checklists

[PYIC-9092]: https://govukverify.atlassian.net/browse/PYIC-9092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.kiro/steering/tech.md
```diff
@@ -6,7 +6,10 @@ The site uses DynamoDb for data storage.
 Configuration is stored in environment variables accessed through dotenv, or in AWS SSM.
 csrf-sync is used to provide csrf protection for the forms on the site.
 Axios is used to send web requests (e.g. to `core-back`)
-The base template for `core-front` is actually stored in the `@govuk-one-login/frontend-ui`
+The base template for `core-front` is `ipv-core-base.njk` from the `@govuk-one-login/frontend-ui` npm package.
```
AmritSidhu: Good additional info but are you able to make this bullet pointed, when previewing the page it doesn't read too nicely

views/README.md
```diff
@@ -1,158 +1,189 @@
...
-     - `translateWithPageContext(pageContext, contextKey)` - translates using a context value from the pageContext object. Throws an error if the context key doesn't exist.
-     - `translateWithPageContextOrFallback(pageContext, contextKey)` - translates using a context value, falling back to the base translation if the context key doesn't exist.
-     - `fullKeyWithContext(pageContext, contextKey)` - returns the full translation key with context suffix appended.
-   - Other useful filters:
-     - `setAttribute(key, value)` - adds an attribute to an object (used for conditional error messages).
-     - `GDSDate` - formats a date in GDS style (e.g. "1 January 2025"), respecting the current language.
-     - `DateDayAndMonth` - formats a date showing only day and month.
-     - `jsonToList` - converts a JSON string to an HTML list.
+    - `translate` & variants are defined in `src/config/nunjucks.ts`.
+        - `translate` takes the reference as a key to find the content saved in `locales` files for english/ welsh.
```
AmritSidhu: english / welsh be capitalised

views/README.md
```diff
@@ -1,158 +1,189 @@
...
+    - Additional translation filters:
+        - `translateToEnglish` - always translates to English regardless of current language.
+        - `translateWithPageContext(pageContext, contextKey)` - translates using a context value from the pageContext object. Throws an error if the context key doesn't exist.
+        - `translateWithPageContextOrFallback(pageContext, contextKey)` - translates using a context value, falling back to the base translation if the context key doesn't exist.
+        - `fullKeyWithContext(pageContext, contextKey)` - returns the full translation key with context suffix appended - used to pass the translation key of the title to base page template.
+    - Other useful filters:
+        - `setAttribute(key, value)` - adds an attribute to an object (used for conditional error messages).
+        - `GDSDate` - formats a date in GDS style (e.g. "1 January 2025"), respecting the current language.
+        - `DateDayAndMonth` - formats a date showing only day and month.
+        - `jsonToList` - converts a JSON string to an HTML list.
```
AmritSidhu, line 63: 'a HTML list'

views/README.md
```diff
@@ -1,158 +1,189 @@
...
     ```
 
-- Sets the page title using English/ Welsh content.
+- Set the page title using English/ Welsh content.
+    ```html
+    {% set pageTitleKey = 'pages.proveIdentityNoPhotoId.title' | fullKeyWithContext(pageContext, 'ninoOnly') %}
+    ```
+    - `fullKeyWithContext` appends a context suffix to the translation key, allowing different title text depending on the page context. Most pages don't need this — a simple `{% set pageTitleKey = 'pages.myPage.title' %}` is usually sufficient. Only use `fullKeyWithContext` if the page title needs to vary based on context.
+
+- Set the back link (if the page should have one).
```
AmritSidhu: Can we also add a small comment here saying this url should reference the page itself. And there is a handler in the core-back which takes care of this event

views/README.md
```diff
@@ -173,40 +204,24 @@ Breaking this example down:
...
 
 1. **[ipv-pages.ts](../src/constants/ipv-pages.ts)** - Add a new entry to the `IPV_PAGES` object with a constant name and the page slug (matching the Nunjucks template filename), e.g.:
    ```typescript
    MY_NEW_PAGE: "my-new-page",
    ```
 
-2. **[page-contexts.ts](../src/types/page-contexts.ts)** - If your page uses a `pageContext` (for context-aware translations), add an entry to the `PageContextMap` type with the page slug and its context shape, e.g.:
+1. **[page-contexts.ts](../src/types/page-contexts.ts)** - If your page uses a `pageContext` (for context-aware translations), add an entry to the `PageContextMap` type with the page slug and its context shape, e.g.:
    ```typescript
    "my-new-page": { someContextValue: string };
```
AmritSidhu, line 249: Since we updated the pageContext to handle different styles and value types, could be useful to very breifly explain those and the purpose of the different value types and again how its handled in core-back

.nvmrc
```diff
@@ -1 +0,0 @@
-20.11.1
```
AmritSidhu, line 1: Is this not useful to keep? When switching projects on different node version, we can use `nvm use`

.kiro/steering/tech.md
DanCorderIPV: This is a Kiro steering file so it's not going to be read by people very often - that said I don't think there's any harm in making it nicer for us to read

views/README.md
DanCorderIPV: Not sure I agree - why do you think it should be 'a' rather than 'an'?

.nvmrc
DanCorderIPV: I wasn't sure if anyone actually uses it - we're now on Node v24.16 so it hasn't been updated in a while

views/README.md
DanCorderIPV: There is a dedicated `Page contexts` section around line 67, I'll add more detail there

AmritSidhu (approved):

PR #2936, opened 2026-07-09
PYIC-9162: Add openBanking context to various pages.

undermad:
### What changed

Added open banking contexts to 3 pages:
- no-photo-id-web-find-another-way
- photo-id-web-find-another-way
- pyi-no-match

### Why did it change

To create the new dynamic Post Open Banking front end screens, which will handle errors or drop-offs.

### Issue tracking

- [PYIC-9162](https://govukverify.atlassian.net/browse/PYIC-9162)

## Checklists

[PYIC-9162]: https://govukverify.atlassian.net/browse/PYIC-9162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


views/ipv/page/pyi-no-match.njk
```diff
@@ -8,22 +8,31 @@
 {% set isPageDynamic = true %}
```
undermad, line 8: I noticed this flag is not present in no-photo-id-web-find-another-way and photo-id-web-find-another-way.
Should this flag be set to true in there?

blakeyp: It's a flag we pass in the GA4 tracking, so yes it should be set in all dynamic templates

locales/en/translation.json
```diff
@@ -99,7 +99,9 @@
         "paragraph1": "We could not find a match when we checked your details with another organisation.",
         "paragraph1Nino": "We could not find a match for your National Insurance number when we checked your details with HMRC.",
         "paragraph1BankAccount": "We could not find a match when we checked your details with your bank or building society.",
+        "paragraph1OpenBanking": "We could not complete your identity check using online banking.",
         "paragraph2": "To keep your information safe, we cannot tell you which of your details we were unable to match.",
+        "paragraph2OpenBanking": "To keep your information safe, we cannot tell you what went wrong.",
```
blakeyp, line 104: Probably a question for Aaron/UCD, wonder if we can standardise this content so we use the same (the new line) for all variants of this page

src/test-utils/pages-and-contexts.ts
```diff
@@ -36,6 +36,7 @@ export const pagesAndContexts: PagesAndContexts = {
   "no-photo-id-exit-find-another-way": [],
   "no-photo-id-web-find-another-way": [
     { dropout: { reason: "dropout" } },
+    { openBanking: { reason: "openBanking" } },
```
blakeyp, line 39: Optional as may be overkill for now, but there's possibly a use case for multi-valued context here, something like:
```
kbvFail: { check: "kbv", reason: "fail" }
kbvDropout: { check: "kbv", reason: "dropout" }
openBankingFail: { check: "openBanking", reason: "fail" }
```
(Open Banking dropout scenario uses a different page)

locales/en/translation.json
```diff
@@ -179,9 +181,11 @@
       "titleDropout": "Find another way to prove your identity",
       "header": "Sorry, we cannot confirm your identity this way",
       "headerDropout": "Find another way to prove your identity",
+      "headerOpenBanking": "Find another way to prove your identity",
```
blakeyp, line 185: Do we need to set `title` for Open Banking too?

locales/en/translation.json
undermad: I agree, the if checks become little messy with new open banking context. I'll raise it with Aaron.

src/test-utils/pages-and-contexts.ts
undermad: Hmm, after looking into the current usage of multi valued context, I think it might be unnecessary overkill.

Currently, we're using it on the download app pages, where the "smartphone" context is [used in the code](https://github.com/govuk-one-login/ipv-core-front/blob/main/src/app/ipv/middleware.ts#L418-L431) to generate the correct qrCode/appDownloadUrl and is passed to the render options, where "isAppOnly" is used to generate dynamic text on the page. There are no references to "smartphone" in the .njk template.

For the pages modified in this PR, we don't really need any dynamic logic in the code. We only need dynamic text, where a single valued page context should be sufficient.

I understand the idea behind this approach. This makes the code more readable. Right now, the KBV version is named "dropout", and it's not really obvious from the code that it's related to KBV.

A possible solution would be to rename "dropout" to "kbvDropout", but that would require staging the changes separately. I'd suggest raising a separate ticket for that.

blakeyp: Thanks for looking into it, yes makes sense let’s keep it as is for now

blakeyp (dismissed):

blakeyp (approved):

